PR #33, opened 2016-12-25
translate to Chinese simplified

wyh136:


fix:
hi only po/zh-CN.pot is needed, can you resubmit with this file only?

PR #34, opened 2017-02-01
Various bug fixes

Doweig:


client/app/src/accounts/AccountController.js
```diff
@@ -151,19 +150,13 @@
     self.connection = networkService.getConnection();
 
     self.connection.then(
-      function(){
-
-      },
-      function(){
-
-      },
```
fix: should be kept:
- first function: when promise is resolved
- second function: when promise is failed
- third one: when promise is changed

client/app/src/accounts/AccountController.js
```diff
@@ -878,10 +852,23 @@
     };
 
     function sendArk(selectedAccount){
-      var passphrases = accountService.getPassphrases(selectedAccount.address);
-      var networks=self.net
+      var data={
+        fromAddress: selectedAccount ? selectedAccount.address: '',
+        secondSignature: selectedAccount ? selectedAccount.secondSignature: '',
```
fix: need to prefill with passphrase and second passphrase if saved (you get them with `accountService.getPassphrases(selectedAccount.address);`)

package.json
```diff
@@ -32,12 +32,12 @@
   },
   "homepage": "https://github.com/fix/ark-desktop#readme",
   "devDependencies": {
-    "electron-prebuilt": "^1.4.13",
```
fix: needs prebuilt to build for different platform (windows, macosx, ubuntu), on a sigle os. TBC

Doweig: Yes, renaming because of deprecation notice from the github page
> Note As of version 1.3.1, this package is published to npm under two names: electron and electron-prebuilt. You can currently use either name, but electron is recommended, as the electron-prebuilt name is deprecated, and will only be published until the end of 2016.

client/app/src/accounts/AccountController.js
Doweig: reverted

Doweig:
Ok, ready for second review 👍

package.json
fix: ok great you found it out

PR #62, opened 2017-04-02
Add a Brewfile for OS X users

6ea86b96:
I'm a lazy OS X user and don't like typing. What do you think about adding a `Brewfile`?


fix:
great one!
+5

PR #63, opened 2017-04-02
Add a Makefile

6ea86b96:
What do you think about putting `npm` behind `make`. Makes things a little nicer IMO and could be used for building releases etc later on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,10 @@
+all: install client_install
+
+dev:
+	npm start
+
+install:
+	npm i
+
+client_install:
+	cd client; npm i && npm run bundle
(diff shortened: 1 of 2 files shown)


Doweig:
What's wrong with npm scripts? https://github.com/ArkEcosystem/ark-desktop/issues/61
For example, to build on mac: `npm run dist:macos`

6ea86b96:
@Doweig it could be to do with my deep hatred for NPM and love of Make :) However, it did already simplify getting started. https://github.com/ArkEcosystem/ark-desktop/pull/63/files#diff-04c6e90faac2675aa89e2176d2eec7d8R45

fix:
agree would be better to use a npm script instead, as make is not easily available on all platform.

I would see something like npm run install:dependencies

fix:
closing as not usefull, since it needs to have make installed, and this cann be achieved with npm

Issue #60, opened 2017-03-29
Balance When You Open Up Send Button


boldninja:
It would be cool to have a nice field on how much you have in current wallet. See attached image for better understanding:
![Balance](https://s16.postimg.org/j4yfgrcpx/balance.jpg)

dafty-1:
Fixed in #98

PR #64, opened 2017-04-02
Show remaining balance on the sendArk view

6ea86b96:
For issue #60 

First time using Angular. Maybe there's a better way?

### DIFF
--- a/client/app/src/accounts/AccountController.js
+++ b/client/app/src/accounts/AccountController.js
@@ -902,12 +902,16 @@
 
     function sendArk(selectedAccount){
       var passphrases = accountService.getPassphrases(selectedAccount.address);
+      var big = require('big.js')
       var data={
         fromAddress: selectedAccount ? selectedAccount.address: '',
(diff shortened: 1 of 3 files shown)


fix:
nice use of big number, we should generalise on the whole app.

Doweig:
Adding a new module for this feature is a bit overkill.
There's a snippet that already do the same formatting on the transactions objects here https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/AccountService.js#L237
Just need to use this snippet for account, like a "humanTotal" field

fix:
@doweig can you close this if needed then? or do we needs an update to the PR?

mattbeedle:
Ah, cool, sorry, didn't see there was already a snippet for that @Doweig. I agree, definitely not worth a module.

fix:
closing PR but giving +5 for the work. Thanks!

boldninja:
@mattbeedle and @6ea86b96 can you both provide me with your Ark address for +5 ARK where you want them sent.

PR #84, opened 2017-04-20
Possible comparison bug?

kristjank:
According to docs: The typeof operator returns type information as a string. There are six possible values that typeof returns: "number," "string," "boolean," "object," "function," and "undefined.".
In the code - the comparission is done with 'String'  not 'string'. If data is object, new String('123'), then already first if gets executed, as typeof returns 'object'


fix:
true that! good catch!
+5

PR #96, opened 2017-05-02
Send all button to populate amount with all available funds

### LINKED ISSUE #74: Keyboard shortcut / button to send all available funds
In Electrum, you can type an exclamation mark in the amount field when creating a new transaction, and it would populate with the field with the entire balance, minus the fee. This would be handy feature.

An "All" button would do the same job.

It refers to it here in the CLI, but the same behavior can be done in the UI:
http://docs.electrum.org/en/latest/cmdline.html?highlight=exclamation

dafty235:
Fixes #74.

image


fix:
🎉

client/app/src/accounts/view/sendArk.html
```diff
@@ -25,10 +25,17 @@
...
-          <label>{{'Amount'|translate}} (ARK)</label>
-          <input ng-model="send.data.amount" type="number" step="0.00000001" ng-required="true">
-        </md-input-container>
+        <div layout-gt-sm="row">
+          <md-input-container md-no-float class="md-block" flex-gt-sm="85">
+            <label>{{'Amount'|translate}} (ARK)</label>
+            <input ng-model="send.data.amount" type="number" step="0.00000001" ng-required="true">
+          </md-input-container>
+          <md-input-container class="md-block" flex-gt-sm="15">
+            <md-button ng-click="fillSendableBalance()" class="md-primary" translate>
```
fix: translate should be put outside "active" angular element, otherwise it disappears
<translate>Send All</translate>

dafty235:
👍

fix:
Hi can you merge with last version, since i merged a PR with a fix on amount.

dafty235:
Merged and updated.

fix:
+5

PR #97, opened 2017-05-03
Add refresh account balance button to dashboard

### LINKED ISSUE #76: Homepage reload button
When I receive some ARKs the balances on the homepage is not getting updated and I have to restart the whole client. Why don't you replace the close (X) button in the title bar with reload and when you click it update all balances?

dafty-1:
First real dive into promise logic. This can probably be done in a more elegant way?

Fixes #76

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -154,7 +154,10 @@
                <div class="md-toolbar-tools">
                  <h2><translate>My Accounts</translate> {{ul.network.symbol}}{{ul.myAccountsBalance()}} / {{ul.currency.symbol}}{{(ul.myAccountsBalance()*ul.connectedPeer.market.price[ul.currency.name]).toFixed(2)}}</h2>
                </div>
-               <md-button class="share" md-no-ink ng-click="ul.selected=ul.accounts[0]" aria-label="Switch of dashboard">
+               <md-button class="share" md-no-ink ng-click="ul.refreshAccountBalances()" aria-label="Refresh balances">
+                 <md-icon md-font-library="material-icons">cached</md-icon>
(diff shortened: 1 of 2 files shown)


luciorubeens:
hi @dafty-1, i have a idea for more elegant way, add a boolean param in function `selectAccount`, change `self.selected` to `selected`, and after all code put `if (newParam) self.selected = selected`.. because `self.selected` change current view for selected address..

luciorubeens:
and in your loop, just call `self.selectAccount(self.accounts[i], false)`, this function check connection and show toast notification if is disconnected or when connect.

fix:
true that @luciorubeens
@dafty-1 Could you please update the PR with lucio suggestion. Both of you will get 5 arks for this PR when accepted

dafty-1:
My concern is that `selectAccount` function pulls in all the transactions related to an account and performs some heavy sorting on the timestamps. The only fields visible on the dashboard are the account and balance so this is unnecessary, particularly if this is called multiple times because a user has 20+ accounts added to their client.

If so, adding a parameter to `selectAccount` to 1) skip setting the account on self and 2) skip updating the transactions seems like burying the functionality in a core function (and refactoring all other calls to include this new parameter) and should be delegated to it's own controller function instead for clarity.

Happy to change it of course, what do you think @luciorubeens @fix?

luciorubeens:
Thats true, for more accounts transactions load too slow. So in your code i would just remove promise updating `self.account.balance=`.

luciorubeens:
i'm wrong, your code looks good, sorry.

fix:
+5
sorry for late acception

Issue #83, opened 2017-04-19
Unclear: This account has never interacted with the network. This is a cold wallet


Jarunik:
It is completely unclear to most new users what that should mean.
I think we would be better of without that information for the "grandma" user.

We should at least change it to something like: There were not transactions yet.
Most will not understand "Cold Wallet". It sounds like broken to them.

fix:
Yes true, it should be rephrased not to look scary.

mody199610:
@fix This Should be closed now not confuse anyone

Jarunik:
We get questions about it in slack every day.

PR #104, opened 2017-05-08
More User Friendly

mody199610:
Changed the cold wallet message to a more friendly one that wont confuse ARK Users.

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -309,7 +309,7 @@
                   <md-card-title>
                     <md-card-title-text>
                       <span>
-                        <translate>This account has never interacted with the network. This is a cold wallet.</translate>
+                        <translate>This account has never had any transactions yet.</translate>
                       </span>
                     </md-card-title-text>
                   </md-card-title>
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -309,7 +309,7 @@
                   <md-card-title>
                     <md-card-title-text>
                       <span>
-                        <translate>This account has never had any transactions yet.</translate>
+                        <translate>This account has never interacted with the network. This is a cold wallet. No Outbound Transactions has been done yet.</translate>
                       </span>
                     </md-card-title-text>
                   </md-card-title>


mody199610:
#83 Done @fix

boldninja:
This account has never had any transactions yet. 
Hm I think this could also be confusing to people who have transactions in that address, but no tx out.

dafty-1:
Do we need a message at all? The people who know what a cold wallet is would already know that it's a cold wallet, and the people who don't know get totally spooked by it.

If this is about the address not appearing on any explorers, then perhaps it should be changed to say that instead, in simple terms.

zillionn:
Or something like:
This is an offline savings wallet (cold wallet). To make it visible online, you need to send some funds.

fix:
Well technically a cold wallet is a wallet that never SENT a transaction. A cold wallet technically is a wallet which public key is unknown. It will be known when the first transaction OUT is recorded on the blockchain.

mody199610:
True @fix however we'd need a more friendly message to appear since any new crypto user would freak out if his funds got to his wallet a bit late

boldninja:
@mody199610 cold wallet will stay cold even if you send Ark in it - it only changes from cold to hot once its been used for TX out .

dafty-1:
My understanding of a cold wallet is an address created offline and maintained offline until you create and broadcast a transaction using any machine. If the machine used to send the tx is connected to the internet and has the passphrase typed in, it becomes hot.

What if you generated a transaction offline using the passphrase and then broadcasted it from a separate machine? In my opinion the wallet is still cold?

What I'm trying to say is I don't think cold wallet is the right terminology. It depends on how the wallet was set up and maintained. Saving the passphrase on a library PC doesn't make it cold just because a tx hasn't been sent out yet?

zillionn:
I believe this message is for non-technical people, so it doesn't matter if it's technically correct. You could just write:
Cold wallet. No outbound transactions yet.

fix:
@zillionn sounds good to me. @mody199610 can you update your PR with this?

mody199610:
@fix  Done :)

fix:
+5

PR #106, opened 2017-05-09
Add QR code scanner

luciorubeens:
Add new feature to scan QR code from webcam, useful for import from papper wallet or mobile.

fe86e1f4-acc4-4b86-aa3f-f94024b9633b


zillionn:
If you could make this work when sending ARK, it will be very nice!

luciorubeens:
@zillionn yes, was added these option when sending ARK too!

fix:
great PR!
+5

Issue #110, opened 2017-05-11
add new currency Korean Won


cannabanana:
was thinking that the counter value to ark should also have Korean Won since crypto is starting to also grow there also.

luciorubeens:
Fixed

PR #115, opened 2017-05-13
Add config service and new currencies

luciorubeens:
Used data from fixer.io to add BRL and KRW currencies. A future implementation to add/remove any currency as user prefer.

#110

### DIFF
--- a/client/app/src/accounts/AccountController.js
+++ b/client/app/src/accounts/AccountController.js
@@ -1230,7 +1230,8 @@
             return ;
           }
 
-          if($scope.createAccountDialog.data.repassphrase == $scope.createAccountDialog.data.passphrase){
+          var words = $scope.createAccountDialog.data.repassphrase.split(' ');
+          if($scope.createAccountDialog.data.word3 === words[2] && $scope.createAccountDialog.data.word6 === words[5] && $scope.createAccountDialog.data.word9 === words[8]) {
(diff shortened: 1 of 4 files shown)


client/app/src/accounts/AccountController.js
```diff
@@ -71,7 +71,7 @@
    * @param avatarsService
    * @constructor
    */
-  function AccountController( accountService, networkService, storageService, changerService, $mdToast, $mdSidenav, $mdBottomSheet, $timeout, $interval, $log, $mdDialog, $scope, $mdMedia, gettextCatalog) {
+  function AccountController(configService, accountService, networkService, storageService, changerService, $mdToast, $mdSidenav, $mdBottomSheet, $timeout, $interval, $log, $mdDialog, $scope, $mdMedia, gettextCatalog) {
```
fix: this is service is not used in AccountController

fix:
I don't understand fully configService. I am not confortable with the concept: Should we migrate all "static" data inside? such as 
- https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/AccountController.js#L84
- https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/AccountController.js#L84

fix:
Hi Lucio, anything on it. There is another PR #129 that could be linked to this. Are you working on a new version?

luciorubeens:
Hi. Yes, i'm adding #124 too.

PR #124, opened 2017-05-25
Add a sound effect when receive funds

luciorubeens:
By default that sound has disabled, but user can enable on settings menu. I selected a fun sound from Super Mario just as an example.


fix:
+5

PR #133, opened 2017-06-02
remove about application menu item

bradyo:
Removes the non-functional "About Application" menu item.

https://github.com/ArkEcosystem/ark-desktop/issues/81

screenshot from 2017-06-02 00 13 30

### DIFF
--- a/main.js
+++ b/main.js
@@ -25,8 +25,6 @@ function createWindow () {
     {
       label: "Application",
       submenu: [
-        { label: "About Application", selector: "orderFrontStandardAboutPanel:" },
-        { type: "separator" },
         { label: "Disable screenshot protection (unsafe)", click: function() { mainWindow.setContentProtection(false) }},
         { type: "separator" },
         { label: "Minimize", click: function() { mainWindow.minimize(); }},


fix:
Hi thanks for attention to details:
- Do you mean there is no about panel popping up in your case?
- I would rather use https://www.npmjs.com/package/electron-about-window for a custom about panel

bradyo:
@fix I was looking at a bug another user created https://github.com/ArkEcosystem/ark-desktop/issues/81
I confirmed that clicking the `Application > About Application` menu item does nothing and it didn't appear there was an about page around in the code.
We can remove the menu item for now just so that there isn't a broken button in the app to make things feel more more solid to users.

I like the idea of using `electron-about-window` for the about popup. I just tried hooking it up and wasn't able to get it working with the browserify build pipeline. I think we may need to transpile the js in that module to a lower version to be compatible with the app. I'll take a look at it more later and try to get it working.

For now you can either merge this PR or leave the broken menu item in place and until I can get `electron-about-window` working.

zillionn:
It works on Mac OS:
screen shot 2017-06-02 at 23 35 23

fix:
@bradyo i would rather leave it as is so far, waiting for electron-about-window integration. I give 20 ark bounty for this.

zillionn:
There you go:
https://github.com/ArkEcosystem/ark-desktop/pull/138

fix:
ok +5 for @zillionn and +5 for @bradyo

Issue #109, opened 2017-05-11
Delete Account Unclear


Jarunik:
It is not clear enough what happens if you delete an account in desktop client.

The users need some information what happens. Namely they need to know that it is only removing the account from the client and not really delete the Account itself.

Slack chat example:
------
ionesscumihai1166 [6:07 AM] 
question....when I use delete

[6:52] 
i delete the account from the walet or I really delete the account forever

[6:54] 
I mean if I  delete an account and then install ark desktop on another computer....if I enter my passphrase ....my account with the money still exists?

[6:07] 
i want to uninstall the ark desktop from a computer,,,,and wipee all the data regarding ark. but I see if I uninstall ark desktop and reinstall it....it still has my account accesible without asking for any passphrase

and I do not see any other option

joelafc
[6:11 AM] 
i dont think you can ever delete an account forever

[6:11] 
its tied to your 12 word phrase

djselery [6:11 AM] 
Delete it. Your paraphrase is not going away

[6:11] 
Just that account

ionesscumihai1166 [6:12 AM] 
i'm not worried about my passphrase....a'm worried about the money inside :slightly_smiling_face:

djselery [6:12 AM] 
Your paraphrase is your money

ionesscumihai1166 [6:13 AM] 
ok...i'll create an account....transfer 1 ark and try it first

djselery [6:14 AM] 
If your too tin foil hat about it you can transfer all your funds before you delete the account heh:grinning: (edited)

ionesscumihai1166 [6:15 AM] 
but in this case.....if so....i would suggest for future wallets to use "remove account from wallet" button....'cos if you enter your account and then you have a delete button ......people logically understand that you delete the account not remove it from the wallet

jarunik [6:17 AM] 
I will open it as Issue in github

fix:
Yes deleting account is misleading. It is just removing from the client, but obviously the blockchain is untouched.

zillionn:
Just change the confirm button title to "Delete account from this computer.".

PR #134, opened 2017-06-02
reword delete account to remove account

bradyo:
- reward 'delete' account' to 'remove' account
- make confirm dialog text more clear about the expected results 
- change delete icon to clear icon, since it is a better indicator of actual action

#109 

screenshot from 2017-06-02 01 21 05
screenshot from 2017-06-02 01 21 23


fix:
nice one! +5

Rutix:
FYI if you put closes #109 in the commit message next time it will automatically close the issue too.

PR #136, opened 2017-06-07
Topbar improvements

fileninja:
Hey,

I've changed the Toolbar on the top right a bit. I created some small Tooltips on Mouse Hover (and translated to german already) and changed the Language Selection to Dropdown/Selection-Box for an more intuitive feeling (before it was just switching to the next language).


fix:
many conflicting files because of prettyfiying edition, please make a separate PR for that

fileninja:
Removed the prettyfiying and resolved the master conflicts, resulting from other PR's.
@fix you should be able to merge now.

PR #137, opened 2017-06-07
Modal close buttons

fileninja:
Just a small Usability improvement.
As every Dialog in any case has a "Cancel"-Button, it makes sense to also provide a "X" in the top right corner to close modals for user friendly interaction.
I added it to all actual modal dialogs.
Furthermore I changed the way how "add Watch-Only Address" was handled actually. Every Modal in the Code was handled by an own html template except this one, so i changed it to the regular way here and created a template.
With this change I also added a Close Button here to keep it consistent and solid.

### DIFF
--- a/client/app/src/accounts/AccountController.js
+++ b/client/app/src/accounts/AccountController.js
@@ -746,7 +746,51 @@
     /**
      * Add an account
      */
-    function addWatchOnlyAddress() {
+    function addWatchOnlyAddress(){
+      
+      function cancel() {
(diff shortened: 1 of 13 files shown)


fix:
sorry i was a bit late on this one. Can you remove the commented code form this PR. We don't need to keep it if you improved it :)

fileninja:
alright. will do in a second.

fileninja:
done

fix:
+5

Issue #126, opened 2017-05-30
Delete network tab


Jarunik:
I could not find a way to delete my unused testnet tab in manage networks.

It should also be possible to remove added networks again.

fileninja:
implemented in #139

PR #139, opened 2017-06-07
Network Management Improvements

fileninja:
I've improved the "Manage Networks" functions.
If you add a new network, you don't need to re-render the full page, instead it is now adding just a new tab without even reload the modal. 
Also #126 is implemented. You now have the possibility to delete your own added networks. here is also no re-rendering needed as it updates "live"


fix:
+5

Issue #73, opened 2017-04-13
Send ARKs using delegate username


zillionn:
How about when sending ARKs use delegate username not the wallet hex address? I guess you'll only need to use the API to check if that delegate exists and get it's hex address. It'll be a lot easier this way and a lot more people will register delegate/spent 25 ARKs :).

fix:
well even best to use the label that the user can set

zillionn:
The label is not unique, I even set same label for 2 local accounts.

dafty-1:
How to trust the address returned by the API is legit? For example, I could change my node / explorer to always return my address instead of the true one.

Would an address book be better? User saves an alias and an address, and can then select that alias from a dropdown list. Address book is saved in a map on disk, so no third parties to rely on for valid data.

dafty-1:
Duplicated by #107

fix:
yes i agree address book is better, so the user won't be subject to phishing

fileninja:
implemented in #141

PR #141, opened 2017-06-08
Address-Book

fileninja:
implemented an Address-Book with options to add, delete and modify contacts. to open it, you can click the book symbol in the top menu bar. I've also added an Autocompletion example in the sendArk Function. Just type an username of your added contacts and the text field will fill in the deposited address.

### DIFF
--- a/client/app/src/Addressbook/AddressbookController.js
+++ b/client/app/src/Addressbook/AddressbookController.js
@@ -10,14 +10,14 @@
     var contacts;
 
     self.getContacts = function () {
-      self.contacts = storageService.getGlobal("contacts");
+      self.contacts = storageService.get("contacts");
       if (self.contacts == null || self.contacts == undefined) self.contacts = [];
     }
 
     self.getContacts();
 
     self.save = function () {
-      storageService.setGlobal("contacts", self.contacts);
+      storageService.set("contacts", self.contacts);
       $scope.$apply;
     }
 
--- a/client/app/src/Addressbook/AddressbookController.js
+++ b/client/app/src/Addressbook/AddressbookController.js
@@ -21,6 +21,11 @@
       $scope.$apply;
     }
 
+    self.isAddress = function (address) {
+      var isAddress = /^[1-9A-Za-z]+$/g;
+      return isAddress.test(address)
+    }
+
     self.contactExists = function (name) {
       var i;
       for (i = 0; i < self.contacts.length; i++) {
@@ -56,6 +61,10 @@
           self.showToast('this Contact-Name is already taken, please choose another one: ', contactname);
           return;
         }
+        if(!self.isAddress(contactaddress)) {
+          self.showToast('this seems to be not a valid Address: ', contactaddress);
+          return;
+        }
         self.contacts.push(newcontact);
         self.save();
         self.showToast('Contact successfully added: ', contactname);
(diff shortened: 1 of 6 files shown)


fileninja:
#73 is done with that.

zillionn:
Wow, great job!

fix:
agree great one!

client/app/src/Addressbook/AddressbookController.js
```diff
@@ -0,0 +1,136 @@
...
+    .controller('AddressbookController',
+    ['$scope', '$mdDialog', "$mdToast", "storageService", "gettextCatalog", AddressbookController]);
+
+  function AddressbookController($scope, $mdDialog, $mdToast, storageService, gettextCatalog) {
+
+    var self = this;
+    var contacts;
+
+    self.getContacts = function () {
+      self.contacts = storageService.getGlobal("contacts");
```
fix: should not be stored at global level. Use get/set instead

client/app/src/accounts/AccountController.js
```diff
@@ -1000,8 +1000,9 @@
 
       function querySearch(text){
         text=text.toLowerCase();
-        var filter=self.accounts.filter(function(account){
-          return (account.address.toLowerCase().indexOf(text)>-1) || (account.username && (account.username.toLowerCase().indexOf(text)>-1));
+        var contacts = storageService.getGlobal("contacts");
```
fix: use storageService.get("contacts")

client/app/src/Addressbook/AddressbookController.js
```diff
@@ -0,0 +1,136 @@
...
+      if (self.contacts == null || self.contacts == undefined) self.contacts = [];
+    }
+
+    self.getContacts();
+
+    self.save = function () {
+      storageService.setGlobal("contacts", self.contacts);
```
fix: use storageService.set(...)

fileninja:
Ah I See.. I will rework it

fileninja:
removed the global storage, so your hints are resolved now

zillionn:
Couple things:
1. How about to change the NEW button to something like this, so it's more distinguishable:
<md-button class="md-button" ng-click="ab.addAddressbookContact()"><md-icon md-font-library="material-icons">library_add</md-icon> <b translate>ADD ADDRESS</b></md-button>

2. Also you should put the NEW/ADD button first, because when you add more addresses you need to scroll to the bottom to see it.

3. What's the point of "md-prevent-menu-close"? It doesn't close when I click NEW.

fileninja:
all of your Points are legitimate. I have to admit that I was more Focused in the functionality.. I will push a new improved commit.

fileninja:
here it is

client/app/src/Addressbook/AddressbookController.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+      }
+      return false;
+    }
+
+    self.showToast = function (message, variable) {
+      $mdToast.show(
+        $mdToast.simple()
+          .textContent(gettextCatalog.getString(message) + variable)
+          .hideDelay(5000)
```
zillionn: You could add ".theme('error')" if error toast so it will be red.

fileninja:
good point. added the theme and furthermore added an check if the text fields are empty.

fix:
great PR! +5

luciorubeens:
hi @fileninja, i tested your feature today and i liked! i just add some [fixes](https://github.com/ArkEcosystem/ark-desktop/commit/f189f50dc26b968416765b4a6918becef3232188) to better ux.

PR #144, opened 2017-06-09
Refactored language picker to a more usable pattern than "skipping to the next"

vekexasia:
Hello, 

I've noticed that the current version of the wallet just cycles through all possible languages.

I think the approach proposed in this pull request is a bit better in term of User Experience.

As you can see in the screenshot below, I've created a new menu which opens the all available languages in a Dropdown. Current active language is highlighted. (English in the screenshot)
image

Cheers :)

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -52,9 +52,6 @@
             <md-menu-content flex layout-padding width="6">
                 <md-list>
                   <md-list-item>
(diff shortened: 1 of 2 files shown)


fileninja:
already done in #136

vekexasia:
Oh. It looks likeI need an extra pair of glasses. I was pretty sure I double checked issues and pull requests before starting the work :)

PR #145, opened 2017-06-09
Added some icons and css + fixed dialog wrong buttons positioning

vekexasia:
Hello,

This commit includes two group of things: 

**I've refactored the menus a bit to include some icons**

**Current**

image

**Proposed**

image
---
**Current**
image
**Proposed**
image

Also fixed an issue with auto focusing and angular (see last screenshot "QUIT" is by default selected)

--> I've noticed that the "Vote delegate" dialog uses an anti-pattern in material-design guidelines. Particoularly the actions on the dialog are in the wrong order and not using the md classes: see difference:

***Current*** 
image
***Proposed***
image

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -56,6 +56,12 @@ md-list-item > button {
   width:100%;
   text-align:left !important;
 }
+md-list-item button md-icon {
+  margin-right: 16px;
+}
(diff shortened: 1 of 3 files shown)


fix:
Hi, I like it :)

+5

PR #152, opened 2017-06-19
Language selector

fileninja:
this is the Language-Selector Part from the topbar improvements, which is now implemented as a more usable dropdown.


fix:
+5 nice one

PR #153, opened 2017-06-19
Tooltips in Navigation Bar

fileninja:
this is the Tooltip Part from the topbar improvements. Just added some small "On Mouse Hover" Tooltips to the Topbar, so the User can see what is behind the icon.


fix:
I like it thanks!
+5

Issue #89, opened 2017-04-27
Ark Icon on taskbar when wallet is loaded


michaelthecryptoguy:
The Ark Icon looks great on the taskbar when sitting on your desktop not loaded or open. But when you open your wallet, it is hard to see the A for the ark symbol. I think it is due to no background (Or clear). This is on your taskbar, where your icons for your applications are.

fix:
can you add a screenshot? Can't understand the exact issue

michaelthecryptoguy:
I think the Icon needs a background.
![Ark Icon background needed](https://cloud.githubusercontent.com/assets/24502401/25778880/2e9c888a-32d9-11e7-9809-a7fd4ab6b00e.jpg)

fix:
true that, on ubuntu/debian there is likely some settings i forgot to set for this taskbar icon.

PR #154, opened 2017-06-20
Icon Differentiation

fileninja:
Just a really small Change. Added a snippet to differentiate between different operating systems and added an small Linux Icon resolving #89


fix:
+5

PR #155, opened 2017-06-21
Option to clear localStorage

naaomimore:
Simple option to clear all data from localStorage.

### DIFF
--- a/client/app/src/accounts/AccountController.js
+++ b/client/app/src/accounts/AccountController.js
@@ -129,6 +129,20 @@
       else return null;
     };
 
+    self.clearData = function() {
+      var confirm = $mdDialog.confirm()
+            .title('Are you sure?')
+            .textContent('All your data, including created accounts, networks and contacts will be lost and reset to default.')
+            .ariaLabel('Confirm')
+            .ok('Yes')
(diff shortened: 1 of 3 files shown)


client/app/src/accounts/AccountController.js
```diff
@@ -129,6 +129,20 @@
       else return null;
     };
 
+    self.clearData = function() {
+      var confirm = $mdDialog.confirm()
+            .title('Are you sure?')
+            .textContent('All your data, including created accounts, networks and contacts will be lost and reset to default.')
```
fix: `lost` is too strong here.
I would say `All your data, including created accounts, networks and contacts will be removed from the app and reset to default.` instead

naaomimore: yes much better, agree!

fix: can you push a new version?

fix: also it could be good to have it with translation:
see https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/AccountController.js#L117

fix:
yo
+5

PR #156, opened 2017-06-21
Fix colors from material angular

naaomimore:
Update md-colors directive to use default theme.


fix:
ah good catch, i was wondering why the palette was not clean...
+5

PR #159, opened 2017-07-03
Add support for Moscow clock

Gava97:


luciorubeens:
hi @Gava97, we removed this page because it was unnecessary..

PR #162, opened 2017-07-10
Vote Tab Improvements

fileninja:
I just reworked this tab so it works now like a list is intended for.
Before
ark_vote_1
After
ark_vote_2

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -659,7 +659,7 @@
             </md-tab-body>
           </md-tab>
 
-          <md-tab>
+          <md-tab ng-click="ul.fillAllDelegates(ul.selected)">
             <md-tab-label>
               <translate>Votes</translate>
             </md-tab-label>
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -705,10 +705,9 @@
               </md-table-container>
 
               <md-table-container  ng-if="ul.selected.virtual" flex="100">
-                <md-button ng-click="ul.vote(ul.selected)" md-colors="{background:'default-red'}"><translate>Vote</translate></md-button>
-                <md-button ng-click="ul.addDelegate(ul.selected)" md-colors="{background:'default-green'}"><translate>Add Delegate</translate></md-button>
+                <md-button ng-click="ul.vote(ul.selected)" md-colors="{background:'default-blue'}"><translate>Save Votes</translate></md-button>
                 <a href="#" ng-click='ul.openExplorer("/delegateMonitor")'><translate>Open Delegate Monitor</translate></a>
-                <table md-table md-row-select multiple ng-model="ul.selected.selectedVotes">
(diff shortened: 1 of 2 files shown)


fix:
how can you add delegates that are not in top 51?

zillionn:
@fileninja did you see this PR?
https://github.com/ArkEcosystem/ark-desktop/pull/120

fileninja:
@zillion yes, because i have an up to date master.
@fix you're right I just thinked there wasnt this possibility before. but I've just seen that there is a textfield for writing a custom one. I will think about how I can solve that. either prefilling the table with all delegates and a pagination or otherwise giving the possibility to add not listed ones.

zillionn:
I believe this should be the best way to vote:
Voting - only one button "Vote for a delegate" which opens a modal window with a select and text fields (like #120) to add the delegate name and no "Next" button but "Vote (1 ARK)" which will directly send the vote tx and add the delegate in the Votes tab.
Unvoting - instead of a checkbox, a delete icon and when you click it show a confirm dialog with button "Unvote (1 ARK)" which will send the unvote tx and delete the delegate from the list.

fileninja:
@zillionn this is definetly an option. But if we think about multiple votes i feel more comfortable with an multi select table. just uncheck the ones i dont want anymore and check the ones i want to add and click save. this will remove, add and update everything as i want in one step.

boldninja:
@fileninja but there is only 1 vote per Ark address so multiple selections are kind of unnecessary? What would be cool is you get a list of all delegates - if you are voting for someone that one is at the top of the list and some kind of label like currently voting for. and if you want to change a vote you select another one from the list below and it will unvote currently voting one in 1 tx and after 1 confirmation for unvoting tx is confirmed it will automatically vote for the one you want to change vote for (informing that it will cost you 2 ARK currently to change vote). If you are not voting for anyone then no currently voting for is there or something of that sorts.

fileninja:
@boldninja you're right. just thinked about https://github.com/ArkEcosystem/AIPs/issues/1 which seems to get in "multiple ranked votes" direction.
if we stay at single votes so far, this PR is useless and I will close it.

boldninja:
I think there will be still some time before AIP#1 goes on the table (might be few months) and it needs to be properly tested on devnet if we decide on that route.

fileninja:
alright. i will close the PR right here but will let the branch open in my fork.

Zillion:
Wrong Zilllion pal

<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.
www.avg.com
<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Issue #157, opened 2017-06-22
make address book first class citizen


fix:
Instead of the hacky watch-only address, we should move address book as first citizen.

fileninja:
@fix isnt this the same as just deleting the Watch-Only-Feature? Or do you want to refactor the existing watch-only-buttons to address-book ones?

fix:
well yes and wire address book in the left panel list

PR #164, opened 2017-07-11
Replaced watch-only addresses with addressbook contacts

fileninja:
I Replaced the old Watch-Only-Address Feature with the fully integrated Addressbook.
You can do the same things with your contacts as with the old Watch-Only-Addresses for sure, for example viewing the Account Statistics:
ab_1

#157 should be done with this


zillionn:
And when I update to the new wallet, will my Watch-Only addresses move to the Addressbook?

fileninja:
no as the localstorage string is another. it would be hard to change it to watch-only because the object is another.

Issue #101, opened 2017-05-06
Ui


Moustikitos:
Horizontal scrollbar is at the bottom of transaction list so I have to scroll down all transactions to scroll them horizontally.

fileninja:
can't reproduce, for me on linux the scrollbar is always pinned on bottom

zillionn:
Moustikitos is your resolution lower than 1024x768? The scroll appears only when window width is lower than 950px.

Moustikitos:
No my res is higher

zillionn:
Then just resize the whole window.

Moustikitos:
![image](https://user-images.githubusercontent.com/15268977/26948271-01454332-4c96-11e7-81e4-1b9bfe0cc655.png)

![image](https://user-images.githubusercontent.com/15268977/26948285-0e816e22-4c96-11e7-873a-7f5ff5332fc4.png)

zillionn:
You need to first disable the screenshot protection from the menu.

Issue #49, opened 2017-03-06
Add pagination to transaction list


dafty-1:
The transaction list is limited to the 100 most recent transactions. There is currently no way to view transactions beyond this.

Can we add pagination to the list?

alexbarnsley:
@luciorubeens I think this needs closing since you added infinite scroll

dafty-1:
@alexbarnsley Agree, will close this ticket.

PR #166, opened 2017-07-11
Added Pagination to Transaction List

fileninja:
just another small one which solves #49 and also may do a temporary tricky fix to #101 because the scrollbar is'nt so much down in default.
@fix @Doweig the API-Limit is set to 50 Transactions per Request. It may be useful to increase it a bit. If the Limit will be increased, the pagination is definetly a bit more usefull ;-) 

Screenshot:
pagination_transaction

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -444,7 +444,7 @@
                       </tr>
                     </thead>
                     <tbody md-body style="white-space:nowrap;">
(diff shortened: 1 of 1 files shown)


fix:
yes will need to improve API on ark-node

luciorubeens:
Fixed in #327

PR #172, opened 2017-07-21
Income/expend in Addressbook

### LINKED ISSUE #149: Filter to see Vote Rewards
Hi, please can we have a filter in the wallet, to show Ark Received from Vote rewards (Or even just from a specific person, if that is easier. Infact, yeah, that would be better, especially if you switch delegates at some point). 
Then also have a count to show the Total amount received from said person?

naaomimore:
Add an information in each contact with the total sent and received. the icon only shows if any address has been selected.
Until to see the total received by delegate

maybe fixes #149


fix:
@naaomimore i have reworked a bit the integration of contacts, can you adapt this PR to it (move it on the left menu bar)

PR #178, opened 2017-07-28
Filter on contact sheet

naaomimore:
Only show "Register Delegate", "Timestamp Document" and "Second Passphrase" if the selected account is mine


luciorubeens:
Same as #172, now we have the address book instead of watch-only address and the contact-sheet is no longer displayed for them

Issue #148, opened 2017-06-10
Format option for wallet


N1ckGreek:
Add dd/mm/yyyy format option to the wallet.

fix:
You mean in transactions list? 

I suggest to create a settings popup window panel (like manage networks), because now settings are getting quite crowded

dafty-1:
Only America uses the current mm/dd/yyyy format. The rest of the world uses dd/mm/yyyy.

Suggestion to make dd/mm/yyyy as default and configuration to change this in settings popup.

fix:
agree that would be a nice improvement.

wilruns:
should just make the client follow user's PC date format

PR #179, opened 2017-07-28
implemented Time-Formatting

fileninja:
Added an Time-Format (MM/DD/YYYY vs DD/MM/YYYY) Setting in the Settings Popdown.
The Date Variables in Angular now using this new variable.
Resolving #148


luciorubeens:
would be better to use `toLocaleDateString()` instead, something like #242

PR #190, opened 2017-08-09
Refactor/modules and styles

peterdouglas:
Hi team,
Summary of changes in this pull request;
- first stage of modularization of angular code (separate into modules, re-organise files, etc)
- fix code style issues using eslint (this is why there are so many changes in some of the files, will be completed in next stage of modularisation)
- add watch and auto build processes to make dev easier (instructions for running have been added to the README as an extension to the current workflow).
- alter directory structure to have files compiled into a separate build directory
- remove items and files no longer in use

Next step in the modularization process would be to split AccountController into several different directives / controllers and add routes, and to break AccountService into smaller parts.

Let me know if you have any questions.

Thanks.

PR #195, opened 2017-08-13
Refactor/file system and module naming

peterdouglas:
This was previously part of PR #190 .

This PR changes the file system layout and the module names and configuration to be more modular.

Also contains a very small amount of refactoring (some vars changed to lets mainly).

Thanks.


peterdouglas:
@luciorubeens hey Lucious. :) Split the PR into 3 parts like you suggested. It's been sitting here a while, so I will integrate the new master into it. Is there anything else you would like to see done?
Thanks.

fix:
@luciorubeens can you look at this PR in priority, i want this to be part of upcoming version

luciorubeens:
@fix I'm validating all others PRs first so this will not make conflicts..
hi @peterdouglas, sorry for my late reply. so if we're going to refactor the code, I think we can use this style guide: [johnpapa/angular-styleguide](https://github.com/johnpapa/angular-styleguide/)

peterdouglas:
@luciorubeens ok. No problem. I will git merge with the current master branch and refresh the PRs when I'm back home today; as I haven't done anything since the PRs were lodged, there shouldn't be too many conflicts that git can't solve (although I'm sure there will be a few!). The guide looks fine - most of it (or similar) is incorporated in the changes in the PR. Eslint (which will be incorporated through grunt) is set to use the airbnb standard, which typically follows industry best practices. I have made some temporary changes to the eslint config to allow for it to not break horribly and still test.

The main work still required will be reducing controller/service size more, and there are 2 massive files that incorporate lots of the functionality that I didn't touch in part 1.

peterdouglas:
@luciorubeens I have merged all of the conflicts as of yesterday at 10pm AU Time. It looks like another change has been made to the main branch, so I might need to do another quick one.

peterdouglas:
@luciorubeens looks good from what I have gone through - there are a lot of changes in there that I have made in #196 , but your file naming scheme is definitely closer to current best practice. Nice one! For larger modules, I'm not a fan of the 'everything in a flat directory' approach, but I am guessing that is done to encourage smaller modules and components. 

We should probably close the PR's I have open then, as this change will make merging more difficult than just applying the addition of grunt tooling and build centralisation (#196) to the new branch, and #197 is largely based around optimizing code so that we can use a real-time linter as part of the build process to ensure we stay at (or near) industry best practice for code (currently I have set it to use es6, but if the team would prefer sticking to es5.1 we can do that too). And I can then just make the grunt changes to the new refactor branch. 

Would have preferred to know this is what we would do before investing so much time into the process, but oh well! Nice work @luciorubeens . 👍

luciorubeens:
hmm true that, so i will close your PR's and waiting to you push again with the changes. yes, definitely we need to lint and I prefer es6..

good job 👍🏻 +5

peterdouglas:
@luciorubeens Sounds good. :) Should be done by the first half of next week.

PR #198, opened 2017-08-16
Select option for currency

SBero:
I changed the hidden currency switcher to be a select/option for an easier user experience. Now the user can very quickly change to the currency of their choice without having to potentially cycle through the entire hidden list.


zillionn:
How about to move it in the settings menu below the language selection?

luciorubeens:
@zillionn Yes, i agree. @SBero Can you move it to settings?

SBero:
@luciorubeens Sure. I'll move it into Settings.

SBero:
@luciorubeens Should be good to go now.

luciorubeens:
nice! +5

Issue #185, opened 2017-08-08
The Quit button is not aligned with other items.


dnovacik:
The Quit button seems to be a bit off.
![image](https://user-images.githubusercontent.com/9606240/29069416-39ae0c98-7c3b-11e7-8441-000eab93452c.png)

luciorubeens:
Solved 9d336046cd7e8869cdb5f83e1e465426025d8508.

PR #199, opened 2017-08-16
Fix for quit button alignment #185

SBero:
This resolves the issue #185 with the quit button being not in left alignment with the other buttons in the on/off "Application" panel.


luciorubeens:
hi @SBero! +5

Issue #187, opened 2017-08-08
Import signed messages to verify


Jarunik:
Currently you have to copy/paste 3 fields to verify a signed message.

It would be easier to be able to import the entire signed message as well:

{"publickey":"02c7455bebeadde04728441e0f57f82f972155c088252bf7c1365eb0dc84fbf5de","signature":"30450221008ad2cfe1dc76a90758ed84856036baf42d8713f5d889ad237154468348dfceeb022010732eeb5725e98f3677e584e533f4442c048013591fcd371a3dc81c41961490","message":"Thanks a lot to all my voters for your support. Jarunik"}

fix:
true that

luciorubeens:
Fixed in #200

PR #200, opened 2017-08-16
implemented verifying of copied messages

fileninja:
actually there is the possibility to verify messages by klicking on the button(for own messages) or writing all details into the verify dialog. It would also been useful if you can verify the copied string (which you may receive from the other user). For this I added an addtional tab into the verify dialog.
resolving #187


luciorubeens:
+5 👍🏻

Issue #186, opened 2017-08-08
Make My Accounts and Add / Import "Static" on Main


boldninja:
![scroll](https://user-images.githubusercontent.com/25795454/29073647-b36be53a-7c4c-11e7-8a0b-8a598332c701.jpg)

If you have more accounts (so that scrollbar becomes visible) on the main menu if you scroll down the "My Accounts" stays up and is not visible - maybe would be better if that is seperate and visible even as you scroll down of your Ark accounts and maybe move import address there as well.

![scroll1](https://user-images.githubusercontent.com/25795454/29073734-00b805ee-7c4d-11e7-84bb-a07a0426a8a6.jpg)

Possible solution:

![solution](https://user-images.githubusercontent.com/25795454/29073915-bcc79286-7c4d-11e7-81ab-c365a4cd7661.jpg)

fix:
you fired :)

boldninja:
no, why me :'(

fileninja:
You're a hell of a ninja.. finding things anyone else without eagle eyes would even notice

PR #201, opened 2017-08-16
improved scrolling account dashboard

fileninja:
#186 is done.
the buttons are now in the header which is now sticky to the top. just the list will scroll.


luciorubeens:
+5! I improved height and color to handle new themes option

Issue #202, opened 2017-08-16
Feature request: Automated payments/ rules


Zeehenk:
Hi everyone,

I'm a total n00b with coding (read: I can't do it at all), but I would like to see what's possible with this request. Firstly my issue:

I have voted for a delegate, but this dalegate has capped my max voting weight. This means that I have made a second wallet for voting with my remaining ARK.
I am getting weekly payments by the first delegate to my "capped wallet", but I want to get this payout to my second walled directly. What I would like to set up are some rules like I can do with Outlook. For example for this:

If ARK received from [ADDRESS], than [SEND] [SAME AMOUNT ARK] to [ADDRESS]

I think this would be a great feature for people (like me) who can't code. Even one line of code seems scary for me to do... I'm too afraid to make mistakes in there and lose my ARK.

Obviously this "RULE" feature is not limited to above example. You could have rules to only send ark if you received an X amount from a particular address etc....

Hope I explained everything well enough :)

Cheers!

fileninja:
Maybe Duplicate of #180

Zeehenk:
Thanks, posted it there too

luciorubeens:
Let's keep in #180

Issue #180, opened 2017-07-29
Plugin / Script support for desktop


Jarunik:
We should provide plugin/extension and script support for the desktop.

It would be really great to be able to write and distribute simple extensions of the desktop.

Then you could easily automate stuff and build things into/ontop of the desktop client.

Would allow the desktop client to become the hub for all kind of tools.

You could then also include profit sharing/optimisaiton overviews as an example

fileninja:
I really like the idea. With this er Should provide an Plug-in/Script API. Maybe the ark-js functions or something Complete new

fix:
while the idea is definitely nice, this raise the issue of trusting 3rd part script/plugin, as we (ark team) will not have a full control of what the scripts will be doing...

not sure what the best option

fileninja:
generally i would say the user has to check what type of script he loads and if it has malicious backgrounds. We should really hardl define what type of events goes out and what functions we expose.

Zeehenk:
Hi everyone,

I'm a total n00b with coding (read: I can't do it at all), but I would like to see what's possible with this request. Firstly my issue:

I have voted for a delegate, but this dalegate has capped my max voting weight. This means that I have made a second wallet for voting with my remaining ARK.
I am getting weekly payments by the first delegate to my "capped wallet", but I want to get this payout to my second walled directly. What I would like to set up are some rules like I can do with Outlook. For example for this:

If ARK received from

, than to
I think this would be a great feature for people (like me) who can't code. Even one line of code seems scary for me to do... I'm too afraid to make mistakes in there and lose my ARK.

Obviously this "RULE" feature is not limited to above example. You could have rules to only send ark if you received an X amount from a particular address etc....

Hope I explained everything well enough :)

Cheers!

Kind of the same topick here:
(https://github.com/ArkEcosystem/ark-desktop/issues/202#issuecomment-322736785)

zillionn:
To accomplish this you need to save your passphrase for address A in the wallet/plugin so it can send ARK from address A to B. Saving the passphrase is a very bad idea.

sleepdefic1t:
@Jarunik @fix @fileninja @zillionn 

If you think of ArkClient as a backend in this scenario, 
it can basically become an RPC client.


Or even a simple App URI could open a transaction request.
`arkclient://transaction/create?recipientId=AxX...Xx&amount=100000000&vendorField=......`

I have a friend learning and trying to do a "pay with Ark" button for websites;
[...]

PR #207, opened 2017-08-16
Plugin Loader

fileninja:
Just developed a small plugin loader. Its event driven, meaning the wallet calls an event (which we need to define in the code where we want to trigger it) and the plugins may decide to implement a function for handling this event. The other way around its possible to call defined functions which are capsulated in an own scope.
To create a plugin just create a new folder in the "plugins" directory, call it however you want. In that directory create 2 files (plugin.conf and plugin.js). in the conf file you can define eventhandlers which will be handled in the plugin.js file. An example can be viewed in the sampleplugin folder. you may want to copy the whole sampleplugin directory to the plugins folder and see it in action.

Indeed there are still some improvements possible. one thing would be to do the function calling asynchronous. the other one would be the extension/adding of events and callback functions to create a bigger using range. But this ones I will do on a additional commit and/or Pull Request.

Note that the plugins will be loaded at start, so to add a plugin you need to restart the application.

May resolving #202 and #180

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -1001,6 +1001,7 @@
   <script src="src/accounts/AccountService.js"></script>
   <script src="src/accounts/AccountController.js"></script>
   <script src="src/network/NetworkService.js"></script>
+  <script src="src/plugin/pluginLoader.js"></script>
   <script src="src/changer/ChangerService.js"></script>
   <script src="src/ledger/LedgerService.js"></script>
   <script src="src/directives.js"></script>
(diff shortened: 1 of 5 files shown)


luciorubeens:
It will be useful! Merged manually +5 👍🏻

fix:
Just big warning:
**This could be the gate for malwares**

I would improve the UI, where the other can easily check which plugin is loaded, when a new one is getting installed (with optional optin), if the plugin signature has changed etc...

Issue #182, opened 2017-08-08
need dark/night mode


djselery:
Themes are cool but the bright white is too much, need an option to switch to nightmode.

luciorubeens:
Fixed #294

PR #209, opened 2017-08-16
Night/Dark Mode

SBero:
I've created the initial setup and implementation for dark mode #182.

An issue that does need to be considered further, is that if you use anything other the default theme you will have issues with Dark Mode.

The toggle for it can be found under the 'Settings' panel

### DIFF
--- a/client/app/app.js
+++ b/client/app/app.js
@@ -113,3 +113,28 @@ document.body.addEventListener('contextmenu', (e) => {
     node = node.parentNode;
   }
 });
+
+function turnOnDarkMode() {
+  "use strict";
+  var headID = document.getElementsByTagName("head")[0];
+  var cssNode = document.createElement('link');
+  cssNode.type = 'text/css';
+  cssNode.rel = 'stylesheet';
+  cssNode.href = 'assets/dark-mode.css';
+  cssNode.media = 'screen';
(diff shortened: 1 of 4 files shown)


boldninja:
Just a suggestion - if you select night/dark mode 
1. It saves your current theme you had selected
2. Switches to default if not default and enables dark mode
3. If you disable it reads what you had before if it wasn't default theme

Don't know if its possible just a suggestion as I'm not a dev.

SBero:
Good minds think alike! This is along the same lines that I was thinking was probably the best way to do it as well.

SBero:
Ok I've updated it with a fix to allow the users theme to be stored and restored when they toggle dark mode, so that it always works now.

luciorubeens:
Sounds good, but angular material solve this with `dark()` method. [Take a look](https://material.angularjs.org/latest/Theming/03_configuring_a_theme).

SBero:
I've tried this with material. It's inconsistent and the dark() method doesn't seem to apply to the application properly. Some of the themes don't even do anything when called. This solution resolves it for all of the themes/styles that the user would choose.

luciorubeens:
Fixed in e974bee51c167c582fdf4b29b90a9496990596eb

Issue #206, opened 2017-08-16
Message verification form doesn't return error with wrong data


biolypl:
Try insert this values in message verification form and click verify:
Message: 30450221008ad2cfe1dc76a90758ed84856036baf42d8713f5d889ad237154468348dfceeb022010732eeb5725e98f3677e584e533f4442c048013591fcd371a3dc81c41961490
Public Key:
02c7455bebeadde04728441e0f57f82f972155c088252bf7c1365eb0dc84fbf5de
Signatur:
Thanks a lot to all my voters for your support. Jarunik

Wallet hangs for me. Another combinations return "normal" error

fileninja:
true that. there was a mistake for not checking the input against hexadecimal so it was'nt possible to convert it to hex and with that no return happened.

fileninja:
there you go

luciorubeens:
Fixed in #212

PR #212, opened 2017-08-17
implemented input check against hex in message verify

fileninja:
as the title says. resolving #206

### DIFF
--- a/client/app/src/accounts/AccountController.js
+++ b/client/app/src/accounts/AccountController.js
@@ -1936,18 +1936,9 @@
         var message = $scope.verify.message;
         var publickey = $scope.verify.publickey;
         var signature = $scope.verify.signature;
-        var res = accountService.verifyMessage(message, publickey, signature);
+        var result = accountService.verifyMessage(message, publickey, signature);
         $mdDialog.hide();
-        var message = gettextCatalog.getString("Error in signature processing");
-        if(!res)
-        {
-          message = gettextCatalog.getString("Error in your Input.");
-        } else if (res == true) {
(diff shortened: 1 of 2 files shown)


client/app/src/accounts/AccountService.js
```diff
@@ -354,6 +354,12 @@
     };
 
     function verifyMessage(message, publicKey, signature) {
+      //check for hexadecimal, otherwise the signature check would may fail
+      var re = /[0-9A-Fa-f]{6}/g;
+      if(!re.test(publicKey) || !re.test(signature))
+      {
+        return;
```
fix: should return false instead

client/app/src/accounts/AccountController.js
```diff
@@ -1939,7 +1939,10 @@
         var res = accountService.verifyMessage(message, publickey, signature);
         $mdDialog.hide();
         var message = gettextCatalog.getString("Error in signature processing");
-        if (res == true) {
+        if(!res)
```
fix: not needed

client/app/src/accounts/AccountService.js
fileninja: Yeah i also Thinked about this but you can difference between unverified or wrong Input

fix: then i would rather return something else from the method.
like `{success:false, error:"error message detailing error"}`

fileninja: Yeah that would be better indeed. I will do.

fileninja:
here you go @fix

luciorubeens:
+5

Issue #211, opened 2017-08-17
Hint for invalid transaction timestamp


Jarunik:
Can we add hint that you have to sync computer time please?

luciorubeens:
Related: #57

Issue #57, opened 2017-03-27
Relay on Internet Time Rather Than Users OS Time


boldninja:
In order to prevent timestamp issues that users experience when executing TX, ARK Client could (with the help of API call) get and set internal clock from the internet (either from time.nist.gov or if possible get time from public node) when it loads up and if users has it loaded for longer periods of time update it like every hour, else it gets the time every time the ARK Desktop Client is started. If API call to get time cannot be reached (eg. server unavail) it would just switch to user time (what it currently uses).

PR #217, opened 2017-08-24
Relay on Internet Time

fileninja:
Hey guys,
just implemented a small service to get the current time via json time service in the internet, returning the current american time. if it cant call this service for whatever reason it will switch to local time and returning a normal Date Object.
Also replaced all the new Date().getTime() Calls with this new Service.

#57 and #211 should be done with that.

### DIFF
--- a/client/app/src/time/TimeService.js
+++ b/client/app/src/time/TimeService.js
@@ -2,36 +2,25 @@
   'use strict';
 
   angular.module('arkclient')
-    .service('timeService', [TimeService]);
+    .service('timeService', ['$http', TimeService]);
 
   /**
    * TimeService
    * @constructor
    */
-  function TimeService() {
-    function __getJSON(url) {
-      return new Promise(function(resolve, reject) {
-        var xhr = new XMLHttpRequest();
-        xhr.open('get', url, true);
-        xhr.responseType = 'json';
-        xhr.onload = function() {
-          var status = xhr.status;
-          if (status == 200) {
-            resolve(xhr.response);
-          } else {
-            reject(status);
-          }
-        };
-        xhr.send();
-      });
-    };
+  function TimeService($http) {
     this.getTime = function() {
-      __getJSON('http://date.jsontest.com/').then(function(data) 
-      {
-        return new Date(Date.parse(data["date"]+" "+data["time"]));
-      }, function(error) {
-        return new Date();
-      });
+      $http.get('http://date.jsontest.com/', { timeout: 2000 })
+        .then(function(resp) {
+          var date = new Date(Date.parse(resp.data["date"]+" "+resp.data["time"]));
+          //GMT-4 New York Time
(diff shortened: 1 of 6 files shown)


client/app/src/time/TimeService.js
```diff
@@ -0,0 +1,35 @@
...
+    .service('timeService', [TimeService]);
+
+  /**
+   * TimeService
+   * @constructor
+   */
+  function TimeService() {
+    function __getJSON(url) {
+      return new Promise(function(resolve, reject) {
+        var xhr = new XMLHttpRequest();
```
luciorubeens: any specific reason to not use $http instead of xhr?

client/app/src/time/TimeService.js
```diff
@@ -0,0 +1,35 @@
...
+          }
+        };
+        xhr.send();
+      });
+    };
+    this.getTime = function() {
+      __getJSON('http://date.jsontest.com/').then(function(data) 
+      {
+        return new Date(Date.parse(data["date"]+" "+data["time"]));
+      });
```
luciorubeens: missing a catch(?).. if cant call this url returns local date

client/app/src/time/TimeService.js
fileninja: No, in this case its the same.. its just my Habit. Should i Change?

client/app/src/time/TimeService.js
fileninja: Should be handled by the “then“ Error function but will Test again.

client/app/src/time/TimeService.js
luciorubeens: yes, please. it's better to keep the same standard as other services

fileninja:
@luciorubeens changed to $http and checked the local time fallback which worked fine. So actually it returns GMT-4 (New York Time) which you may want to change depending on the default network time. If it doesnt reach the server for whatever reason, it will fallback to clientside time

client/app/src/time/TimeService.js
```diff
@@ -0,0 +1,24 @@
...
+   * @constructor
+   */
+  function TimeService($http) {
+    this.getTime = function() {
+      $http.get('http://date.jsontest.com/', { timeout: 2000 })
+        .then(function(resp) {
+          var date = new Date(Date.parse(resp.data["date"]+" "+resp.data["time"]));
+          //GMT-4 New York Time
+          date.setHours(date.getHours() - 4);
+          return date
```
fix: mmm the issue here is that it return a promise and not the real date, also date should be UTC

fileninja: What do u mean with promise?
Every Server which is Not in your Hand is a promise. Gmt is actually the same as utc but the URL can also be changed to a utc one

luciorubeens:
Solved #220

PR #218, opened 2017-08-26
Save/Import signed transaction file

### LINKED ISSUE #213: Add "Broadcast Offline Generated TX" window
Can you please add feature to be possible to sign transaction on offline computer and send it signed online? It will give much more security level.

naaomimore:
Basic implementation of a new feature to save the signed transaction to json file and import to broadcast.

Fixes #213


luciorubeens:
nice, +5.
but it can be more resourceful, like a option to save signed or unsigned

PR #220, opened 2017-08-27
57 timeserver update

SBero:
#57 / #217 - Here's my implementation for this that I did last week. It's different than the other one in that the other one doesn't actually account for server latency from the time the timestamp is retrieved from the server.

### DIFF
--- a/client/app/src/network/NetworkService.js
+++ b/client/app/src/network/NetworkService.js
@@ -2,13 +2,13 @@
   'use strict';
 
   angular.module('arkclient')
-    .service('networkService', ['$q', '$http', '$timeout', 'storageService', NetworkService]);
+    .service('networkService', ['$q', '$http', '$timeout', 'storageService', 'timeService', NetworkService]);
 
   /**
    * NetworkService
    * @constructor
    */
-  function NetworkService($q, $http, $timeout, storageService) {
+  function NetworkService($q, $http, $timeout, storageService, timeService) {
 
     var network = switchNetwork(storageService.getContext());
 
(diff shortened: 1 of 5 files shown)


fix:
this solution looks better.

I would just make 2 comments:
- timeout to connect to time server should be taken care off (otherwise the app can feel very unresponsive when disconnected from internet)
- make an estimation of time shift with current time and apply this shift when cannot connect to time server api

SBero:
I believe this should take care of both of these now. Added a 2s timeout.

client/app/src/network/NetworkService.js
```diff
@@ -176,32 +181,42 @@
...
+              'nethash': network.nethash
+            },
+            timeout: 5000
+          }).then(
+            function(resp) {
+              timeService.getTimestamp().then(
+                function(success){
+                  deferred.resolve(resp.data);
+                  peer.isConnected = true;
+                  peer.delay = timestamp - peer.lastConnection.getTime();
```
luciorubeens: success or timestamp? 🤔

client/app/src/time/TimeService.js
```diff
@@ -0,0 +1,60 @@
...
+    .service('timeService', ['$q', '$http', '$interval', TimeService]);
+
+  /**
+   * TimeService
+   * @constructor
+   */
+  function TimeService($q, $http) {
+
+    
+    var timeServerUrl = "http://www.timeapi.org/";
```
luciorubeens: timeapi.org refused connection

client/app/src/network/NetworkService.js
```diff
@@ -176,32 +181,42 @@
...
-          peer.isConnected = true;
-          peer.delay = new Date().getTime() - peer.lastConnection.getTime();
-          connection.notify(peer);
-        },
-        function(resp) {
-          deferred.reject("Peer disconnected");
-          peer.isConnected = false;
-          peer.error = resp.statusText || "Peer Timeout after 5s";
-          connection.notify(peer);
+      timeService.getTimestamp().then(
```
luciorubeens: this will request the time server on each peer call, so it would be better to get timestamp directly from peer response because the network loop pick up the transactions list every 10 seconds

SBero: Actually looking at this more closely, I don't believe this change is required at all. The peer.delay is simply the latency to make the peer connection. Shouldn't need any server timestamp and local timestamp works fine.

client/app/src/network/NetworkService.js
SBero: Should be timestamp, but not needed anymore (See previous comment).

client/app/src/time/TimeService.js
SBero: Swapped this service with another one. Oddly enough, my development machine can get a response but my other one doesn't? No idea why that's happening.

SBero:
@luciorubeens This should resolve those findings in the code. 

- peer.delay server timestamp isn't actually needed upon looking at the code again.

- The timeserver I'm able to access on my dev machine for some reason, on another one I can't, and doing a sanity check with a service says its up. So no idea why there's the issue. Swapped out for a new service.

luciorubeens:
nice! 👍 +5

but ark-js also need improvement, to accept custom timestamp in transaction creation

boldninja:
@SBero Can you provide me with your ARK address for bounty distribution?

PR #225, opened 2017-08-31
Second Passphrase Creation Workflow

darfraider:
Added a new workflow to the Second Passphrase creation process.  Users were not notified of the 5 Ark fee for the creation of the second passphrase.  Now when user starts the process, a dialog informs them of this fee and they have the option to cancel/exit or click next to proceed with the process.


client/app/src/accounts/view/createSecondPassphraseConfirmation.html
```diff
@@ -0,0 +1,34 @@
...
+        </md-button>
+      </div>
+    </md-toolbar>
+    <md-dialog-content>
+      <div class="md-dialog-content">
+        <md-card md-colors="{'background': 'red-100'}" flex="100">
+          <md-card-title>
+            <md-card-title-text>
+              <span>
+                <translate><center>Second passphrase fee is <b>5 Ark</b>.  If you wish to continue, click NEXT.</center></translate>
```
boldninja: Can you make it read global constants rather than setting it hardcoded? We'll be changing fee structure soon.

darfraider: What's the best place to put that.. in app.js?  I see const in there already.  

In addition, I noticed that the in AccountService.js there is also hard-coded values for the second passphrase (in satoshis i.e. if (account.balance < 500000000) {...).  I can update this to globalize the values because not only is that value hard-coded, but in other areas as well. For example, extra delegate is 25 Ark but hard coded as if (account.balance < 2500000000) {

boldninja: I'm not coder just thinking out loud and looking for future releasing.

@luciorubeens What do you think?

luciorubeens: need to get fees from current peer, the api endpoint: https://node1.arknet.cloud/api/blocks/getfees

luciorubeens: just change `Ark` to current network token: `{{ul.network.token}}`

luciorubeens: @darfraider don't need a new dialog to confirm, just send the transaction to `validateTransaction` method in accountController

darfraider: validateTransaction only checks to ensure the user has enough Ark in their account.  But it doesn't warn them that they are about to spend that much right?  I've asked a few people that run 1.3.1 and they said the process went through and charged them but they had no clue it was going to be 5 ARK otherwise they would have not created the second passphrase.

darfraider: Nevermind -- I'll build the master branch because this must be new code (not in 1.3.1).  I'll look at leveraging the existing validateTransaction function.

zillionn: Why don't you just write the amount inside the create button:
CREATE (Ѧ5)      CANCEL

PR #230, opened 2017-09-02
Price ticker pre mainnet

gsilvestri:
Branch con i price ticker aggiornati pre MainNet.


gsilvestri:
Sorry: wrong repository.

This is a pull request into kapucoin repository: I don't understand why it has gone into ArkEcosystem reposity.

Delete at the first possible occasion.

My bet.

PR #234, opened 2017-09-05
Decreased initial window size, created minimum window size

krau612:
The window can no longer be made so small that nothing can be seen. Also setup a smaller initial window size (previously it was almost the whole screen).


luciorubeens:
Hi @krau612. Hmm, this way is not dynamic, see how it starts on my screen:

captura de tela 2017-09-19 as 11 24 22

it looks bad

luciorubeens:
Fixed it in #257

Issue #61, opened 2017-04-01
sending tx to self shows red instead of gray 


cannabanana:
small bug not super important.  sending tx to yourself shows as red on the desktop wallet but should be gray.

fileninja:
the send ark button is always red.. not just for sending to yourself.
So what should be done here?

luciorubeens:
@fileninja is on transactions list, total column

luciorubeens:
Fixed in #235

PR #235, opened 2017-09-05
Sending ARK to yourself now shows up as grey, not red

krau612:
Fix for #61. Sending ARK to yourself now shows up as grey instead of red.
Also fix for #259. Total now shows amount sent and amount received (ex. "-1.1 + 1") so that you can see how much you sent and got back.


luciorubeens:
+5 👍🏻

Issue #232, opened 2017-09-04
KRW display


infinitytn:
Hello

Just saw new version support KRW and korean. Thanks.

1. Comma add to KRW.
In KRW we use comma(,) in every 3 digits. 
E.g) 1,000,000

2. Decimals in KRW
We dont use decimals so just round up is fine.

Thanks

luciorubeens:
Fixed in 606d2d3de8fbcfac74c4a38746551f3acc98099b

PR #242, opened 2017-09-05
Acccount currency balance will now have comma separators

ckhatri:
Inspired by #232. Will add commas as separators.

Can be expanded to have a mapping for each currency to its locale string, and then use that so it's right for any language! 

Just wanted to get this out there really quick and see if y'all would be OK with doing that? If so I could add it!

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -20,14 +20,14 @@
     body {
       background-color: #000
     }
-    
+
     html {
       border-radius: 10px;
     }
     /**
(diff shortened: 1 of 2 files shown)


luciorubeens:
hi @ckhatri, we already have the language/currency mapping: [account.controller.js#L48](https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/account.controller.js#L48)

merged manually, good solution! thanks +5 👍🏻

Issue #244, opened 2017-09-06
Localisation: Contact Address Hover


Spikarija:
There is an error upon mouse hover on the contact addresses list where the popup on top of "i" symbol is displayed in English (Income:.... Expend:...) even if another language is chosen. Attached and emphasized in red.
![popup](https://user-images.githubusercontent.com/31689626/30103361-b45f3458-92f2-11e7-8275-f7736f043180.jpg)

fix:
yes aside from the fact that the words are mismatched:
expend -> income
income -> expense

ckhatri:
Went ahead and fixed that in my PR!

luciorubeens:
Fixed in #246

PR #246, opened 2017-09-06
Fixes transaction not translating next to income/expend issue #244

ckhatri:
Was looking at issue number #244 and noticed that transactions wasn't translating so I fixed that. 
Looks like the issue with the word income and expend is that there are no translations in translations.js so those would need to be added?

Wondering where we could verify translations before adding those keys and their translations to each language! Or how to best to proceed to get those translations and then I'll add 'em!


luciorubeens:
+5

Issue #248, opened 2017-09-08
currency and languages are not persistent : ArkWallet 1.3.2


pimoussefrnl:
Hi, 

i found that the currency and language are not persistent (randomly) if you change one of the settings.

Example Mainnet:
Account is in French + Eur => Switch to BTC restart Ark Client : Back to Eur
Account is in Dutch + Eur => Switch to USD restart Ark Client: Back to Eur
Account is in Dutch + Eur => Switch to HKD + Polish, restart: Back to Dutch and now HKD

Example Devnet:
Account is in English => switch to Bulgarian, restart Ark Client => Back to English

- tested with W10 x64 on two differents computers
- Reproducible and can be recorded if needed
- Same issue mainnet or devnet
- Connected to internet or not doesn't change the issue

SBero:
I've looked into this a little bit. It's bizarre in that sometimes it happens and sometimes it doesn't as if I make the change and the app hasn't written it to the disk or something.

Possibly related to this? https://github.com/electron/electron/issues/8416

fix:
well this is normal. The way it works is the following:
- each time you store value (like changing settings), the storageService store it in-memory (for fast write/read)
- every 10s or so, the status of the storage is written to localStorage (for persistence). see https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/storage/StorageService.js#L79

So if you close client before 10s after changing some settings, it might happen that it has not been persisted. We should somehow hook into the "close client event" to force the storageService to persist data.

SBero:
Should be fixed now with my PR.

luciorubeens:
Fixed in #253

PR #253, opened 2017-09-11
248 saving bug

SBero:
Made a few consistency updates to the Currency and added a Window flush state on window close. #248


luciorubeens:
hmm this has not changed anything in currency, I think you meant the language..
merged manually, thanks! +10 👍🏻
please make a separate PR next time..

Issue #252, opened 2017-09-11
disable screenshot protection should show "disabled" when clicked


wilruns:
as title

ckhatri:
Good catch! Fixed it.

leimig:
Shouldn't this thread be closed? Since it's fixed.

ckhatri:
My pull request hasn't been merged so it'll stay open until then.

leimig:
Ah, I see :)

luciorubeens:
Fixed in #255

PR #255, opened 2017-09-11
adds update screenshot protection function fixes issue #252

ckhatri:
Let me know if y'all want anything changed!

https://github.com/electron/electron/issues/2717

According to this they don't do dynamic updating, and so this is the way to update the menu items at runtime


luciorubeens:
nice! +5 👍🏻

PR #257, opened 2017-09-12
Update to maintain window state upon close #256

### LINKED ISSUE #256: Feature Request: Store the clients window size on close
On starting the Ark Client the window opens up really big. The client should capture and store the window dimensions before it closes and then reapply the settings on the next use.

SBero:
#256 Feature Completed.


luciorubeens:
great! +5 👍🏻

PR #265, opened 2017-09-13
Fixed issue where contact with same address as selected address could…

krau612:
… not be edited.

Creating a contact with the same address as the opened account would create weird behavior. The user would not be able to edit the contact (or delete it) because it was the same address as the account. Now, it is possible to create a contact with the same address as the account, edit that contact, and delete that contact.


luciorubeens:
Merged manually! +5 👍🏻

PR #231, opened 2017-09-03
added validateTransaction method support for createSecondPassphrase 

darfraider:
- added validateTransaction method to support the createSecondPassphrase method.  This method originally did not show the dialog to allow user to see the fee associated with this creation process and would blindly charge the user.
- there were some inconsistencies in createSecondPassphrase with the interchangeable use of 'account' and 'selectedAccount'.  This has been made consistent in this method to reflect the same use across other methods in the controller.

PR #271, opened 2017-09-16
Added Second Passphrase Confirmation Dialog to warn uses about fee

krau612:


luciorubeens:
merged a previous PR #231, so we have now the same confirmation dialog for each transaction  (transfer, vote, register delegate and register secondpassphrase) before broadcast, but if you want to add the cost info in the register dialog you need to get fees from peer because soon the values will change..

endpoint: https://node1.arknet.cloud/api/blocks/getfees

PR #272, opened 2017-09-16
Improve Create Second Passphrase Dialog

leimig:
- Prevent users from submitting Second Passphrase Dialog before filling it
- Create checkbox to make sure users know about the 5 ARK fee
- Translate some texts to pt-BR


Trauzki:
Yes, please! I spent 5 Ark that I didn't want to spend.

luciorubeens:
merged a previous PR #231, so we have now the same confirmation dialog for each transaction  (transfer, vote, register delegate and register secondpassphrase) before broadcast, but if you want to add the cost info in the register dialog you need to get fees from peer because soon the values will change..

endpoint: https://node1.arknet.cloud/api/blocks/getfees

PR #279, opened 2017-09-20
Update nl.po

mcactive:
Updated Dutch translations to make it more consistent:
- use of informal "je" instead of formal "u"
- use of "passphrase" instead of "wachtwoord"
- some improvements based on the usage of the translations in the app
- changed some wrong translations


fix:
@mcactive please use http://osjc1wl.oneskyapp.com/collaboration/project?id=95031 for translations instead

PR #281, opened 2017-09-20
Add live reload to Electron workflow

j-a-m-l:
This PR adds something that I find very useful:
the `npm run-script live-reload` command will listen to changes in the project tree and will reload the Electron app.


luciorubeens:
added cross-env package to fix environment on Windows, and index.html to resources to fix live-reload on Mac..

+5! 👍🏻

PR #286, opened 2017-09-22
Confirm Unvote transaction before Voting

### LINKED ISSUE #86: Switching Votes is cumbersome
It is too much effort to switch your vote from one delegate to another.

Currently:
1) Unvote
2) Wait 
3) Vote 

It is not clear how long you have to wait. It would be good to see somehow when the unvote is done and you can vote again.

It would be even better if you could unvote and vote someone else at the same time (like the voting list indicates but does not let you do it).

alexbarnsley:
This is designed for the purpose of voting for a new Delegate, while still voting for someone else.

It's the same process when already voted, just with a dialog:

1. Add second delegate
2. Untick currently voted delegate
3. Vote
4. Sign unvote transaction
5. Send
6. Wait for Unvote to confirm
7. Sign new vote transaction
8. Send

Resolves #86

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1041,7 +1041,7 @@
           }
         });
         if (toVote.length > 1) {
-          formatAndToastError('You can only vote for 1 delegate');
+          return formatAndToastError('You can only vote for 1 delegate');
         }
 
         var performVote = function() {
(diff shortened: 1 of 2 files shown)


luciorubeens:
It works like a charm, but I'll not merge it since the next protocol update will fix this to the same transaction. Good job!

alexbarnsley:
No worries! I thought that would be the case after AIP 11 was announced 😁 do you think it's worth implementing the wait screen for transactions to be confirmed anyway, possibly as an option when making the transaction?

PR #287, opened 2017-09-22
Pressing refresh button now gives user a response

ckhatri:
For the refresh button on the home page next to my accounts, when you click it now there isn't any response 

For user experience reasons I believe it'd be better that the icon, when clicked, gives some response back to the user, even if the amounts don't update. 

Not sure if I need to create an open issue for this, but the idea is simple and I think this would lead to a better UX.


luciorubeens:
+5

PR #290, opened 2017-09-23
Toggle screenshot protection status, with warning

### LINKED ISSUE #103: Disabling screenshot protection does not allow for re-enabling.
A user who disables screenshot protection does not receive the option to enable it going forward. To reactivate screenshot protection the wallet must be reinstalled.

alexbarnsley:
Nothing glamorous. I didn't look at translating the title text as it doesn't do it at the minute, nor does it translate the menu item label. If someone could give guidance, I could give that a try too. Assuming it's possible with the current translation system?

Resolves #103

### DIFF
--- a/main.js
+++ b/main.js
@@ -17,6 +17,7 @@ const fork = require('child_process').fork;
 let mainWindow
 
 var ledgercomm
+var screenProtected = true;
 
 function createWindow () {
     // Create the browser window.t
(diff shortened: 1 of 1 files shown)


alexbarnsley:
It looks like this has already been done based on the conflicts. #103 can be closed in that case

PR #293, opened 2017-09-24
Show the button to toggle the list of accounts on the dashboard too

j-a-m-l:
Show, on the dashboard, the accounts (and contacts) list button when there is at least 1 account or 1 contact, or the Ledger is connected.

Previously it was necessary to had selected an account. Since account are unlock when they're imported, that didn't provide any special security measure.

That previous behaviour, forced users to select an account to see their contacts (which are global, not per account), which doesn't make any sense.

In fact, this change permits a more secure way to watch the balance of an account:
- I could add my addresses as contacts, without unlocking them, and see their balances easily (other wallets have a similar feature, like "Watch addresses"). It would be like a read-only access.

Apart from this change, it modifies the HTML structure, so the sidebar would be displayed on wide screens.

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -120,6 +120,7 @@ md-sidenav md-list {
 
 #content {
     overflow: scroll;
+    background-color: transparent;
 }
 
 #content {
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -59,6 +59,7 @@
   </div>
 
   <div ng-cloak="" style="height: 100%" layout="column">
+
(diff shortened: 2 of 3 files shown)


j-a-m-l:
@luciorubeens I've fixed all the conflicts

luciorubeens:
@j-a-m-l I want to create a new dashboard with the contacts, the latest transactions, price chart, network/peer status, income/expend, notifications.. We will ask our design to do something, so for now I'm closing this.

PR #295, opened 2017-09-24
Avoid duplicated contacts or contacts that are the same that user addresses

j-a-m-l:
These changes avoid using the same contact name several times. It also avoids creating contacts that have the same name or address than any user address.

BTW: these errors don't seem to be translated.


luciorubeens:
+5 👍🏻

PR #296, opened 2017-09-24
Add Delegate - Feedback when no internet connection

alexbarnsley:
When no internet connection is available, choosing to "Add Delegate" doesn't show the dropdown. If it's not obvious that internet connection isn't available, it looks like a UI glitch. Initially reported on Reddit:

https://www.reddit.com/r/ArkEcosystem/comments/7210ck/voting_problem_in_client/


luciorubeens:
Nice! +5 👍🏻

PR #301, opened 2017-09-25
Network name color fix

SBero:
Fix for the network text color on the dark theme.


luciorubeens:
needs to highlight the current network

PR #302, opened 2017-09-25
now when you hover over the icons in wallet page, will show text

ckhatri:
Currently if you hover over the send button and edit contact button it doesn't show any text, but the export account does. This fixes it so the word send and edit contact will show over the hovers of the icons.


luciorubeens:
good catch! +5

PR #304, opened 2017-09-26
add spanish as a language choice

ckhatri:
Noticed that Spanish wasn't added as a language choice, and it's fully completed here 
http://osjc1wl.oneskyapp.com/collaboration/project?id=95031

Now you can select Spanish, the translations work from what I saw. 

I put it under Greek since the code for that is el: and I noticed that the languages was in alphabetical order but English.


luciorubeens:
oh, true that.. +5 👍🏻

PR #305, opened 2017-09-26
Update balance

ckhatri:
Noticed that the balance doesn't use the method written earlier called myAccountsCurrencyBalance which makes the code a bit cleaner. So updated to use that. Initially changed it from being a button to not a button, but changed it back to being a button since it creates space and looks a bit nicer.

Reason I didn't also change the Ark value to use the method is because I think having the exactish ark in your wallet somewhere is nice. Having the localization for the specific currency keeps it nice and uniform throughout the program though!


luciorubeens:
oh yeah! agree +5

boldninja:
@ckhatri can you provide me with your Ark address?

PR #309, opened 2017-09-29
Add 'use strict' into all services/controllers

noahgr:
Hey guys, I noticed that we don't have strict-mode set for all services/controllers. Now we will be able to handle some errors in `account.controller`.

Cheers.


boldninja:
@noahgr can you provide me your ARK address? (p.s. : PRs merged on 1st of October will count towards October)

PR #311, opened 2017-09-29
Remove recipient address and change type to label on the show transaction

### LINKED ISSUE #307: Delete 'To' Address When Voting
When you are voting you see From and To going to same address so might be confusing for people just starting with ARK as it basically goes to the forger that will include voting tx in the block:

votinggg 1
So maybe if type == '3' hide 'To' field
also we could maybe rename Types into what they actually are in client eg. when sending its type = 1 so might be more UI friendly if types instead of numbers in client say what it is (Sending, Voting, Registering 2nd pass, Registering a delegate).

noahgr:
Hi, this should resolve #307. I think the recipient address need to show only in `type == 0`.

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -2159,6 +2159,7 @@
         send: send,
         cancel: cancel,
         transaction: transaction,
+        label: accountService.getTransactionLabel(transaction),
         // to avoid small transaction to be displayed as 1e-8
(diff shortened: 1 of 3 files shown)


boldninja:
That was quick thanks :D 
@luciorubeens can you review and include before next version release?

luciorubeens:
Wow, hat-trick! +15 due your previous PR.

Thanks.

PR #314, opened 2017-10-01
Get fees from api before create the transaction

noahgr:
Hi. I read the blog and saw that we will change the fees soon and I noticed that the `account.service` uses fixed values to validate and create a transaction.

In api we have this route: https://ark.brianfaust.me/#/Block/blocks_getFees.

I get the fees from it but I made a default variable in case the user is not connected.

Please let me know if you need any changes.

### DIFF
--- a/client/app/src/accounts/account.service.js
+++ b/client/app/src/accounts/account.service.js
@@ -16,6 +16,14 @@
 
     var ark = require('arkjs');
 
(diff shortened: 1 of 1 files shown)


luciorubeens:
Finally! I had already commented about it in other PR's.. #272

+5 👍🏻

PR #315, opened 2017-10-01
More testing infrastructure (closes #297)

### LINKED ISSUE #297: Improve the testing infrastructure
Currently there are only 4 tests and there isn't any kind of mention anywhere.

I'd like to change that, so with those tests:

- Explain how to start with tests on README
- Change CONTRIBUTING to emphasise the utility of adding tests
- Split the current `test/index.js` into helpers/utils/page-objects to reuse some parts
- Add some additional tests
- Configure a CI

What do you think?

j-a-m-l:
This is related to issue https://github.com/ArkEcosystem/ark-desktop/issues/297, and continues the changes started at https://github.com/ArkEcosystem/ark-desktop/pull/308.

Notes about these changes:
 - I've used the `should` syntax because the previous tests were using it. Maybe it's a good time to change it to something less intrusive (`assert` or `expect`).
 - I've added Karma to have a simpler and faster way to tests Services, Filters, etc. instead of using E2E for all things. It's using PhantomJS, but a newer browser would be a better option.
 - Network (and other things) should be mocked on E2E tests, but, since that is a quite time-consuming thing, I think that it would be better creating new tasks for it or doing it incrementally.
 - I've added a sample Travis configuration, but, since you're going to configure the project (and I don't know if you've your own preferences, workflow, CI infrastructure) I delegate the final job to you.
 - I've added Devtron too because some people could find it useful


spresnal:
Are no semicolons the standard?

j-a-m-l:
Since, this project doesn't use any linter or enforces any code style yet, I've omit some unnecessary semicolons.

In fact, @spresnal I'd like to use a real (de facto) standard way: https://github.com/ArkEcosystem/ark-desktop/issues/323 .

luciorubeens:
Great! easily legible and well structured +10 👍🏻

PR #316, opened 2017-10-04
Issue 194 -- added Developer Mode

darfraider:
Added Developer Mode in Settings

Loads by default in **Disabled** mode:

image

Enabling **Developer Mode** enables the "Switch Network" button on toolbar to allow user to switch networks:

image

Settings are retained between sessions using the Storage Service helper functions


boldninja:
Hmm don't know if Developer mode is correct term here since you can add custom networks here as well. Maybe 'Custom Network Mode' or 'Additional Networks Mode' ? Thinking out loud since you can add for instance Kapu or Blockpool network in ARK client.

fix:
should be "advanced mode" and having an angular filter "advancedMode" would be great to filter out components for advanced users when switch off.

boldninja:
Yeah 'Advanced Mode' is even better.

fix:
also translation file should be left out the pr

darfraider:
1. Should I backtrack here and pursue the filter method or update text to read Advanced Mode?

2. Curious, why should the translation file be left out of the PR? How does it get updated in the release builds? Do I still update it in my local copy or unnecessary since I’m not committing it?

Issue #194, opened 2017-08-12
Add Dev Mode slider to wallet settings


blib423:
in the interest of simplification, a slider in the settings menu may be useful to turn developer functions on and off, similar to web browsers. By default the slider would be off, making TESTNET and DEVNET invisible to the wallet user, preventing confusion by accidentally clicking and changing networks on the main screen.

PR #317, opened 2017-10-04
Added Advanced Mode

SkyHyperV:
This is for feature request Issue #194...

Advanced Mode is **Disabled** by default

image

Enabling it allows the _Switch Networks_ menu item to appear.  We can use ng-if in div/menu tags to hide/show UI information.

image

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -59,7 +59,7 @@
         <md-tooltip>
           <translate>Switch Network</translate>
         </md-tooltip>
-        <md-menu>
+        <md-menu ng-if="ul.advancedMode">
           <md-button aria-label="Switch Network" ng-click="ul.openMenu($mdOpenMenu, $event)">
             <md-icon md-font-library="material-icons">wifi</md-icon>
           </md-button>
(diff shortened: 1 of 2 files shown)


darfraider:
Also, how do I maintain / update the translation for the string I added “Advanced Mode”?

spresnal:
Might be nice to wrap this up in a custom angular directive. Then if the functionality needs to be extended/modified it all lives in one place.

luciorubeens:
Great! but @spresnal is right, we need a directive. Perhaps broadcast a message when swipe the advanced mode and watch on directive.

darfraider:
Agreed. I’ll implement the directive and update the PR.

alexbarnsley:
I've not tested, but what happens if you're in devnet and then disable advanced mode? Also, I don't think this now complies with the Standards (semi-colons, etc). Worth getting the conflicts resolved and running `npm standard` to get an idea I think.

alexbarnsley:
Closing due to lack of a response.

PR #318, opened 2017-10-04
wrote a new format currency method which will localize any number passed in

ckhatri:
Currently My Accounts: Currency localization is different than the values shown in the Price and market cap section, so this method will take in any value and localize it so we have consistency.


client/app/src/accounts/account.controller.js
```diff
@@ -586,12 +586,30 @@
...
 
       var currencyBalance = Number(balance * price).toLocaleString(languageCode, options);
-      
+
       if (currencyName == "btc") currencyBalance = currencyBalance.replace("BTC", "Ƀ");
 
       return currencyBalance;
     }
 
+    self.formatCurrencyVal = function(val) {
```
luciorubeens: `formatCurrencyBalance` and `formatCurrencyVal` sounds very similar, can you merge both into a single function?

client/app/index.html
```diff
@@ -177,7 +177,7 @@
                   <b><translate>Price</translate></b>
                 </div>
 
-                <span class="md-secondary">{{ul.currency.symbol}}{{ul.connectedPeer.market.price[ul.currency.name]}}</span>
+                <span class="md-secondary">{{ul.formatCurrencyVal(ul.connectedPeer.market.price[ul.currency.name])}}</span>
```
spresnal: These look like good use cases for a custom angular filter

luciorubeens: yeah! I agree @spresnal

ckhatri:
create a myAccountsCurrencyBalance to get the value of the account in whichever currency, and then pass that into formatCurrencyVal to format it.

ckhatri:
Created the custom filter and implemented it!

ckhatri:
there was a merge conflict, so I went ahead and fixed it

ckhatri:
should be good to go now

luciorubeens:
nice +5 👍🏻

PR #325, opened 2017-10-09
Fix Application Menu to match Settings Menu

krau612:
Buttons in the Application Menu now fill the width of the menu, like the Settings Menu.


luciorubeens:
+5 👍🏻

PR #326, opened 2017-10-09
Set BTC to default currency and first in currencies list

krau612:
Previously, if a currency was not set, the currency option in the settings menu would be blank even though the currency defaulted to BTC. The currency still defaults to BTC, but it will now show in the settings menu that the currency is BTC. It also makes sense to make BTC first in the list as it's more universal than USD (which was first in the list).


luciorubeens:
nice +5 👍🏻

PR #288, opened 2017-09-22
Include ledger in balance total

alexbarnsley:
Resolves #239

PR #330, opened 2017-10-10
swapped method for getting accounts to fix amount display bug

liamreoch:
fixed the account amount bug by replacing the getAllAccounts() call with myAccounts()

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -600,7 +600,7 @@
     };
 
     self.myAccountsBalance = function() {
-      return (self.getAllAccounts().reduce(function(memo, acc) {
+      return (self.myAccounts().reduce(function(memo, acc) {
         return memo + parseInt(acc.balance || 0);
       }, 0) / 100000000).toFixed(2);
     }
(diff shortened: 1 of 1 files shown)


luciorubeens:
Hi @liamreoch, `getAllAccounts` was made by @alexbarnsley to include the amount of ledger accounts.. #288

client/app/src/accounts/account.controller.js
```diff
@@ -600,7 +600,7 @@
     };
 
     self.myAccountsBalance = function() {
-      return (self.getAllAccounts().reduce(function(memo, acc) {
+      return (self.myAccounts().reduce(function(memo, acc) {
```
alexbarnsley: Just a thought. Maybe change https://github.com/liamreoch/ark-desktop/blob/571d1a0c585f09eb3bd77c1bcbdcd8468711e337/client/app/src/accounts/account.controller.js#L586 to use `self.myAccounts()` instead of `self.accounts` (instead of changing this line). The reason being is that this will stop the balance for ledger wallets. Maybe give it a try 🙂

alexbarnsley:
Oh, just saw your comment @luciorubeens xD

alexbarnsley:
I don't think I considered virtual folders when getting the balance. I presume that's what is filtered out in `myAccounts()`

liamreoch:
Thanks guys, I overlooked that. I've put that method call back and it's now filtering out virtual folders.

luciorubeens:
sounds good! thanks +5

PR #335, opened 2017-10-10
Custom Backgrounds

### LINKED ISSUE #320: Have user be able to upload custom images for background appearance in wallet
You guys have a bunch of pre-selected images which are cool, it might be neat to let the user upload their own image.

krau612:
Allow the user to upload custom backgrounds. Resolves #320

screenshot at 2017-10-10 16-42-36

screenshot at 2017-10-10 16-53-10

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1534,7 +1534,7 @@
           filters: [
             { name: 'Images', extensions: ['jpg', 'png', 'gif'] }
           ],
-          properties: ["openFile"],
+          properties: ["openFile"]
         };
         var userPath = 'assets/images/user/';
         var dirPath = path.resolve(__dirname, userPath);
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1490,6 +1490,7 @@
       delete themes['dark'];
 
       var backgrounds = {
+        user: {},
         colors: {
           'Midnight': '#2c3e50',
           'Asbestos': '#7f8c8d',
@@ -1534,7 +1535,26 @@
         require('electron').remote.dialog.showSaveDialog(function(fileName) {
           if (fileName === undefined) return;
 
-          console.log(fileName);
+          var baseName = path.basename(fileName);
+          var newFileName = path.join(dirPath, baseName);
+
+          var readStream = fs.createReadStream(fileName);
+          readStream.on("error", function(err) {
+            console.log("Error Reading File: " + err);
+            return;
+          });
+
+          var writeStream = fs.createWriteStream(newFileName);
+          writeStream.on("error", function(err) {
+            console.log("Error Writing File: " + err);
+            return;
+          });
+          writeStream.on("close", function(ex) {
+            manageBackgrounds();
+            return;
+          });
+
+          readStream.pipe(writeStream);
         });
       }
 
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1575,6 +1575,8 @@
               .hideDelay(3000)
             );
           } else {
+            var name = path.parse(file).name;
+            delete backgrounds['user'][name];
             $mdToast.show(
               $mdToast.simple()
               .textContent(gettextCatalog.getString('Background Removed Successfully!'))
@@ -1582,7 +1584,6 @@
             );
           }
         });
-        manageBackgrounds();
       }
 
       function isImage(file) {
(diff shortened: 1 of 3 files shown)


client/app/src/accounts/account.controller.js
```diff
@@ -1523,6 +1528,93 @@
         }
       };
 
+      function upload() {
+        var options = {
+          title: "Upload Image",
+          filters: [
+            { name: 'Images', extensions: ['jpg', 'png', 'gif'] }
+          ],
+          properties: ["openFile"],
```
spresnal: Unnecessary comma here

client/app/src/accounts/account.controller.js
```diff
@@ -1523,6 +1528,93 @@
...
+        };
+        var userPath = 'assets/images/user/'
+        var dirPath = path.resolve(__dirname, userPath);
```
spresnal: Semicolon consistency (Need that linter ASAP @j-a-m-l )

client/app/src/accounts/account.controller.js
```diff
@@ -1523,6 +1528,93 @@
...
+            console.log("Error Writing File: " + err);
+            $mdToast.show(
+              $mdToast.simple()
+              .textContent(gettextCatalog.getString('Error Adding Background.'))
+              .hideDelay(3000)
+            );
+            return;
+          });
+          writeStream.on("close", function(ex) {
+            $mdToast.show(
```
spresnal: Probably a good idea to pull the toast logic into its own method where you pass in the text to be displayed

client/app/src/accounts/account.controller.js
alexbarnsley: I personally don't see a problem with this. It's valid syntax [see here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas). Also, it means it's 1 less line on a diff when an additional value is added at the end of the object.

spresnal: It's valid syntax, however, the linter which we are moving to (StandardJS) flags it by default. See **Trailing commas not allowed** here: https://standardjs.com/rules.html#javascript-standard-style

That being said, I do kind of like how it keeps the diff a bit cleaner.

spresnal: The syntax is valid, however, the linter which we are trying to move to (StandardJS) flags it by default. See **Trailing commas not allowed.** here: https://standardjs.com/rules.html#javascript-standard-style

That being said, I do kind of like the idea of keeping the diff a bit cleaner.

spresnal: The syntax is valid, however, the linter which we are trying to move to (StandardJS) flags it by default. See **Trailing commas not allowed.** here: https://standardjs.com/rules.html#javascript-standard-style

That being said, I do kind of like the idea of keeping the diff a bit cleaner.

alexbarnsley: I'm all for a standard, so in that case I agree 😄 

I've always done trailing commas on arrays to save on the diffs. I can't argue with standards though xD

krau612: I went ahead and removed the comma. I wasn't aware of the StandardJS style (and honestly I put it there on accident lol). Thanks guys!

client/app/src/accounts/account.controller.js
krau612: Added the semicolon. That was my bad. Thanks!

client/app/src/accounts/account.controller.js
krau612: Thanks for the feedback. I went ahead and made a showToast function. Now the toasts are all made with either showToast or formatAndToastError. I also removed my unnecessary log statements. Thanks!

fix: there are already a few Toast function, but it is not rational:
- https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/account.controller.js#L278
- https://github.com/ArkEcosystem/ark-desktop/blob/master/client/app/src/accounts/account.controller.js#L338

Would be nice to revamp the toast management into a single identified service instead

fix: would be for another PR

luciorubeens:
Nice, I'll upload many gifs 😁
Upload is ok, but the ui should be improved:

- Hide the upload button in the themes tab.
- Delete button must be next to each image (perhaps an icon near the top), so you don't need to select the image first to delete.
- After delete/upload don't call the `manageBackgrounds`, just update the array.
- After delete, update the background to default.

krau612:
@luciorubeens Thanks for the suggestions! I went ahead and implemented all of them. One issue though is that when clicking the delete button, which is now on the image grid tile, the image is selected for a split second before it is deleted. Do you know how I could fix this?

luciorubeens:
Hmm I did not noticed but I added `preventDefault()`.. good job! 👍🏻 +5

PR #338, opened 2017-10-11
Use single AddressBookController for account view

alexbarnsley:
This is to fix initial add of contact. If you guys could test please, that'd be great!

Edit: to clarify, the issue is when a contact is first added to the client. Removing and re-adding doesn't work. Complete reset of data is required to replicate the issue (prior to my changes). After resetting the data, click on the "Add Contact" button in the left sidebar and add a new contact. It won't appear until a reload

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -379,11 +379,11 @@
     </div>
   </div>
 
-  <div class="account-view" ng-if="ul.selected" layout="column" flex>
+  <div class="account-view" ng-if="ul.selected" ng-controller="AddressbookController as ab" layout="column" flex>
     <md-content flex layout="row">
 
       <md-sidenav ng-click="ul.toggleList()" md-is-locked-open="$mdMedia('gt-md')" md-component-id="left" class="md-whiteframe-z2">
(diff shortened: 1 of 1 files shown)


fix:
nice, also i have found tricky not to be able to rename a contact. I think that should be possible:
- either deny rename if a contact with same name exists
- either use a hidden id key in contact model so you can have same name for different contacts

I think first option should be ok (as contact name act as an identifier in the mind of users)

alexbarnsley:
No worries, I'll take a look! Shall I roll this into the same PR?

fix:
yes make another PR, so we can reject the new one more easily, and still accept this one ;)

alexbarnsley:
I'll make another PR which will be so perfect, you couldn't possibly reject 😉

luciorubeens:
+5!

boldninja:
@spresnal can you provide me with your Ark address - I'd like to give you a small token of appreciation for helping on providing your valuable input on a lot of these PRs and a mention in our monthly github blog post.

alexbarnsley:
👌

PR #339, opened 2017-10-12
Fixes for virtual folders

alexbarnsley:
This covers 2 issues:

- Could not rename a virtual folder at all. The option did nothing.
- Prevents adding the same foldername twice, showing a message to that effect.


luciorubeens:
Nice! +5 👍🏻

PR #341, opened 2017-10-13
Stop Import/Create account if no network

alexbarnsley:
This fixes an issue with not being connected to a network, when importing or creating an account it just does nothing with no feedback at all. I've implemented a toast notification when clicking "Import Account" or "Create Account" - see below. Feel free to test the before and after of not being connected to a network to see the issue I'm referring to.

image

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -871,15 +871,6 @@
       storageService.set('playFundsReceivedSound', self.playFundsReceivedSound, true);
     }
 
-    self.dialogCheckConnected = function(scope, e) {
-      if (!self.isNetworkConnected) {
-        formatAndToastError(
-          gettextCatalog.getString('You cannot do that until you connect to a network.')
-        );
-        self.cancel();
-      }
-    }
-
     /**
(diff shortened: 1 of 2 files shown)


fix:
well the thing it SHOULD be possible to create/import account while offline.

That is because this function fails https://github.com/alexbarnsley/ark-desktop/blob/c46e25d162f5e2cc1dd37bd3084a4c1d286b4524/client/app/src/accounts/account.service.js#L158 if not connected.

Maybe add a fail function where the account is still created/imported

or here fails gradefully (ie `deferred.resolve`) if no network https://github.com/alexbarnsley/ark-desktop/blob/c46e25d162f5e2cc1dd37bd3084a4c1d286b4524/client/app/src/accounts/account.service.js#L83

alexbarnsley:
Sounds good to me @fix, I'll make the amends :)

alexbarnsley:
@fix I've made a change so now it adds an "empty" account if the getFromPeer request fails. Please could you give it another test, thanks!

luciorubeens:
Nice! 👍🏻 +5

PR #347, opened 2017-10-14
Rename Contact

alexbarnsley:
Ability to rename a contact already in the account. Moved some stuff around so I didn't create any smelly code.

### DIFF
--- a/client/app/src/components/addressbook/addressbook.controller.js
+++ b/client/app/src/components/addressbook/addressbook.controller.js
@@ -67,24 +67,20 @@
       self.getContacts();
       if (self.trim(name) == "") {
         self.showToast('This Contact Name is not valid', name, true);
-        console.log('This Contact Name is not valid', name, true);
         return;
       }
       if (self.trim(address) == "" || ! self.isAddress(address)) {
         self.showToast('This Contact Address is not valid', address, true);
-        console.log('This Contact Address is not valid', address, true);
         return;
       }
 
       var newContact = { name: name, address: address };
       if (!isRename && self.addressExists(address)) {
         self.showToast('A Contact with this Address already exists', address, true);
-        console.log('A Contact with this Address already exists', address, true);
         return;
       }
       if (self.contactExists(name)) {
         self.showToast('A Contact with this Name already exists', name, true);
-        console.log('A Contact with this Name already exists', name, true);
         return;
       }
 
@@ -96,12 +92,10 @@
 
       if (existsIn(knownAccounts.map( a => a.address ), address)) {
         self.showToast('This Address is already taken, please add a different one', address, true);
-        console.log('This Address is already taken, please add a different one', address, true);
         return;
       }
       if (existsIn(knownAccounts.map( a => a.username ), name)) {
         self.showToast('This Name is already taken, please use a different one', name, true);
-        console.log('This Name is already taken, please use a different one', name, true);
         return;
       }
 
--- a/client/app/src/components/addressbook/addressbook.controller.js
+++ b/client/app/src/components/addressbook/addressbook.controller.js
@@ -63,6 +63,58 @@
       }
     }
 
+    self.addressBookAddContactRecord = function(name, address, callback, isRename, suppressNotice) {
+      self.getContacts();
+      if (self.trim(name) == "") {
+        self.showToast('This Contact Name is not valid', name, true);
+        console.log('This Contact Name is not valid', name, true);
+        return;
+      }
+      if (self.trim(address) == "" || ! self.isAddress(address)) {
+        self.showToast('This Contact Address is not valid', address, true);
+        console.log('This Contact Address is not valid', address, true);
+        return;
+      }
+
+      var newContact = { name: name, address: address };
+      if (!isRename && self.addressExists(address)) {
+        self.showToast('A Contact with this Address already exists', address, true);
+        console.log('A Contact with this Address already exists', address, true);
+        return;
+      }
+      if (self.contactExists(name)) {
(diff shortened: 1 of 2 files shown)


alexbarnsley:
@fix - I won't accept any rejections :trollface:

client/app/src/components/addressbook/addressbook.controller.js
```diff
@@ -63,6 +63,55 @@
...
+
+      var newContact = { name: name, address: address };
+      if (!isRename && self.addressExists(address)) {
+        self.showToast('A Contact with this Address already exists', address, true);
+        return;
+      }
+      if (self.contactExists(name)) {
+        self.showToast('A Contact with this Name already exists', name, true);
+        return;
+      }
```
spresnal: Would be nice to pull all this error handling out into a method. Then follow the error handling with a single return if possible.

Might consider adding some unit tests for the logic as well to make sure everything is being handled appropriately.

client/app/src/components/addressbook/addressbook.controller.js
```diff
@@ -63,6 +63,55 @@
...
+      self.save();
+      if (!suppressNotice) {
+        self.showToast('Contact successfully added', name, false);
+      }
+      if (typeof callback === 'function') {
+        callback();
+      }
+
+      return true;
+    }
```
spresnal: Can this method be broken into smaller unit-testable chunks? Do we need to return false when an error comes up above?

client/app/src/components/addressbook/addressbook.controller.js
```diff
@@ -157,25 +169,40 @@
...
+              self.contacts[i].address = address;
+              contactSaved = true;
+            }
+          }
+          if (contactSaved) {
+            self.showToast('Contact successfully saved', name, false);
+          } else {
+            self.showToast('Could not find the Contact to save', name, true);
+          }
+        } else {
```
spresnal: A lot of possible code paths here. Unit tests? 😈

alexbarnsley:
Thanks @spresnal - i'll implement some tests a bit later!

alexbarnsley:
@spresnal what are your thoughts on doing these tests? I'll basically need to have data in the "client" such as wallets and contacts already before performing my tests. Would I set this up at the start of each test? I don't really write many unit tests 😬

spresnal:
I wanted to put some example tests together, but ran out of time. Anyways, I'll try to answer your question:

>I'll basically need to have data in the "client" such as wallets and contacts already before performing my tests. Would I set this up at the start of each test?

The solution to this is stubbing out those method calls. We want to test code specifically within addressbook.controller.js and not the logic of any of the services being injected through the angular constructor.

You can either set that up a single time for all the tests (before) or before/after each (beforeEach/afterEach). It depends on the cost of the operation (in this case inexpensive) and whether the tests change the state of the object. For example, if a few tests are adding contacts to storage this may cause other unit tests to fail (in this case you'd want to reset the state of your storageService mock before/after each test). On the other hand, if you're only reading from storageService for all these tests you'd want to just use a before.

I would look into sinonjs (you may have to add it to the project) for stubbing out methods. Here's an example of stubbing an API call https://github.com/spresnal/league-of-legends-seed/blob/Basic-Setup/server/api/leagueApis/championMasteryApi.test.js. You can find a lot of other (better) examples with some searching.

Hope this helps!

alexbarnsley:
Thanks for the explanation @spresnal (I forgot to say). I've been trying to get the first service test working. It's close but whenever I instantiate accountService (or networkService), it does a `var ark = require('arkjs');` which I've got working, but that then does `require('./lib/transactions/crypto.js')` which fails:

```
npm test


> ArkClient@1.4.1 test C:\Users\alexj\projects\ark-desktop
> karma start test/karma.conf.js --single-run

22 10 2017 14:34:24.594:INFO [karma]: Karma v1.7.1 server started at http://0.0.0.0:9876/
22 10 2017 14:34:24.597:INFO [launcher]: Launching browser PhantomJS with unlimited concurrency
22 10 2017 14:34:24.603:INFO [launcher]: Starting browser PhantomJS
22 10 2017 14:34:27.565:INFO [PhantomJS 2.1.1 (Windows 8 0.0.0)]: Connected on socket 0pHKJ1v9dNdoibm4AAAA with id 88950999
PhantomJS 2.1.1 (Windows 8 0.0.0) ERROR
  Error: Cannot find module './lib/transactions/crypto.js'
  at C:/Users/alexj/projects/ark-desktop/node_modules/mocha/mocha.js:1


PhantomJS 2.1.1 (Windows 8 0.0.0): Executed 0 of 2 SUCCESS (0 secs / 0 secs)
npm ERR! Test failed.  See above for more details.
```

Tried doing absolute path and still nothing. faustbrian is helping me try and get it sorted so hopefully some progress soon!

luciorubeens:
Hmm I think the contacts should be stored as a json instead of array, something like this:

`var contacts = { "address": "name", "address2": "name" }`

So to add/edit would be a little bit easier: `contacts[address] = name`
And to remove: `delete contacts[address]`

spresnal:
@alexbarnsley sorry if I potentially created some merge conflicts for you. I didn't realize the changes here still hadn't gone in! Anyways, if you reference #379 , you may find what the issue is on your end.

alexbarnsley:
Don't worry about conflicts! I can sort those. May take a look at your test PR as may be helpful to with the tests I needed here! Thanks!

alexbarnsley:
@faustbrian been intentionally leaving as lowest on my priority list. Will close as a lot has changed and don't remember since was before I was part of the team. Will leave the branch in case anyone wants to pick it up, and will create a git issue for it.

PR #348, opened 2017-10-16
Fix/splashscreen

### LINKED ISSUE #331: Hit / Miss When Splash Image Shown
Currently there is a "hit and miss" on when image is shown with splash screen.

splashimage

Haven't been able to pinpoint on when it doesn't show and when it does.

krau612:
I moved all of the link tags to the end of the body in index.html so that all of the style sheets and scripts would load after the splash screen is displayed. I think the inconsistency of the splashscreen was caused by some things being loaded before and some things being loaded after the splash screen is shown. It seems to load the splash screen and display the image consistently now.

Also, I fixed the alignment of some tags which is why the diff is so large.

Resolves #331

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -9,12 +9,6 @@
   <meta name="description" content="">
   <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no" />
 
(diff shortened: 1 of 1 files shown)


luciorubeens:
True that.. merged manually! +5 👍🏻

PR #349, opened 2017-10-16
Disable screenshot protection setting if the platform is linux

### LINKED ISSUE #336: Screenshot Protection Doesn't Work on Linux
I'm running Ubuntu Mate 17.04 and the latest version of the wallet. I can still take screenshots with no issues. I think it's an issue with electron, however, because https://electron.atom.io/docs/api/browser-window/ only has macOS and Windows listed for "win.setContentProtection(enable)".

krau612:
Because Electron doesn't have content protection on Linux, I think the screenshot protection settings button should be disabled on Linux. Users shouldn't be given an option that they may think provides some security if the option actually does nothing, it's misleading.

Resolves #336

### DIFF
--- a/main.js
+++ b/main.js
@@ -150,7 +150,7 @@ function createWindow () {
             })
         },
         { type: "separator" },
-        { label: "Disable screenshot protection (unsafe)", click: function() { updateScreenshotProtectionItem(); }},
+        { label: getScreenshotProtectionLabel(), click: function() { updateScreenshotProtectionItem(); }, enabled: process.platform !== "linux"},
         { type: "separator" },
         { label: "Minimize", click: function() { mainWindow.minimize(); }},
         { label: "Maximize", click: function() { mainWindow.maximize(); }},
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -177,6 +177,10 @@ function createWindow () {
     }
   ];
 
+  if (process.platform === "linux") {
+    template[0].submenu[2] = { label: "Screenshot Protection Not Available On Linux", enabled: false};
+  }
```
spresnal: I don't think this is the best way to go about it -- specifically i'm talking about referencing template element 0 & submenu element 2. If a submenu gets added before the screenshot label then this will cause unintended effects.

Would be preferred to call a method from line 153 that gets the appropriate screenshot label

krau612:
Good call, thank you @spresnal . I created a function to get the proper label.

One issue though is that template[0].submenu[2] is still used in updateScreenshotProtectionItem().

luciorubeens:
+5 👍🏻

PR #350, opened 2017-10-16
Implement Toast Service

### LINKED ISSUE #344: make Toast management as a service
The way to show toast message is a bit chaotic here. A service would be good instead with function like (displayed in level order):
- toastService.error
- toastService.success
- toastService.warn
- toastService.log
- toastService.debug

A config would set level of messaging (default level `success`)

Also the toast events could be logged into some log files for debug for instance

alexbarnsley:
@fix @luciorubeens please take a look and see if this is implemented the way you were looking for. As I said in the issue, I've replaced all existing toasts to use the Service, but have set them up as either SUCCESS or ERROR. Please see below as I've added colours for each toast - GitHub has blown the images up so they're not that blurry in practice.

Config can be changed in `client/app/config/config.js`:

`logFile` - true or false - appends to `logs/ark.log` if enabled.
`level` - what log messages to show/log. Below is the mapping - this value shows itself and above (E.g. 1 shows ERROR and SUCCESS messages)
`defaultHideDelay` - how long until the toast disappears. Some places have been overridden, but the majority will use this value

Logging Enum
```
0: ERROR
1: SUCCESS
2: WARN
3: LOG
4: DEBUG
```

ERROR
image

SUCCESS
image

WARN
image

LOG
image

DEBUG
image

Resolves #344


client/app/src/services/toast.service.js
```diff
@@ -0,0 +1,97 @@
...
+        typeName = self.TypeName[type];
+      }
+      if (!stopTranslate) {
+        message = gettextCatalog.getString(message);
+      }
+      var toast = $mdToast.simple()
+        .hideDelay(hideDelay || self.hideDelay)
+        .textContent(message);
+      if (typeName) {
+        toast.theme(self.TypeName[type]);
```
spresnal: Did you intend to use typename here vs self.TypeName[type]?

client/app/src/services/toast.service.js
```diff
@@ -0,0 +1,97 @@
...
+      show(message, self.TypeEnum.LOG, hideDelay, stopTranslate);
+    }
+
+    function debug(message, hideDelay, stopTranslate) {
+      show(message, self.TypeEnum.DEBUG, hideDelay, stopTranslate);
+    }
+
+    self.show = function(message, type, hideDelay, stopTranslate) {
+      if (type > self.loggingType) {
+        return;
```
spresnal: Maybe we should log an error here before returning?

client/app/src/services/config.service.js
```diff
@@ -0,0 +1,35 @@
+(function() {
+  'use strict';
+
+  angular.module('arkclient.services')
+    .service('configService', ['$mdToast', 'gettextCatalog', ConfigService]);
```
spresnal: Looks like some of these services aren't being used

client/app/src/services/toast.service.js
```diff
@@ -0,0 +1,97 @@
...
+  var fs = require('fs');
+
+  angular.module('arkclient.services')
+    .service('toastService', ['configService', '$mdToast', 'gettextCatalog', ToastService]);
+
+  /**
+   * ToastService
+   * @constructor
+   */
+  function ToastService(configService, $mdToast, gettextCatalog) {
```
spresnal: getTextCatalog*

client/app/src/services/toast.service.js
```diff
@@ -0,0 +1,97 @@
...
+    self.logToFile = function(message, typeName) {
+      if (!self.logFile) {
+        return;
+      }
+      self.fileStream = self.fileStream || fs.createWriteStream('logs/ark.log', {flags: 'a+'});
+      if (!self.fileStream) {
+        return;
+      }
+      self.fileStream.write(
+        '[' + new Date().toISOString() + '] ' + (typeName ? typeName.toUpperCase() + ': ' : '') + message + '\n'
```
spresnal: Would look a bit cleaner to use string interpolation here

client/app/src/services/config.service.js
alexbarnsley: Ah yes, I removed the parameters from the constructor, but not here!

client/app/src/services/toast.service.js
alexbarnsley: Yeh it may be worth doing this. I will get it to log all messages to the log file, if enabled.

client/app/src/services/toast.service.js
alexbarnsley: nice spot! :)

client/app/src/services/toast.service.js
alexbarnsley: Agreed, will do :)

client/app/src/services/toast.service.js
alexbarnsley: This is passed in. I agree, I'd use camelcase normally but wanted to keep it consistent with the value on line 6/7 (as well as other places like `account.controller.js`):

```
angular.module('arkclient.services')
  .service('toastService', ['configService', '$mdToast', 'gettextCatalog', ToastService]);
```

alexbarnsley:
Thanks @spresnal! I've made all your changes aside from the camel case comment, if you'd like to take a look at my response :)

I agree with the hideDelay. I just kept it consistent with current functionality for now with the option of changing in the future if needed. Would be good to get the opinion of @fix and/or @luciorubeens, as to whether it's something we need to worry about.

luciorubeens:
Great! +10 to ConfigService and ToastService 👍🏻

PR #351, opened 2017-10-16
Show 'Well Confirmed' once the number of confirmations is greater than 50.

krau612:
Tooltip shows the actual number once well confirmed. This is the way the new explorer displays confirmations, so I updated this to match.


luciorubeens:
+5 👍🏻

PR #352, opened 2017-10-17
added copy address or copy public key when you hover over that

ckhatri:
Now when you hover over the address it'll say Copy address or the public key


luciorubeens:
+5! 👍🏻

boldninja:
@ckhatri ark address still the same?

ckhatri:
yup!

boldninja:
Cool sent

PR #353, opened 2017-10-17
Moved splashscreen css to css file

ckhatri:
Splashscreen css is currently in the html file, to clean up code decided to move it to the app.css file. 

I also moved splashscreen.png to the images folder.


client/app/assets/app.css
```diff
@@ -6,6 +6,8 @@ body {
     margin: 0px;
     padding: 0px;
     background-position: center;
+    background-color: #000;
+    border-radius: 10px;
```
spresnal: The border-radius wasn't being applied to the body before. Was this an intended change?

ckhatri: nice catch, fixed!

luciorubeens:
can you fix the conflict?

ckhatri:
Should be fixed now!

luciorubeens:
Hmm I tried it and we can't do this because the css is loaded before the splash screen..

PR #354, opened 2017-10-17
Sort peers & fix bug with filtered peer status

alexbarnsley:
This is to improve the initial connection to the network. More recently, there's been a lot of connect/disconnect activity. It now looks for the best peers based on highest Block Height, and lowest Delay.

@luciorubeens I couldn't use orderBy - it just didn't work (silently failing). Instead, I did this. I also didn't store in the serviceProvider as suggested, so i check on the first index of getGoodPeer and sort then, once.

Feedback welcome!

### DIFF
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -286,6 +286,13 @@
         //peer.ip=network.peerseed;
         return;
       }
+      if (index === 0) {
+        peers = peers.filter(function(peer) {
+          return peer.status == "OK";
+        }).sort(function(a, b) {
+          return b.height - a.height || a.delay - b.delay;
+        });
+      }
       peer.ip = "http://" + peers[index].ip + ":" + peers[index].port;
       getFromPeer("/api/blocks/getheight").then(function(response) {
           if (response.success && response.height < peer.height) {
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -268,11 +268,13 @@
       if (!network.forcepeer) {
         getFromPeer("/api/peers").then(function(response) {
           if (response.success) {
-            let peers = response.peers.filter(function(peer) {
-              return peer.status == "OK";
+            getFromPeer('/api/peers/version').then(function(versionResponse) {
+              let peers = response.peers.filter(function(peer) {
+                return peer.status == "OK" && peer.version === versionResponse.version;
+              });
+              storageService.set("peers", peers);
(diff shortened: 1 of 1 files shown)


alexbarnsley:
Removing the filter as it's a copy of another (does it twice)

alexbarnsley:
I believe I'm done making changes now. There was a bug where it filtered down peers based on status, but it didn't use this filtered down list when making the initial connection to a peer (on first load of the desktop client).

Also now sorting peers based on Height DESC, Delay ASC

client/app/src/services/network.service.js
```diff
@@ -268,10 +268,13 @@
       if (!network.forcepeer) {
         getFromPeer("/api/peers").then(function(response) {
           if (response.success) {
-            storageService.set("peers", response.peers.filter(function(peer) {
-              return peer.status == "OK";
-            }));
-            findGoodPeer(response.peers, 0);
+            getFromPeer('/api/peers/version').then(function(versionResponse) {
+              let peers = response.peers.filter(function(peer) {
```
spresnal: Does there need to be validation for response.success like the above api call here?

alexbarnsley: Not a bad shout. It shouldn't be necessary but will add it anyway and will let it fallback on the stored peers if it fails 👍

alexbarnsley: Pushed the change for you to take a look 👀

luciorubeens:
I think getting the same version from the current peer does not make sense because if the selected peer is using an earlier version the user will get stuck in that version

alexbarnsley:
@luciorubeens would that be possible without manually changing the networks? It all starts from the initial network which would be the master wouldn't it?:

https://github.com/alexbarnsley/ark-desktop/blob/f168b793e5b0b10e704a633bb1a461ba35a04d3a/client/app/src/services/network.service.js#L84-L114

Maybe I've misunderstood, I though it connected to the master network first to get the peers. I'll double check when I'm at a machine.

I'm happy to take it out though if you think it will be an issue!

luciorubeens:
yeah, but suppose we updated the ark-node, and this default peer has not updated yet.. I'm just making sure that we don't have to manually update this. I'll merge but I think we can improve a bit more.. +5! 👍🏻

alexbarnsley:
That's fair enough @luciorubeens 😁 I'll have a think about whether there's a better way to check version numbers. Perhaps any matching version or greater would work? That way you're not stuck on an old version. Does that sound like an option?

PR #357, opened 2017-10-21
Exit ledger-worker process on quit signal

krau612:
An issue was reported on [reddit](https://www.reddit.com/r/ArkEcosystem/comments/77tjkj/uninstalled_ark_client_taking_over_computer/) saying that instances of Ark were still running after the client was closed. I found that the ledger-worker was not being closed when ArkClient was closed, so ledger-worker.js and main.js were still running.

I fixed it by calling process.exit() in ledger-worker when it receives a quit message. All parts of the client now exit completely when the window is closed.

### DIFF
--- a/ledger-worker.js
+++ b/ledger-worker.js
@@ -9,5 +9,8 @@ setInterval(()=>{
   })
 }, connected ? 5000:1000)
 
-
-
+process.on('message', function (message) {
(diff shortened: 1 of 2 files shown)


luciorubeens:
sounds good! 👍🏻 +5

PR #361, opened 2017-10-23
Removed nextCurrency function

ckhatri:
Currently there exists a next currency function and a button where its used on the transactions page where we show the balance in the currency of their choice. If you click that button it will cycle through the list of currencies one by one. So you have to click it to go through all the currencies and go back to your original. That or change your settings to whichever currency.

Not too sure of the utility of this function. If the user clicks on it because it's a button and seems clickable, they'll now need to change their currency back to what it was through the settings or by clicking it over and over again. 

If they're looking for a currency they can look at the dropdown list in the settings, which would help them find their currency faster. If they want to know all the currencies available they could look at the list as well rather than cycling through.

So it seems like the function just causes a minor inconvenience to the user and should be removed since it isn't providing any great utility.

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -29,7 +29,7 @@
       display: none;
       overflow-x: hidden;
     }
-    
+
     [ng-cloak].splash {
       background-color: #06060c;
       background-image: url("splashscreen.png");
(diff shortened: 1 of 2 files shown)


client/app/index.html
```diff
@@ -503,7 +503,7 @@ <h2 translate>
               </span>
               <span class="md-subhead description">
                   <translate class="balance">Balance</translate> {{ul.network.symbol}}{{ul.selected.balance/100000000}}
-                  <md-button ng-click="ul.selectNextCurrency()">{{ul.formatCurrencyBalance(ul.selected.balance/100000000)}}</md-button>
+                  <md-button>{{ul.formatCurrencyBalance(ul.selected.balance/100000000)}}</md-button>
```
alexbarnsley: Is there another element we can change this to, since it's no longer a usable button?

alexbarnsley:
All looks good. Does what it says on the tin. I made 1 small comment but have another question which could directly affect that. Is it worth changing the currency on the main screen to also be the drop-down? Each item in the dropdown could then show the value in its respective currency? Just a thought as to whether this would make it easier for the user

client/app/index.html
ckhatri: Ye, was thinking of just removing the button and styling it a bit better.

luciorubeens:
I think it makes sense after the merge of #361, so I'll remove this manually.. +5 👍🏻

PR #362, opened 2017-10-23
Hide 'Remove' button for a Ledger account

### LINKED ISSUE #360: Remove account from Ledger
Is it possible or we should remove that function?

alexbarnsley:
Resolves #360

Can't remove a Ledger account


luciorubeens:
nice! +5 👍🏻

PR #363, opened 2017-10-23
Feature/always show btc

### LINKED ISSUE #359: Feature request : Currency 1 and Currency 2
Hello,

I've been using the wallet for a while now and I think I would be better if you could choose two currencies in the settings. In the previous releases you had to click on the price to cycle through all available currencies. Now you have to go in the settings to change currency. My idea is a mix of you had in the past and what you have now : you choose two currencies in the settings (not just one) and you switch between those two currencies by clicking on the price. The thing is, as a european, I'm only interested in knowing the price of arks in bitcoins and euros not in just one of the two and not in any other currency.

alexbarnsley:
Resolves #359 

This is subject to approval, but it gives an idea!

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -110,7 +110,7 @@
                 <md-icon>attach_money</md-icon>
                 <p translate>Currency</p>
                 <md-select aria-label="Currency" class="currency md-secondary md-no-underline" ng-model="ul.currency" ng-init="ul.currency" id="price-list" ng-change="ul.changeCurrency()">
-                  <md-option ng-value="opt" ng-repeat="opt in ul.currencies track by opt.symbol" style="text-transform: uppercase">{{ opt.name }}</md-option>
+                  <md-option ng-value="opt" ng-repeat="opt in ul.currencies track by opt.symbol" ng-if="opt.name != 'btc'" style="text-transform: uppercase">{{ opt.name }}</md-option>
                 </md-select>
               </md-list-item>
               <md-list-item>
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -319,7 +319,7 @@
                 {{it.username||it.address}}  <span ng-if="it.delegate">({{it.delegate.rate}})</span>
               </span>
               <span class="md-secondary">
-                {{ul.network.symbol}}{{it.balance/100000000}} ({{ul.formatBitcoinBalance(it.balance/100000000)}})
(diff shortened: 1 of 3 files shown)


luciorubeens:
I honestly think that is not a good option, we should use ARK as a currency and not just linking the total balance to bitcoin.. but it's just my opinion, what do you think?

alexbarnsley:
That's a fair point. I created this knowing that it may not get approved as the related issue hadn't been confirmed. I'll ask on slack what people think, see if anyone replies!

luciorubeens:
@fix @boldninja

alexbarnsley:
Mixed thoughts on Slack. Some don't want to see bitcoin all the time. Others want the option of seeing both

boldninja:
Can't we make a switch in the options (by default disabled) - always show btc + currency of your choosing and if enabled BTC is not shown anymore in the currency list, if switched back it shows BTC again in the list.

alexbarnsley:
Easy enough to do @boldninja, just need an agreement I guess 😁

boldninja:
I'm not a fan of 2 currencies myself, but if we can provide an option for those who want it why not. By default should be disabled in my opinion can be confusing, but those who'd be enabling it would presumably know what they are doing.

krau612:
I would prefer being able to quickly toggle between BTC and my selected currency without going into the settings menu. I wouldn't want to always see BTC, or Ark + BTC + my currency. Just Ark + (my currency or BTC)

alexbarnsley:
@krau612 out of curiousity, why do you only want to toggle bitcoin and not show it all the time?

krau612:
I think it's too crowded showing three different amounts (Ark, BTC, and USD in my case) all the time. I personally don't care about my BTC value too much, so I only check it occasionally. Showing Ark plus one other currency would look better. I think being able to quickly switch between a local currency and BTC would satisfy most users because most people care about their value in local currency and/or BTC.

alexbarnsley:
That's fair enough @krau612. I'm happy to go with that if you are @luciorubeens ? If so, I'll open a new PR with the changes since most of this is not really needed I don't think

luciorubeens:
Sure, but how do you will do that?

alexbarnsley:
Good question. I was thinking a button next to the currency button with the Bitcoin symbol (and tooltip saying to toggle on/off). Clicking it will flag something in the controller to change which currency we're looking at. What do you think? Maybe also show bitcoin in the dropdown, as I think the toggle should reset at the end of the session

luciorubeens:
Nice, sounds good! Store the selected currency before switching to bitcoin, to switch back.. So I'll close this PR

PR #365, opened 2017-10-25
removes deprecated method

ckhatri:


alexbarnsley:
I'd do this as part of your other PR #361 because they're related to the same method. Because that other hasn't been merged yet, the method in this PR is technically still used

ckhatri:
Agreed, will do that

ckhatri:
Ye agreed, I'll do it there as well

luciorubeens:
Ok +5

PR #342, opened 2017-10-13
Added libusb to Travis addons

krau612:
The Travis build was always failing because it was missing libusb-1.0-0-dev

PR #366, opened 2017-10-25
working travis

ckhatri:
Looks like electron uses fsevents, which is for OSX only. When you try to do the travis build on linux, this causes it to break. 

So I've set it so Travis uses os x as its operating system for the build.

Tested this on my fork, and it seems to work and pass the builds.

https://travis-ci.org/ckhatri/ark-desktop/builds/292445286


luciorubeens:
The build is working after #342

PR #370, opened 2017-10-27
Fix current account balance

noahgr:
Hi, i noticed that the currency balance is showing the total balance of all accounts


luciorubeens:
True that, good catch! +5 👍🏻

PR #371, opened 2017-10-27
Fix translate tag on public key tooltip

noahgr:
The tooltip does not translate correctly because the translation tag does not work with properties inside


luciorubeens:
Agreed! +5 👍🏻

Issue #368, opened 2017-10-25
HI i want some Add these options.


plzpickausername:
hard to explain on english
i make some picture

1) Add contents of address book (my accounts)
![a](https://user-images.githubusercontent.com/32923222/32012947-7dfda7c2-b9f4-11e7-908e-262454522041.png)
![a2](https://user-images.githubusercontent.com/32923222/32012955-8316dd82-b9f4-11e7-9c84-9910b5868662.png)

2) Add target _blank things 
![b](https://user-images.githubusercontent.com/32923222/32012350-812e038a-b9f2-11e7-94a3-947763282f1b.png)

zillionn:
1. I agree.
2. I disagree. Opening new windows is a bad user experience.

plzpickausername:
zillionn // if u just click then not open new window but 
right click and click the [open new window] or wheel click make new window

PR #372, opened 2017-10-28
Add accounts to list of addresses to send Ark to

krau612:
When sending Ark, the list of addresses/names that popup now shows your accounts and contacts, not just contacts. They are also sorted by account label / contact name.

Partially addresses #368


luciorubeens:
Sounds good! +5 👍🏻

PR #373, opened 2017-10-28
Toggle Bitcoin Currency

### LINKED ISSUE #359: Feature request : Currency 1 and Currency 2

alexbarnsley:
I believe this Resolves #359 

As discussed on #363 

@luciorubeens - take a look and let me know please!


alexbarnsley:
@luciorubeens I've fixed the conflicts

client/app/src/accounts/account.controller.js
```diff
@@ -613,6 +619,15 @@
       return balance * price
     }
 
+    self.toggleBitcoinCurrency = function (force) {
+      self.bitcoinToggle = force !== undefined ? force : !self.bitcoinToggle
```
spresnal: I think it'd be a good idea to rename bitcoinToggle to something more indicative of it being an active or inactive bool value (ex: btcValueActive). Also, you may as well use a ternary instead of the if/else.

client/app/src/filters/filters.js
```diff
@@ -13,14 +13,15 @@
       }
     })
     .filter('amountToCurrency', function () {
-      return function (amount, scope) {
+      return function (amount, scope, checkBitcoinToggle) {
```
spresnal: bitcoinToggleIsActive maybe?

alexbarnsley:
Valid points for readability. Will make those changes, thanks!

alexbarnsley:
@spresnal made your suggested changes. @j-a-m-l didn't know if you wanted to give this a quick check, as I touched upon some new stuff during conflicts (account-card).

alexbarnsley:
Hey @j-a-m-l, no worries, thanks! Out of curiosity, why would you like commits reduced? Also, is rebasing possible over merge commits - I've never tried! I'll work on some tests as they'll be useful! May need your help with this as last time I tried unit tests I was not winning (#347)

I'll look into somewhere to add the BTC toggle on the main dashboard 👌

j-a-m-l:
Yes, it's possible to rebase any commit. I asked you that to try to have a cleaner history that will be easier to search, although I'm not sure about which would be the best way to achieve that.

I've added more tests recently, so you could have a reference. Anyway, you could create the tests, but keep them ignored (`xit('this test...')` if you're having troubles and I'd complete them later.

alexbarnsley:
Thanks @j-a-m-l i'll give tidying them up a try once I'm done with tests and toggle on dashboard. I've done a few tests, just making sure they work lol. I'll let you know if I have any trouble 👌

client/app/src/components/account/templates/account-card.html
```diff
@@ -64,9 +64,15 @@
       </md-icon>
       </span>
       <span class="md-subhead description">
-          <translate class="balance">Balance</translate> {{$ctrl.ul.network.symbol}} {{$ctrl.ul.selected.balance/100000000}}
-          <md-button ng-click="$ctrl.ul.selectNextCurrency()">{{$ctrl.ul.selected.balance/100000000 | amountToCurrency:$ctrl | formatCurrency:$ctrl.ul }}</md-button>
-        </span>
+        <translate class="balance">Balance</translate> {{$ctrl.ul.network.symbol}} {{$ctrl.ul.selected.balance/100000000}}
```
spresnal: It'd be nice to standardize the 100000000 value as a variable. @j-a-m-l created a 'UNIT' variable (for arkthosi's) in one of his recent commits I believe. Not sure the variable is exposed to this logic here though.

I don't think it's necessary for it to be apart of this pull request, but something to keep in mind if we see it keep coming up.

alexbarnsley: Good point @spresnal - I'll raise a git issue as you're right, not sure it relates to the toggle, but it definitely needs updating for consistency. Thanks!

j-a-m-l: Yes, @spresnal I created that `UNIT` constant on different places as a transition to centralized its value as a configuration.

BTW: do we have a better name than "arkthosi"?

alexbarnsley: I've raised this as a separate PR - #404 (not found? 😂)

alexbarnsley:
@j-a-m-l i've added a toggle on the dashboard and a couple of tests. Would you mind running the tests to see if they work for you? I had other tests failing but wasn't sure if it was my setup. Will be creating a new linux based setup tomorrow anyway so can re-test then.

Assuming you're happy with the tests and dashboard toggle, I'll rebase some commits. thanks!

alexbarnsley:
Thanks @spresnal 🙂  tests seem to be failing on CI. The error corresponds to a line that doesn't seem to exist. That's all I need to look at 😬

alexbarnsley:
@j-a-m-l tests are passing. Are you happy? If so, I'll squash commits and merge

j-a-m-l:
@alexbarnsley there're some conflicts, could you review them, please?

alexbarnsley:
@j-a-m-l fixed the conflicts 👌

PR #375, opened 2017-10-30
Add extensions and settings file to .vscode

noahgr:
Added vscode-standardjs as a recommended extension and basic editor options.


luciorubeens:
+5 👍

PR #376, opened 2017-10-30
[Lint] Format code

noahgr:
Hi, i formatted the code using the standardjs, this is a lot of changes but I just removed the spaces, the semicolons and other minor changes. So do not worry about merging this or validating each line.. Cheers!

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,4 @@
-module.exports = function(grunt) {
-
+module.exports = (grunt) => {
   grunt.initConfig({
     jshint: {
       files: ['Gruntfile.js', 'client/app/src/**/*.js'],
@@ -12,22 +11,21 @@ module.exports = function(grunt) {
     nggettext_extract: {
       pot: {
         files: {
-          'po/template.pot': ['client/app/index.html','client/app/src/**/*.html','client/app/src/**/*.js']
+          'po/template.pot': ['client/app/index.html', 'client/app/src/**/*.html', 'client/app/src/**/*.js']
         }
-      },
+      }
     },
     nggettext_compile: {
       all: {
         files: {
           'client/app/src/utils/translations.js': ['po/*.po']
         }
-      },
+      }
     }
-  });
+  })
 
-  grunt.loadNpmTasks('grunt-contrib-jshint');
-  grunt.loadNpmTasks('grunt-angular-gettext');
+  grunt.loadNpmTasks('grunt-contrib-jshint')
+  grunt.loadNpmTasks('grunt-angular-gettext')
 
-  grunt.registerTask('default', ['nggettext_compile']);
-
-};
+  grunt.registerTask('default', ['nggettext_compile'])
+}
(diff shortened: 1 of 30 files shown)


alexbarnsley:
I wish I could explain why no semi-colons bothers me so much. I know it's a standard but it just doesn't feel right 😂

noahgr:
@alexbarnsley I've the same feeling, unfortunately it's the standard :cry:

alexbarnsley:
😩

im-not-mad-just-disappointed

luciorubeens:
Hmm it seems that nobody here wants to remove the semicolons, I think that instead of the standardjs we can use the eslint with our rules..

boldninja:
Oh look at the boys making their own standards who is cool now

semicolumn

alexbarnsley:
I did find this when trying to find out why get rid of semicolons in case anyone is interested: https://github.com/Flet/semistandard

@boldninja leave the developers to talk important things 😂

krau612:
Not the semicolons! :sob: 

jk I actually don't mind getting rid of them lol

j-a-m-l:
I don't care too much about using 1 style or another, but semicolons are almost useless, so why adding them when it's not necessary?
 - http://inimino.org/~inimino/blog/javascript_semicolons
 - http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding
 - https://www.youtube.com/watch?v=gsfbh17Ax9I

zillionn:
Because you're used to use semicolons when you write in other languages.

alexbarnsley:
Agree with @zillionn, there are more languages with a semi than not. For me as well (yes preference), a semi colon visually tells the developer it's the end of a statement. 

I'm all for a standard and keeping everything consistent though, and will get used to whatever is used

j-a-m-l:
@zillionn I'm used to languages that don't use semicolons (Ruby, Python...) so to me and more people that isn't a valid argument because different people have different background. So the standard should be for JS only, not thinking about other languages.

For me, an argument that is independent of the personal background would be @alexbarnsley position. But I think that it has a big flaw, because all JS developers know that the end of a line is visually the end of the statement, so adding a `;` at the end of the line doesn't add any special meaning.

IMHO it's just a reminiscence of the times when JS developers didn't know how to use them (http://inimino.org/%7Einimino/blog/javascript_semicolons).

Of course, it's impossible to create a standard that all people love, so maybe it's better to use 1 that it's used in more projects although we don't like all their decisions (I don't like several of them), than creating a new one and discussing about our personal preferences.

zillionn:
That was just my opinion. I don't have much time to contribute so it's not up to me. Probably the people who contribute most should decide which standard to use.

luciorubeens:
"Are we going to perpetuate this misleading tradition?" :joy:

@j-a-m-l Ok, you convinced me. Let's get used!

semicolon

@noahgr +5! 👍🏻

alexbarnsley:
That meme 😂

j-a-m-l:
I've to admit that the real reason for using StandardJS is adding a badge...

StandardJS

Issue #398, opened 2017-11-14
Price doesn't seem to update


rpolve:
Price doesn't seem to update on my desktop wallet, whatever the currency, nor am I able to force it to do so.

Arkclient 1.4.1
Ubuntu 16.04.3 LTS

j-a-m-l:
This is related to https://github.com/ArkEcosystem/ark-desktop/issues/221.

ckhatri:
Looks like v5 of the api (which we use) was deprecated October 31st. 
https://www.reddit.com/r/coinmarketcapjson/comments/6wrv72/new_version_v8_is_stable/

Currently whenever I try to query for any /ticker for v8 there's an internal server error at the moment. So you can't get the specific information about any ticker right now. 

If we use the official api https://api.coinmarketcap.com/v1/ticker/ark/ it only gives prices in btc/usd. But we could make multiple requests to get all the currencies using the convert parameter (which only takes one currency), which seems redundant.

PR #400, opened 2017-11-14
Updated to official api

ckhatri:
Just uploading this as mockup of a solution to issue #398


ckhatri:
I can clean it up! But wanted to see if y'all thought if this was a good solution before doing so.

fix:
sounds good to me, yet again the northpole api is frozen...

ckhatri:
@fix Great! 

@j-a-m-l last commit was done to fix dashboard.html since I use the available supply in the ng-if. Used this cause market cap is dependent on the supply * price. Seem good?

client/app/src/services/network.service.js
```diff
@@ -131,13 +131,24 @@
...
-          if (res.data.price && res.data.price.btc) {
-            res.data.price.btc = Number(res.data.price.btc).toFixed(8) // store BTC price in satoshi
+          if (res.data[0] && res.data[0].price_btc) {
+            res.data[0].price_btc = Number(res.data[0].price_btc).toFixed(8) // store BTC price in satoshi
           }
-          storageService.set('lastPrice', { market: res.data, date: new Date() }, true)
-          peer.market = res.data
+          storageService.set('lastPrice', { market: res.data[0], date: new Date() }, true)
+          peer.market = res.data[0]
+          $http.get('https://api.fixer.io/latest?base=USD', { timeout: 2000}).then( function (result) {
```
spresnal: Do we care about storing the value in every different one of the currencies below? Or do we just care about what the user has active?

If we just care about what the user has active, then maybe it'd be a good idea to consider passing in the convert param during the API call to coinmarketcap. That'd save us from doing the additional API call to api.fixer.io.

client/app/src/components/dashboard/account-box.controller.js
```diff
@@ -34,7 +34,7 @@
     this.myAccountsCurrencyBalance = () => {
       const market = this.accountCtrl.connectedPeer.market
       const currencyName = this.accountCtrl.currency.name
-      const price = market ? market.price[currencyName] : 0
+      const price = market && market.price ? market.price[currencyName] : 0
```
spresnal: If market or market.price is falsey does the user see 0 for their balance? If that's the case, I'd double check that we throw up a toast error message.

alexbarnsley: I just hit a similar issue to this on my PR. market is undefined until a connection to a peer is established. It resolves itself as soon as that connection happens (which already has toasts related to it) 🤔 I thought about setting to zero initially too. Perhaps we should cache the market data so we at least show something?

spresnal: I like the idea of caching the most recently retrieved value. That still leaves the case where the user doesn't have the cache built yet, but that'd be a big improvement over displaying 0 for returning users.

alexbarnsley: Yeh probably the best way to go!I think I'll do the same on my PR, default to zero. The caching / loading of cache should be done either here or another PR I think. As long as it's done before the next release of the desktop client, no one will notice too much! Perhaps if there's no market data, we show a hyphen (`-`)?

I guess the process would be along the lines of:
- Check cache for market data (load if available)
- Fetch market data from API
- Update live and cache data

Does that sound okay?

client/app/src/filters/filters.js
```diff
@@ -33,7 +33,7 @@
         var localeVersion
 
         if (currencyName === 'btc') {
-          localeVersion = 'Ƀ ' + val
+          localeVersion = 'Ƀ ' + Number(val).toFixed(5)
```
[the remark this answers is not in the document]
ckhatri: Just wanted to cap the bitcoin value to some amount of digits. Chose five just because that was the value slightly above.

client/app/src/services/network.service.js
```diff
@@ -131,13 +131,24 @@
       // peer.market={
       //   price: { btc: '0' },
       // }
-      $http.get('http://coinmarketcap.northpole.ro/api/v5/' + network.token + '.json', { timeout: 2000 })
+      $http.get('https://api.coinmarketcap.com/v1/ticker/' + network.token, { timeout: 2000 })
         .then(function (res) {
-          if (res.data.price && res.data.price.btc) {
-            res.data.price.btc = Number(res.data.price.btc).toFixed(8) // store BTC price in satoshi
+          if (res.data[0] && res.data[0].price_btc) {
+            res.data[0].price_btc = Number(res.data[0].price_btc).toFixed(8) // store BTC price in satoshi
```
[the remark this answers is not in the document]
ckhatri: Agreed, I just left that as is since it was already there. But in cleanup was planning on fixing some styling stuff up.

client/app/src/components/dashboard/account-box.controller.js
ckhatri: If the get request for the coinmarket api is faulty, then the rejection function uses the cached value. If there is no cached value, then we return zero. 

I've gone ahead and added a toast error message in that case. Saying we're unable to get market data.

client/app/src/services/network.service.js
ckhatri: I wrote it this way to mimic the old api functionality, which gave us the value in every different currency. We simply store all the currency values and toggle the right one. 

It is a bit wasteful on space since we don't need all the currencies, only the toggled active one. Only downside I could see to that is that we'd have to have a new api call each time the user toggles another currency (which probably will be rarely). 

I'd be down to work on implementing that, just trying to get this change out asap since the old api is frozen and none of the wallets are showing the right price.

client/app/src/services/network.service.js
```diff
@@ -326,6 +328,29 @@
...
+    }
+
+
+    // Updates peer with all currency values relative to the USD price.
+    function updatePeerWithCurrencies(peer, res) {
+        $http.get('https://api.fixer.io/latest?base=USD', { timeout: 2000}).then( function (result) {
+            const USD_PRICE = Number(res.data[0].price_usd)
+            var currencies = ["aud", "brl", "cad", "chf", "cny", "eur", "gbp", "hkd", "idr", "inr", "jpy", "krw", "mxn", "rub"]
+            var prices = {}
+            currencies.forEach(function(currency) {
```
alexbarnsley: There's a shorter way of doing a function (not a massive difference or required, but just an alternative):

```
currencies.forEach(currency => {
    prices[currency] = result.data.rates[currency.toUpperCase()] * USD_PRICE
})
```

client/app/src/components/dashboard/account-box.controller.js
alexbarnsley: I think if we could also cache the data within `updatePeerWithCurrencies` too, then load that as early as possible on client launch by default. The user may not then experience seeing zero at all, once that cache has been populated. You could override the cache assuming the coinmarketcap process is available.

I've commented possible caching changes in this PR.

I home that makes sense?

client/app/src/services/network.service.js
```diff
@@ -326,6 +328,29 @@
...
+                prices[currency] = result.data.rates[currency.toUpperCase()] * USD_PRICE
+            })
+            prices["btc"] = res.data[0].price_btc
+            prices["usd"] = res.data[0].price_usd
+            peer.market.price = prices
```
alexbarnsley: Cache the response here:

```
storageService.set('peerCurrencies', prices, true)
```

Then roughly Line 22 of NetworkService could become:

```
var peer = {
  ip: network.peerseed,
  network: storageService.getContext(),
  isConnected: false,
  height: 0,
  lastConnection: null,
  price: storageService.get('peerCurrencies') || { btc: '0.0' }
}
```

This will need testing. Also worth seeing what the other guys think of this approach

alexbarnsley:
I hope you don't mind. I have created a new branch & PR to do additional tweaks, along with a CMC ticker for individual networks (with fixes)

You did great work on this, just wanted to get it merged into master :)

+5! 👍

ckhatri:
Totally understand! I'd want it done asap and we're in different time zones so it works pretty well!

client/app/src/services/network.service.js
fix: setting true in `storageService.set('peerCurrencies', prices, true)` force a full save of the storage into localstorage. If it happens too often the app can be unresponsive. Otherwise the storage is getting flushed to localstorage every 10s or so.

unless this is critical (like before shutting down the app) i would remove this flag.

alexbarnsley: @fix that's good to know! I think this is only called once per network connection but I'll check and make the necessary changes regardless as this isn't critical to force a flush for the prices. This has been merged already as a separate PR so I'll make a quick fix for it. Thanks!

alexbarnsley:
Indeed! Thanks @ckhatri 😁

Issue #404, opened 2017-11-16
Make UNIT consistent across codebase


alexbarnsley:
Currently there's instances of `100000000` in the HTML side, which would be better and more consistent using the UNIT setup in the controller(s). E.g. `{{ it.balance / ac.UNIT }}` instead of `{{ it.balance / 100000000 }}`

j-a-m-l:
This has been already accomplished

PR #407, opened 2017-11-17
added arktoshi constant solving #404

ckhatri:
Created the arktoshi constant equivalent to Math.pow(10, 8)


client/app/src/accounts/account.controller.js
```diff
@@ -120,7 +120,9 @@
     ]
 
     // 1 ARK has 100000000 "arkthosi"
```
spresnal: Rename that comment above ^

client/app/src/accounts/account.controller.js
```diff
@@ -120,7 +120,9 @@
...
-    const UNIT = Math.pow(10, 8)
+    const ARKTOSHI_VAL = Math.pow(10, 8)
+
+    self.ARKTOSHI = ARKTOSHI_VAL
```
spresnal: I'm thinking this should belong in a different file for constants

spresnal:
And thus, the Arktoshi was born (pending approval ofc 👍)

client/app/src/accounts/account.controller.js
ckhatri: Wasn't entirely sure where to put the constant. Where do you think it should be?

I put it in the AccountController since all the html files w/ the value used that controller. So it minimized the scope.

client/app/src/accounts/account.controller.js
```diff
@@ -120,7 +120,9 @@
     ]
 
     // 1 ARK has 100000000 "arkthosi"
-    const UNIT = Math.pow(10, 8)
+    const ARKTOSHI_VAL = Math.pow(10, 8)
```
alexbarnsley: I think ARKTOSHI_UNIT may be a better name for it. Also because you've renamed UNIT, you'll need to do the same for all other UNIT occurrences in this file. For example:

[Line 516](https://github.com/ckhatri/ark-desktop/blob/ad209b9cfda83322f3c45c6cbca2e3a9e6309f95/client/app/src/accounts/account.controller.js#L516)
[Line 657](https://github.com/ckhatri/ark-desktop/blob/ad209b9cfda83322f3c45c6cbca2e3a9e6309f95/client/app/src/accounts/account.controller.js#L657)

client/app/src/accounts/account.controller.js
alexbarnsley: Maybe create `client/app/config/constants.js` or something to that affect? I did notice UNIT had been declared in a few controllers, so it can replace those too.

client/app/src/accounts/account.controller.js
ckhatri: Thanks for the catch! Whoops.

alexbarnsley: No worries! We've all been there 😆

ckhatri: Yup! Got 'em all I think.

alexbarnsley: Hmm. Sorry, it may not work still. For example, the first snippet below is looking for a global variable / a const called "ARKTOSHI_UNIT", but it's currently called ARKTOSHI_VAL.

```
amount: parseInt(resp.send_amount * ARKTOSHI_UNIT)
```

You could either rename the const to ARKTOSHI_UNIT or remove the const completely and just reference self.ARKTOSHI_UNIT.

```
const ARKTOSHI_UNIT = Math.pow(10, 8)
```

or

```
amount: parseInt(resp.send_amount * self.ARKTOSHI_UNIT)
```

Does that make sense?

alexbarnsley: If I'm wrong and it works, please say! I've not tested yet (I tend to check code first then test the end result) 😄

luciorubeens:
Nice for controllers, but we need a filter to handle this in the UI.. +5 👍🏻

PR #412, opened 2017-11-18
Only request currencies we want from api

ckhatri:
Had to change our base to the Euro since fixer.io can't do multiple parameters (I think).

This requests only the currencies we want from fixer.io, rather than get a list of all the currencies conversion rates.

Minor change, but now we won't be storing information we don't need. Or we could add all the currencies but that's 30+, which might make the currency list a bit too large?

### DIFF
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -153,8 +153,7 @@
         failedTicker()
         return;
       }
-
-      $http.get('https://api.coinmarketcap.com/v1/ticker/' + (network.cmcTicker || 'ARK'), { timeout: 2000 })
+      $http.get('https://api.coinmarketcap.com/v1/ticker/' + (network.cmcTicker || 'ARK') + '/?convert=EUR', { timeout: 2000 })
       .then(function (res) {
         if (res.data[0] && res.data[0].price_btc) {
           res.data[0].price_btc = convertToSatoshi(res.data[0].price_btc) // store BTC price in satoshi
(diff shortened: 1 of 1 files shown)


client/app/src/services/network.service.js
```diff
@@ -348,22 +347,32 @@
 
     // Updates peer with all currency values relative to the USD price.
```
spresnal: >// Updates peer with all currency values relative to the USD price. 

relative to the EUR price* or could change the method name below to be more specific and remove the comment

client/app/src/services/network.service.js
```diff
@@ -348,22 +347,32 @@
...
         storageService.setGlobal('peerCurrencies', prices)
       })
 
       return peer
     }
 
+    function createCurrencyApiCall(currencies) {
+        var get_request = 'https://api.fixer.io/latest?symbols='
+        currencies.forEach(function(currency) {
+            get_request += '' + currency.toUpperCase() + ','
```
spresnal: A template literal would be a bit cleaner here

client/app/src/services/network.service.js
ckhatri: Ye updated the comment!

client/app/src/services/network.service.js
ckhatri: Cleaned it up by using .toString on the currencies array instead, should have done that initially.

alexbarnsley:
Just so you know @ckhatri, I think it can do many currencies:

https://api.fixer.io/latest?base=USD&symbols=GBP,EUR,AUD

boldninja:
@ckhatri did you check fixer.io page?

"The rates are updated daily around 4PM CET every working day."

So I don't think we need to be doing API call every 5 minute for this :) .
I'd make it like this:
1. When you start ARK client get price (same if you restart or switch networks and go back).
2. If you have it opened for longer periods of time maybe do a call every few hours or around that specified time.

ckhatri:
@boldninja  I'll get on that!

PR #414, opened 2017-11-18
add arktoshi to ark filter, rather than using a constant and dividing

ckhatri:
As per @luciorubeens suggestion in #407 

Created a filter to convert arktoshi into ark.

### DIFF
--- a/client/app/src/accounts/account.service.js
+++ b/client/app/src/accounts/account.service.js
@@ -18,7 +18,7 @@
 
     self.defaultFees = {
       'send': 10000000,
-      'vote': 100000000,
+      'vote': ARKTOSHI_UNIT,
       'secondsignature': 500000000,
       'delegate': 2500000000,
       'multisignature': 500000000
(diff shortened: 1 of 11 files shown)


client/app/src/filters/filters.js
```diff
@@ -1,7 +1,7 @@
 ;(function () {
   'use strict'
 
-  angular.module('arkclient.filters')
+  angular.module('arkclient.filters' )
```
spresnal: Does the linter flag that additional space? (Also, on line 1 can you remove that semicolon -- unless it's necessary for some reason)

ckhatri: I'm not sure if the semicolon is necessary, but it's in most of the files so I think it's on purpose for some reason.

client/app/src/accounts/account.service.js
```diff
@@ -18,7 +18,7 @@
 
     self.defaultFees = {
       'send': 10000000,
-      'vote': 100000000,
+      'vote': ARKTOSHI_UNIT,
```
luciorubeens: This is the voting fee, it's not related to arktoshi

luciorubeens:
merged manually +5 👍🏻

PR #418, opened 2017-11-19
fix/ignore standardjs linter errors and add to travis

spresnal:
Fixed or ignored all the errors that came up when running standard. We'll need to revisit some of the ignored lines & test files to get the ignored errors resolved (ideally separate issues to get them corrected). However, this should be a good step in the right direction because we can get Travis running the linter for new commits.

I'll comb through this commit again a little later to double check all the changes, but should be close to what we need.

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -221,7 +221,7 @@
     if (!self.network.themeDark) self.network.themeDark = false
 
     // will be used in view
-    self.currentTheme = 'default';//self.network.theme
+    self.currentTheme = 'default'// self.network.theme
 
     // set 'dynamic' as the default theme
     generateDynamicPalette(function (name) {
@@ -242,7 +242,7 @@
 
       var transactions = selected.transactions || []
 
-      if (transactions.length > 0 && transactions[0].confirmations == 0) {
+      if (transactions.length > 0 && transactions[0].confirmations === 0) {
         return self.refreshCurrentAccount()
       }
 
@@ -414,8 +414,6 @@
 
     self.getMarketInfo(self.selectedCoin)
 
-    var setExchangBuyExpirationProgress = function (timestamp) {}
-
     self.buy = function () {
       if (self.exchangeEmail) storageService.set('email', self.exchangeEmail)
       if (self.selectedCoin) storageService.set('selectedCoin', self.selectedCoin)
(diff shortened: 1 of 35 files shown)


client/app/src/accounts/account.controller.js
```diff
@@ -490,16 +486,16 @@
         }
       }, 200)
 
-      self.exchangeSellTransaction = transaction
-      changerService.monitorExchange(resp).then(
+      self.exchangeSellTransaction = transaction // eslint-disable-line no-undef
```
spresnal: These two lines appeared to be the artifact of some refactoring. Wasn't sure if it was appropriate to remove them or not.

client/app/src/accounts/account.controller.js
```diff
@@ -1229,14 +1225,14 @@
...
-      if (self.network.themeDark) {self.currentTheme = 'dark'}
+      if (self.network.themeDark) { self.currentTheme = 'dark' }
     }
 
     // Compare vibrant colors from image with default material palette
     // And returns the most similar primary and accent palette
     function generateDynamicPalette (callback) {
       if (!self.network.background) {
-        callback(false)
+        callback(false) // eslint-disable-line standard/no-callback-literal
```
spresnal: Wasn't sure what the appropriate pattern to use here was so these got ignored

client/app/src/components/account/signing-tab.controller.js
```diff
@@ -28,14 +28,13 @@
           .clickOutsideToClose(true)
           .title(message)
           .ariaLabel(message)
-          .theme(self.currentTheme)
+          .theme(self.currentTheme) // eslint-disable-line no-undef
```
spresnal: No reference to 'self' here so this appears to be an issue that needs fixing

client/app/src/components/account/signing-tab.controller.js
```diff
@@ -61,7 +60,7 @@
             $mdDialog.hide()
           },
           function (error) {
-            accountCtrl.showMessage(error)
+            accountCtrl.showMessage(error)  // eslint-disable-line no-undef
```
spresnal: No reference to 'accountCtrl' here so this appears to be an issue that needs fixing as well

alexbarnsley:
All looking good to me, but I'm no StandardsJS expert! Everything seems to be working as expected though 👌

spresnal:
@j-a-m-l I resolved new merge conflicts. Please give this a review when you are available.

alexbarnsley:
@spresnal done a test of a few key areas - can't see any issues! Nice work 😄 

+5! 👌

PR #419, opened 2017-11-20
Refresh accounts on connect to network

alexbarnsley:
Resolves #419


alexbarnsley:
Unit Test has been made inactive - required accountController from within the test itself, which doesn't currently exist (this note is more for you @j-a-m-l)

PR #426, opened 2017-11-21
Use filters amountToCurrency and formatCurrency

ckhatri:
Noticed in some places that we weren't utilizing the filters when possible, went ahead and fixed it in a few html files.


ckhatri:
Fixed! Sorry about that. Using the filter already adds the symbol so I had to take the hardcoded symbol out.

alexbarnsley:
Looks good to me!

+5 👌

PR #428, opened 2017-11-22
Call fixer periodically

ckhatri:
As per @boldninja suggestion. The idea is that we only update the conversion rates if it's around 4pm, or it's been greater than a day since the stored value was saved. Otherwise we just use the stored value since the conversion rates for the api haven't changed.


j-a-m-l:
@ckhatri Could you resolve the conflicts? They're related to your other PR (https://github.com/ArkEcosystem/ark-desktop/pull/427)

ckhatri:
Ye I'll get that done within a day!

client/app/src/services/network.service.js
```diff
@@ -354,25 +354,83 @@
...
+      storageService.setGlobal('peerCurrencies', prices)
+      deferred.resolve(prices)
       return deferred.promise
     }
 
+    // Updates the currency conversion rates IF necessary
+    // Necessary if it isn't stored, or if the stored value is too old
+    function updateCurrencyConversionRates (peer) {
+      if (storageService.getGlobal('conversionRates') !== undefined) {
+        var priceObj = storageService.getGlobal('conversionRates')
```
spresnal: We should probably utilize 'let' instead of 'var' unless there is a specific use case that doesn't allow it

alexbarnsley:
Apologies for the delay @ckhatri - I think it may be worth looking into @spresnal's suggestion, and maybe simplifying the process of checking if we're past 4pm (CET) of the day. I think the `checkToUpdateConversionRates` may not quite work as expected (because storedHour and currHour should both above 16, if I've understood correctly. 

https://stackoverflow.com/a/23620607 - maybe that sort of thing will help?

ckhatri:
Yup! Will look into it.

ckhatri:
Closing for #452

PR #435, opened 2017-12-04
Add confirmation to warn user of the cost of second passphrase

ckhatri:
Some people on the subreddit were mentioning there wasn't any popup or information about how creating a second passphrase costs ark. This toast warning message should popup and warn them how much it'll cost!

### DIFF
--- a/client/app/config/config.js
+++ b/client/app/config/config.js
@@ -1,7 +1,7 @@
 module.exports = {
   notice: {
     logFile: false,
-    level: 1,
+    level: 2,
     defaultHideDelay: 5000
   }
 }
--- a/client/app/config/config.js
+++ b/client/app/config/config.js
@@ -1,7 +1,7 @@
 module.exports = {
   notice: {
     logFile: false,
-    level: 2,
+    level: 1,
     defaultHideDelay: 5000
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+*.un~
 *.zip
 Ark\ Client-*
 .tags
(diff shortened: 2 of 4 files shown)


client/app/config/config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   notice: {
     logFile: false,
-    level: 1,
+    level: 2,
```
j-a-m-l: Why this change?

ckhatri: The main reason for that change is because every time we send a toast message some message types are shown to the user such as error, success, but others are supposed to be logged only such as log and debug. I believe warning should be shown to the user, not logged?

Currently it does

if (type > self.loggingType) {
        self.logToFile(message, typeName)
        return
      }

j-a-m-l: Yes, a confirmation would be a new intermediate screen after clicking "NEXT" on the same window, or a pop-up with "I agree" or something similar.

ckhatri: Added an alert that shows before. They must click OK before the secondPassphrase starts.

The toast message is also there still.

j-a-m-l: Good work.

I'd remove the toast message because toast notices should be a way to reflect the end of an action on the background, an error, etc. In this case, is unnecessary and it could even be distracting.

Apart from that, the only change I'd do is using different texts:
"Are you sure?" => "Second passphrase fee"
"WARNING! Second passphrase creation costs ' + secondPhraseArkVal + '  '+ currency +'.'"
"CREATE" => "CONTINUE"

j-a-m-l: BTW @ckhatri What do you think about https://github.com/ArkEcosystem/ark-desktop/issues/431 ?

ckhatri: Should I keep the changes to the levels though? Since we do want toast warnings to be shown not logged? Or should I revert that back?

Agreed with the changes, I'll get those done soon.

j-a-m-l: I'd remove entirely the toast message and I'd revert the notice configuration to level 1.

ckhatri: I got rid of the toast message and reverted the notice!

j-a-m-l:
+5 Good work!

PR #436, opened 2017-12-04
Update nl.po

stevenhuijgens:
- More uniform texts,
- Improved translations

Multiple texts do not seem to be translated, e.g.: Settings > 'Currency', 'Play sound when receiving transactions?', 'Refresh accounts automatically?'. Also: 'Export account'.


boldninja:
We are using this to translate : http://osjc1wl.oneskyapp.com/collaboration/project?id=95031

stevenhuijgens:
@boldninja It seems like the texts are finalized, therefore I cannot suggest any improvements

PR #446, opened 2017-12-09
Add border to QR Code on Dark Theme

Wolvan:
While trying to scan the QR Code of my wallet address in the wallet, the
dark theme ended up being problematic as my phone was unable to find the
borders of the QR Code. The dark theme's QR Code should have a white
border around it to counteract this problem with scanning when using
dark theme.

I opted for an 8px border and added an element offset margin so it stays
in the place it is supposed to be.

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -409,6 +409,12 @@ md-toast.debug .md-toast-content {
     object-fit: cover;
 }
 
+/* Add a white border around the QR code to make it easier on QR Code readers to decode on dark theme */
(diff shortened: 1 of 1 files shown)


j-a-m-l:
Good work! +5

boldninja:
@Wolvan can you contact me on Slack or post your ARK address here for Github bounty?

Wolvan:
AMtXb2ZPxHYFUhB4WBRfUAt21opkEuvC8p here you go! Thanks a lot for accepting this PR!

PR #448, opened 2017-12-10
Make Ark values in transaction tab selectable

Nasicus:
I noticed that the actual Ark values (total) are not selectable because they are "disabled buttons".
This really bugged me since all other values in this table are selectable and the value is in my opinion one of the most important things and should definitely be selectable.

So what I did the following:
Changed them all to `span`'s, gave them the button style and removed (with a second class) the styles which made the text not selectable.

The before and after look was the exact same for me :)

image

Let me know what you think!
Note that this is my first commit here and I'm not sure if my apporach is good :)

BTW: There may be other places in code where this should be done, but I thought before I look for them I wanted to get feedback on this.

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -174,6 +174,13 @@ md-button.menuBtn>md-icon {
     padding: 0 15px;
 }
 
+.selectable-text {
+    -webkit-user-select: text ;
+    -moz-user-select: text ;
+    -ms-user-select: text ;
+    user-select: text ;
+    cursor: text;
+  }
 
 /* Utils */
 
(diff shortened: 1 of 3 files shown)


j-a-m-l:
I think that you approach is correct :)

Nasicus:
Thanks.

I searched for all occurrences of `md-button ` but did not find any other place where this should be required.

The only place which may be in question is here:
image

However I personally don't think it's necessary, especially since there are actions on the "folder name" if you click.

j-a-m-l:
That button should be changed, so the amount would be selectable too.

Nasicus:
Fair enough!

Done 👍

j-a-m-l:
+5 thanks

PR #451, opened 2017-12-11
Provide feedback when user refreshes account

Nasicus:
I noticed that when you click the refresh button no _progress_ and no _feedback_ is provided.

I added an animation to the refresh button when a refresh is being done (which most often you can't even see because it's too fast - but you never know.. ;) ) and added feedback (toast) when the refresh is done.

The whole thing looks like this (note: I added a manually timeout of 3000ms for demo purposes):

random

I also disabled "button-spamming" which basically means when a request is already executing clicking the button again will not do another request.

I have 3 questions:

- What do you think in general? ;)
- In the error case should the user get a specific error message or is my implementation enough?
- Concering translations: Is it correct that I just pass them as a normal string to the toast service or do I need to do something more?

Greetings :)

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -191,6 +191,10 @@ md-button.menuBtn>md-icon {
     background-color: transparent !important;
 }
 
+.rotate-360 .md-ripple-container {
+    display: none;
+}
+
 @-moz-keyframes rotate-360-keyframe { 100% { -moz-transform: rotate(-360deg); } }
 @-webkit-keyframes rotate-360-keyframe { 100% { -webkit-transform: rotate(-360deg); } }
 @keyframes rotate-360-keyframe { 100% { -webkit-transform: rotate(-360deg); transform:rotate(-360deg); } }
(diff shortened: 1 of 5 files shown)


Nasicus:
So.....

> When the user has the "Refresh accounts automatically" setting on, showing the toast every n seconds is very annoying. When that option is disabled and the user clicks on the refresh button is the expected behaviour.

=> fair enough, did not see that => fixed

> Use the animation and the same behaviour on the dashboard refresh button too.

=> done, note that I had to fix a bug which always occurs for me (and prevented me to even _test_  my refresh logic), I did the easiest possible fix, not sure if there need to be done more (however maybe not in this PR since it's not related to it and  think my fix is also correct)

After I added the animation and logic to the second button I realized that there was a lot of code and logic duplication, so I decided to refactor it. By doing that I also made the logic way more generic - I'm quite happy with the result :) I hope you are too :D

j-a-m-l:
+5

Yes, it's enough that way: the `toastService` would translate the text as expected.

PR #452, opened 2017-12-12
Call currency rates api periodically. 

ckhatri:
Continuation of #428. Had some merge/git issues so decided to just do a new branch. 

@alexbarnsley now using momentjs, helped clean up the code tons. Thanks @spresnal  for the suggestion!


client/app/src/services/network.service.js
```diff
@@ -347,22 +347,67 @@
...
+        storageService.setGlobal('conversionRates', { rates: result.data.rates, date: new Date() })
+        peer.market.conversionRates = result.data.rates
+      })
+      return peer
+    }
+
+    // Checks if the stored time and the current time has crossed 4pm CET time
+    function checkToUpdateConversionRates (storedDate) {
+      const Moment = require('moment-timezone')
+      const MomentRange = require('moment-range')
```
spresnal: I think it'd make sense to move these 'require' calls out of this method. I assume it's relatively expensive to call it repeatedly.

ckhatri: Agreed! made it global instead.

client/app/src/services/network.service.js
```diff
@@ -17,6 +17,10 @@
     var ark = require('../node_modules/arkjs')
     ark.crypto.setNetworkVersion(network.version || 23)
 
+    const Moment = require('moment-timezone')
+    const MomentRange = require('moment-range')
+    const moment = MomentRange.extendMoment(Moment)
```
alexbarnsley: It would be good if you could rename one of the "moment" constants to make them unique. I know it's case sensitive, but it may confuse people.

client/app/src/services/network.service.js
```diff
@@ -347,22 +351,65 @@
...
+      storageService.setGlobal('peerCurrencies', prices)
+      peer.market.price = prices
+      return peer
+    }
 
+    // Updates the currency conversion rates IF necessary
+    // Necessary if it isn't stored, or if the stored value is too old
+    function updateCurrencyConversionRates (peer) {
+      if (storageService.getGlobal('conversionRates') !== undefined) {
+        let priceObj = storageService.getGlobal('conversionRates')
```
alexbarnsley: I'd move this outside the if, because you're effectively doing the getGlobal twice for the same thing. Chances are performance impact is non-existent. Just thought it might be worth reducing the code smell a bit 😄

ckhatri: Ye fixed up the style like ya suggested!

client/app/src/services/network.service.js
```diff
@@ -347,22 +351,65 @@
...
+      storageService.setGlobal('peerCurrencies', prices)
+      peer.market.price = prices
+      return peer
+    }
 
+    // Updates the currency conversion rates IF necessary
+    // Necessary if it isn't stored, or if the stored value is too old
+    function updateCurrencyConversionRates (peer) {
+      var priceObj = storageService.getGlobal('conversionRates')
+      if (priceObj !== undefined) {
```
alexbarnsley: I think you will need to check for `!== null` here too. When you reset all data in the client, priceObj is null, which !== undefined.

Also, when i was force overriding priceObj.date to a few days ago, updateCurrencies was still being set to false, so it wasn't updating the price. From what I can tell, this results in the price never being updated again. My code from debugging in case it helps:

```
    // Updates the currency conversion rates IF necessary
    // Necessary if it isn't stored, or if the stored value is too old
    function updateCurrencyConversionRates (peer) {
      var priceObj = storageService.getGlobal('conversionRates')
      console.log('priceObj', priceObj)
      if (priceObj !== undefined && priceObj !== null) {
        priceObj.date = "2017-12-26T16:19:29.679Z"
        peer.market.conversionRates = priceObj.rates
        let storedDateString = priceObj.date
        let storedDate = new Date(storedDateString)
        console.log('storedDate', storedDate)
        console.log('priceObj', priceObj)
        var updateCurrencies = checkToUpdateConversionRates(storedDate)
        console.log('updateCurrencies', updateCurrencies)
        if (updateCurrencies) {
          getConversionRatesApiCall(peer)
        }
      } else {
        getConversionRatesApiCall(peer)
      }
      return peer
    }

    // api call to get the conversion rates for currencies
    function getConversionRatesApiCall (peer) {
      console.log('getConversionRatesApiCall', peer)
      var currencies = ['AUD', 'BRL', 'CAD', 'CHF', 'CNY', 'EUR', 'GBP', 'HKD', 'IDR', 'INR', 'JPY', 'KRW', 'MXN', 'RUB']
      var apiCall = createCurrencyConversionApiCall(currencies)
      $http.get(apiCall, {timeout: 2000}).then(function (result) {
        storageService.setGlobal('conversionRates', { rates: result.data.rates, date: new Date() })
        peer.market.conversionRates = result.data.rates
      })
      return peer
    }
```

Could you test and see if you can get it to force update the price (and make sure I'm not being thick)

ckhatri: Nah, I figured out my logic error. I was the dunce :(

Basically I had it set so that the 'update time' is the same day as the storedDate, but if hte stored date is past the 4PM CET then it won't pass the range test, because the update time is before the storedDate.

I got it fixed by checking to see if it's after 4pm (hour 16), if so then moment lets you add a day, I'll have that up in a bit with some more testing.

Good catch.

alexbarnsley:
All seems to be working for me. Not got any issues, and the requests are being triggered!

+5 👍

PR #455, opened 2017-12-13
now will only get check price if not undefined

ckhatri:
Saw in debugging tools that filter was throwing a lot of errors because price wouldn't be defined but we'd try accessing it. 

Now if it's undefined we return 0.


Nasicus:
Note that I also fixed this in my PR #451 - with practially the same fix :)

ckhatri:
Will close! Sorry. Was working on another PR and noticed it, didn't check.

Issue #454, opened 2017-12-12
QR Scanner doesn't work for Address field when sending


alexbarnsley:
You can scan the QR code, it acknowledges it has scanned and closes but it doesn't populate the "Address" field with the scanned address.

PR #456, opened 2017-12-13
Fix that qr scanner doesn't work correctly for 'To' field

Nasicus:
Problem was that `ng-model` is not supported on `md-autocomplete`.

To work around this we have to set the `selectAddress` property which is used on the `md-autocomplete`. However we cannot really set the field directly from the qr scanner directive (because it a "contact").

My solution was to introduce a callback-function parameter (which I personally find nicer anyway, instead of having to relay on, that the object path `send.data` exists).

Please note that I'm not very good with angular 1 - and it may very be possible that there is a better fix for this. I'm always open for feedback :)

fixes: #454

### DIFF
--- a/client/app/src/components/qr-scanner/qr-scanner.directive.js
+++ b/client/app/src/components/qr-scanner/qr-scanner.directive.js
@@ -21,7 +21,13 @@
           toastService.success(`The ${result.type} ${result.qr} has been successfully scanned.`)
         }
 
-        $scope.$parent.send.data[$scope.inputCallback] = result.qr
+        if ($scope.inputCallback) {
+          $scope.$parent.send.data[$scope.inputCallback] = result.qr
+        }
+
+        if ($scope.inputCallbackFuncName) {
+          $scope.$parent[$scope.inputCallbackFuncName](result.qr)
+        }
 
         $timeout(function () {
           $mdDialog.hide()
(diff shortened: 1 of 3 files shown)


client/app/src/components/qr-scanner/qr-scanner.directive.js
```diff
@@ -63,7 +69,8 @@
         onSuccess: '&',
         onError: '&',
         onVideoError: '&',
-        inputCallback: '@'
+        inputCallback: '@',
+        inputCallbackFunc: '&'
```
alexbarnsley: From what I can tell, this isn't used. Is that correct @Nasicus

alexbarnsley:
Just 1 question/comment really. Tested and it's working 👍

client/app/src/components/qr-scanner/qr-scanner.directive.js
Nasicus: I'm not sure I understand: I'ts caled here: https://github.com/ArkEcosystem/ark-desktop/pull/456/files#diff-3671ac06b4eabc94ab421779e16417c1R28 (Line 28) and the "value" is set here: https://github.com/ArkEcosystem/ark-desktop/pull/456/files#diff-9ce1e6b4b80b3cd9b73da170356b437cR17

alexbarnsley: My apologies, I was being thick! 👌

alexbarnsley:
Nice work! +5! 👍

PR #458, opened 2017-12-13
doesn't scroll to top of transactions on refresh fixes #457

### LINKED ISSUE #457: unexpected behavior in transaction listView
First of all: Thanks for that great wallet 👍 

But I detected a small unexpected and irritating behavior with the transaction listView. 
As I wanted to scroll through my older transactions step by step, the list jumped back to the starting position which is not nice from user experience point of view. 

**Platform/Environment** 

- macOS High Sierra v. 10.13.1
- ARK Client 1.4.3 Desktop Wallet

**Steps to Reproduce:**

1. refresh accounts automatically == true
2. open account with a few transactions 
3. scroll to bottom and wait until the next update cycle

**Actual Result**

ListView scrolls automatically to top position 

**Expected** 

ListView stays at the same scroll position

ckhatri:
Thanks to Thomgru for the suggestion in #457

### DIFF
--- a/client/app/src/accounts/account-transactions.controller.js
+++ b/client/app/src/accounts/account-transactions.controller.js
@@ -30,7 +30,6 @@
 
     $scope.$on('account:onRefreshTransactions', function (evt, transactions) {
       reset()
-      angular.element(document.querySelector('.tx-list-container'))[0].scrollTop = 0
       updateTransactions(transactions)
     })
 


spresnal:
Can't argue with a quick fix like that - nice!

alexbarnsley:
@ckhatri nice work, fixes the issue! It creates a small quirk when on a wallet with many pages, because it has to load all those transaction pages again, but it finishes eventually and the scroll stays the same.

+5! 👍

PR #462, opened 2017-12-17
noticed toggling btc didn't change market to btc values

ckhatri:
Toggling BTC using the button wouldn't update the market box to show BTC. We want to store the original/current currency for the toggle though, so we can't just overwrite currency to be btc. 

Updated so the filters pass in isBitcoinToggle.


j-a-m-l:
Good work! +5

PR #466, opened 2017-12-17
fixes couldn't set selectedTextBackgroundColor

ckhatri:
Very small change, not necessary at all, but it was bothering me to have that error, so I decided to fix it. 

Fixes Electron Helper[14028:538660] Couldn't set selectedTextBackgroundColor from default ()
error that you get at startup

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -17,6 +17,9 @@ body {
     width: 12px;
 }
 
+::selection {
+    background: #ADD8E6;
+}
 
 /* Track */
 


alexbarnsley:
@ckhatri how do i recreate this issue? I'm not able to see the Helper message you're referring to.

luciorubeens:
@alexbarnsley true, this is displayed in the terminal
but the message still appears with this PR

ckhatri:
I was getting this bug whenever I highlighted something for the first time in the app. So I would just do npm start -> highlight the price. When I was testing I didn't try opening up dev-tools. So it seems like that still causes the issue, I'll look into seeing if there's a fix for that!

alexbarnsley:
Hm I never get that issue. Would you mind doing a screen recording so I can see your steps / see if resolved? I'm happy to ignore the chromium bug

ckhatri:
Sure thing here: https://gfycat.com/YoungLazyKitten

alexbarnsley:
Seems to have fixed message on console for reload for me too.

Nice +5 👍

PR #467, opened 2017-12-17
Fixes add delegate mddialog error.

ckhatri:
Noticed that when I clicked add delegate more than once, I started to get errors. This fixes it.


alexbarnsley:
Fixes the issue for me! +5 👍

PR #468, opened 2017-12-17
Moved toastService into copy directive

ckhatri:
Previously what happened was ng-click -> called copyToClipBoard which always showed a successful toast message.

Passed in the toastService into the copy directive, so now what happens is the toast service will show the correct response. 

Partially inspired by a reddit post where the user stated that it would say copied to clipboard but then would not be able to past it.

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -235,7 +235,6 @@
     self.toggleList = toggleAccountsList
     self.createSecondPassphrase = createSecondPassphrase
     self.exportAccount = exportAccount
-    self.copiedToClipboard = copiedToClipboard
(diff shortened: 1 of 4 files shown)


alexbarnsley:
@ckhatri Do you know how to replicate the error toast? I can confirm success is working which is great. If not I will still merge, I just thought I'd test if you did know how to replicate.

ckhatri:
I'm not sure how to force the copy command to fail, but I just changed the code so it would throw, which it then caught and showed the 'fail to copy' toast.

alexbarnsley:
As we have no true way to test a failure for copy & paste, I will go ahead and merge. Copy was still working for me

+5! 👍

PR #469, opened 2017-12-17
account label filter moved to filters.js

ckhatri:
Created a function in filters.js called smallid, smallid filter now calls that, and so does account label.


j-a-m-l:
Would you mind to add some tests?

j-a-m-l:
+5

PR #470, opened 2017-12-18
Add contact directly from account overview

### LINKED ISSUE #550: Managing contacts from Explorer view
When trying to add contacts (especially ones who make transactions with you) there is no way to "Add to contacts" function. 

One must tediously copy the address, open contacts, add new contact and paste in the address and fill in the name.

The UX friendly way to enhance this experience is to add a button to every address or a right click menu option for adding the selected address to contacts.

Please make it easier to add people to ones contacts.

stevenhuijgens:
If account != contact -> add contact instead of edit contact

@alexbarnsley Edit: resolves #550

### DIFF
--- a/client/app/src/components/account/templates/main-sidenav.html
+++ b/client/app/src/components/account/templates/main-sidenav.html
@@ -46,7 +46,7 @@
     </span>
   </md-list-item>
   <md-list-item>
-    <md-button class="watch-only-address" ng-click="$ctrl.ab.addAddressbookContact()">
+    <md-button class="watch-only-address" ng-controller="AddressbookController as ab" ng-click="ab.addAddressbookContact()">
       <md-icon md-font-library="material-icons">add</md-icon>
       <translate>Add Contact</translate>
     </md-button>
--- a/client/app/src/components/account/templates/account-card.html
+++ b/client/app/src/components/account/templates/account-card.html
@@ -23,13 +23,22 @@
     </md-button>
 
     <span>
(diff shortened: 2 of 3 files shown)


client/app/src/components/account/templates/main-sidenav.html
```diff
@@ -46,7 +46,7 @@ <h2 translate>
     </span>
   </md-list-item>
   <md-list-item>
-    <md-button class="watch-only-address" ng-click="$ctrl.ab.addAddressbookContact()">
+    <md-button class="watch-only-address" ng-controller="AddressbookController as ab" ng-click="ab.addAddressbookContact()">
```
alexbarnsley: What's the reason for declaring a new instance of AddressbookController here? There used to be a similar set up which meant when adding the first contact, it didn't get populated into the main instance of AddressbookController (i'll test this scenario anyway, but was curious as to why you had to do it here)

stevenhuijgens: @alexbarnsley In the account card, when adding that specific contact, the dialog should be prefilled with the username & address. When adding a contact from the side navigation, the dialog should not be prefilled.

stevenhuijgens: The data is being passed by the ng-click attribute

j-a-m-l:
Thanks @stevenhuijgens.

I've finally merged this PR because I've realized that the changes that we were asking weren't clear enough and that it involves additional refactoring.

PR #473, opened 2017-12-18
added tooltips to peer buttons

ckhatri:
Currently clicking on MAINNET or the peer ip address calls a function. I added the tooltip so that people know exactly what is going to happen before it occurs.


luciorubeens:
+5 👍

Issue #453, opened 2017-12-12
Hardcoded Wallet Fees


Jarunik:
Currently the desktop client applies hardcoded fees.

The desktop client should fetch the fees from the network and apply the correct network fees.

An example where it leads to problems is if you add the blockpool network to the current desktop client. Blockpool lowered their fee constant for delegate registration in their network. The desktop client does not respect these network parameters and will do a vote transaction with too high a fee.

The more clone chains we get it will become more important to allow to use them through the desktop client. I think it is important to have a good integration which fetches all necessary parameters from the network.

PR #474, opened 2017-12-18
Fix multiple bugs with fees & refactor

Nasicus:
So I decided to tackle this guy! Took me quite a while to figure out the main problem :)

Main problem was:

In the desktop we already used the method `getFees` to get fees from the network, however we then used `arkjs` to create our transactions. `arkjs`  always uses the default values for all fees of all transaction types (see here for an example: https://github.com/ArkEcosystem/ark-js/blob/0b8d460a54657b24703e0b7fe59cf2067e00491b/lib/transactions/vote.js#L30 ) and we did never overwrite them.

So basically if a network had different fees for a transaction type, we ignored it and just used the default value...

Secon problem:
When clicking "Send all" in the send dialog, we used a hardcoded fee type, instead of the one from the network (however this is just a ui bug).

**NOTE:**
I will probably add a third commit to refactor this `createTransaction` method, which is imho, horribly big :P

Fixes: #453

### DIFF
--- a/client/app/src/accounts/account.service.js
+++ b/client/app/src/accounts/account.service.js
@@ -570,6 +570,7 @@
             return deferred.promise
           }
 
+          transaction.fee = fees.send
           transaction.senderId = config.fromAddress
 
           if (config.ledger) {
@@ -605,6 +606,7 @@
             return deferred.promise
           }
 
+          transaction.fee = fees.secondpassphrase
           transaction.senderId = config.fromAddress
 
           if (config.ledger) {
@@ -641,6 +643,7 @@
             return deferred.promise
           }
 
+          transaction.fee = fees.delegate
           transaction.senderId = config.fromAddress
 
           if (config.ledger) {
@@ -676,6 +679,7 @@
             return deferred.promise
           }
 
+          transaction.fee = fees.vote
           transaction.senderId = config.fromAddress
 
           if (config.ledger) {
--- a/client/app/src/accounts/account.service.js
+++ b/client/app/src/accounts/account.service.js
@@ -24,6 +24,8 @@
       'multisignature': 500000000
     }
 
+    self.cachedFees = null
+
     self.TxTypes = {
       0: 'Send Ark',
       1: 'Second Signature Creation',
@@ -266,11 +268,17 @@
       return transaction
     }
 
-    function getFees () {
+    function getFees (canUseCached) {
       var deferred = $q.defer()
+      if (canUseCached && self.cachedFees) {
+        deferred.resolve(self.cachedFees)
+        return deferred.promise
+      }
+
       networkService.getFromPeer('/api/blocks/getfees')
         .then((resp) => {
           if (resp.success) {
+            self.cachedFees = resp.fees
             deferred.resolve(resp.fees)
           } else {
             deferred.resolve(self.defaultFees)
@@ -548,7 +556,7 @@
 
     function createTransaction (type, config) {
       var deferred = $q.defer()
-      getFees().then(function (fees) {
+      getFees(false).then((fees) => {
         var account
         var transaction
         if (type === 0) { // send ark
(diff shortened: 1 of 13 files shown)


perryhoffman:
"I will probably add a third commit to refactor this sendTransaction method, which is imho, horribly big" I second this 👍 . It will make it a lot easier to test once it is refactored as well!

Nasicus:
Ok I pushed my refactor of this method...let me quote the commit message what I did:

- remove duplicated code (we had 4 times practically the same code)
- split createTransaction method into 4 public methods, which makes it clear from outside what happens
- fix: potential wrong ark fees in toast messages (within create transaction)
- fix: hardcoded token name "Ark" within toast messages (within create transaction)

I highly recommend reviewing this PR _commit by commit_  to the @reviewer - since the last commit looks like a lot happened even though it didn't.

By the way, while refactoring I also detected an "anomaly" which I'm not sure is a bug or not:
Only when creating a vote transaction we do the following on the transaction (in case of using the ledger):
`transaction.recipientId = config.fromAddress`
On all other transaction types we do NOT do this.
Is it correct, that this is only done here?

@perryhoffman and @reviewer
What do you think about the refactor? I'm not  sure if I'm satisfied... do you have any suggestions / better ideas?

Nasicus:
@j-a-m-l 
What kind of tests do you expect? Testing the createTransactions methods? And if so do I need to mock the ledgerService, storageService and so on? As far as I can see there is no "base" yet for testing this service?
@perryhoffman 
Did you already start writing tests for this method? (just because you mentioned it in your comment)

perryhoffman:
@Nasicus I haven't written any tests for the service methods yet. I was hoping they would be refactored first, it would make writing asserts much more manageable!

You're right though, there currently isn't a base spec file for the accounts service. I just gave it a shot on my local. It seems to work fine, with the addition of 1 mock you were missing `$provide.value('ARKTOSHI_UNIT', Math.pow(10, 8))`. (Ref: https://gist.github.com/perryhoffman/3f00c6b5c2ed8d4741a2b84899add0f8)

Can you double check that you have npm installed properly? And that the arkjs package exists inside the client `client/node_modules` directory (not the root node_modules dir)?

j-a-m-l:
@Nasicus I've tried the last commit (with `npm test` at the root folder of the project, in Ubuntu) and only 1 test has failed.

About which kind of tests would I expect: I'd move those methods outside the `AccountController`, as a service, and then, I'd test simple things initially.

If you prefer, just create the service and fix the one that is failing, and I'll add more tests myself.

Nasicus:
So after what feels like an enternity to set up the "base ground" for testing I'm finally done for the moment ;)

@j-a-m-l 
I extracted the methods into an own service and added "a few" ( ;) ) unit tests.

Check it out and let me know if you need more. 
I wrote this tests which immediately came to my mind.

I will do some more tests (manual on darknet) and rebase my branch this evening.

Nasicus:
Phew... ok:

- rebased
- extracted logic of `arktoshiToArk` into seperate service (just couldn't let them be in accountService, they don't belong there) 
  - note: there are many other places in code where we can now use this service, but I didn't want to have to many changes, so I only changed the files where absolutely necessary. Just tell me if you want me to replace this in this PR or in a new one :) (and what you think of it anywy :D )
- also added tests for this new service
- manually tested send, vote, createSecondPassPhrase & create delegate on darknet

client/app/src/accounts/account.service.js
```diff
@@ -971,6 +801,9 @@
 
       fetchAccountAndForget: fetchAccountAndForget,
 
+      // return a copy of the object, so the original can't be changed
+      defaultFees: JSON.parse(JSON.stringify(self.defaultFees)),
```
j-a-m-l: Probably a better option is freezing the object (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze). It's something that we should start using on more places.

PR #476, opened 2017-12-19
Add Unit Tests for Adding Contacts

ckhatri:
Posting this here even though it isn't done because I wanted to get some feedback to see if I was doing it right! 

@j-a-m-l  @perryhoffman (saw your unit tests from earlier)

### DIFF
--- a/test/components/addressbook/addressbook.controller.js
+++ b/test/components/addressbook/addressbook.controller.js
@@ -85,7 +85,6 @@ describe('AddressbookController', function () {
       beforeEach(function () {
         storageServiceMock.get = sinon.stub().returns(null)
       })
-      console.log("WE RAN HERE")
       it('sets controller contacts to empty array on invalid return value', () => {
         ctrl.getContacts()
 
@@ -95,9 +94,9 @@ describe('AddressbookController', function () {
     })
 
     context("set up addressbook contact modal", () => {
-      console.log("WE RAN TOO")
       it('sets up address book modal', () => {
         ctrl.addAddressbookContact()
+        sinon.assert.calledOnce(mdDialogShowStub)
         expect($scope.addAddressbookContact).to.have.all.keys(['add', 'cancel'])
         expect(typeof $scope.addAddressbookContact.add).to.equal('function')
         expect(typeof $scope.addAddressbookContact.cancel).to.equal('function')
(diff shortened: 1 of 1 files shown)


test/components/addressbook/addressbook.controller.js
```diff
@@ -69,13 +85,42 @@ describe('AddressbookController', function () {
...
-
       it('sets controller contacts to empty array on invalid return value', () => {
         ctrl.getContacts()
 
         expect(storageServiceMock.get.calledOnce).to.be.true
         expect(ctrl.contacts).to.deep.equal([])
       })
     })
+
+    context("set up addressbook contact modal", () => {
```
perryhoffman: This could be wrapped in a sibling block to the `'getContacts'` describe block since its testing a new method. Each method can be described using the method name: `describe('addAddressbookContact', () =>)`

test/components/addressbook/addressbook.controller.js
```diff
@@ -69,13 +85,42 @@ describe('AddressbookController', function () {
...
+      it('sets up address book modal', () => {
+        ctrl.addAddressbookContact()
+        sinon.assert.calledOnce(mdDialogShowStub)
+        expect($scope.addAddressbookContact).to.have.all.keys(['add', 'cancel'])
+        expect(typeof $scope.addAddressbookContact.add).to.equal('function')
+        expect(typeof $scope.addAddressbookContact.cancel).to.equal('function')
+      })
+    })
+
+    context("Adding contacts", () => {
```
perryhoffman: Context can help explain the environment for the asserts within. In your example, there could be asserts for each different path the code could take inside the method. Eg: 
```
context('Adding a contact with an invalid name', () => {
  it('should show an error message', () => {})
})
...
context('Adding a contact with a valid name and address', () => {
  it('should ...', () => {})
  it('should ...', () => {})
})
```

Breaking it up like this makes it a lot easier to see what is being tested.

test/components/addressbook/addressbook.controller.js
```diff
@@ -69,13 +85,42 @@ describe('AddressbookController', function () {
...
+      it('should fail to add due to empty name', () => {
+        let name = ''
+        let address = 'a'
+        ctrl.addAddressbookContact()
+        const retVal = $scope.addAddressbookContact.add(name, address)
+        sinon.assert.calledOnce(mdDialogShowStub)
+        sinon.assert.match(retVal, undefined)
+        sinon.assert.calledOnce(ctrl.getContacts)
+        sinon.assert.match(ctrl.trim(name), '')
+        sinon.assert.calledWith(ctrl.showToast, 'This Contact Name is not valid', '', true)
```
perryhoffman: Be careful when testing against explicit strings, if someone was to change the wording for messages in the app, this assert would fail (even though the functionality is left unchanged).

j-a-m-l:
+5

Good work, although I've made some changes in https://github.com/ArkEcosystem/ark-desktop/pull/493 to be more comprehensible and maintain a similar style (using `expect` instead of `sinon.assert`). Please, @ckhatri have them in mind for your next contributions.

ckhatri:
@j-a-m-l  gotcha!

PR #482, opened 2017-12-23
Unify passphrase input in the whole app

Nasicus:
While doing "a little" testing for my other PR I saw that the input (password) field of the passphrases is different depending on where you are:
 - sometimes you can toggle "view password" (eye icon)
 - sometimes there is a qr-scanner symbol

Also the same code was duplicated throughout the solution.

I extracted the logic into an own component and used this in the whole app.
Now everytime the user has to input the first or the second passphrase.
The field with the visibility togglet button and the qrcode scanner is displayed.

While doing that I one question came up:
 - Somehow the `md-required` class was not set automatically anymore on the label (because it's an own component now). So I did this manually. I don't understand why it is not set anymore and I wonder if I'm doing something wrong...

While doing the refactor I also detected 2 anonmalies which may or may not be bugs:
 -  In the `timestampDocument.html` class the secondpasphrase field is only shown if there is no ledger (`!send.data.ledger && send.data.secondSignature`). However in ALL other places, the ledger check is only added to the normal passphrase field.
- Signing messages only requires the first passphrase (it's the only "action" in the app) - is this correct? (it works though I've tested it)

I tested ALL actions where I replaced this field on the DARKNET.

Note:
Also hadd to do some changes in the `qr-scanner.directive.js`. Since I also have changes there on another PR (#456) I adjusted the file so that it's identical on both PR's and should hopefully not give any conflicts :)

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -142,6 +142,7 @@
   <script src="src/components/qr-scanner/qr-scanner.directive.js"></script>
   <script src="src/components/addressbook/addressbook.controller.js"></script>
   <script src="src/components/emptystate/empty-state.controller.js"></script>
+  <script src="src/components/passphrase-field/passphrase-field.controller.js"></script>
 
   <script src="app.js"></script>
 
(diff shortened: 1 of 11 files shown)


luciorubeens:
Great PR!

`!send.data.ledger` and `send.data.secondSignature` means the same thing, since the ledger wallets can't register a second signature. Just keep the standard.

Yeah, we only need the original passphrase to sign messages. The second passphrase is used to sign transactions, it's useful to the owner of the wallet because it is an extra layer of security.

luciorubeens:
+5 👍

Issue #483, opened 2017-12-23
Export Account should include the entire transaction history


Jarunik:
Currently the export Account button only extracts the last 50 transactions.

It would be helpful to be able to **extract all transactions** for documentation and tax purposes.

Nasicus:
True.

I'll give this a try :)

j-a-m-l:
This is fixed on #485

PR #485, opened 2017-12-23
Get all transactions when exporting account

Nasicus:
I implemented it so that it always gets 50 transactions per call (which I think is the max?).
When the export account button is clicked a modal dialog is shown, which stays open until either a successful or a fail result is returned.
If there occurs an exception in the middle of getting all transactions (e.g. 1500 were already received), then a file is still exported, but contains only these transactions.

Note that this takes VERY long on big accounts!

I tried this for Jarunik:
- He has around 45'000 transactions
- It took 8 minutes
- It took around 900 API calls

I'm not sure if we should now show a selection to the user of how many transactions he wants to export (eg. 50, 100, 500, 1000, All) or even a selection for which time span the transaction should be exported (but I don't think the API offers this possibility yet? )

Well let me know what you guys think :)

Addresses: #483


Nasicus:
I wasn't happy with what I did.

So I implemented the dialog where the user can choose how many transactions to export.
Because I think we need this anyway.
The whole thing looks now like this:

demo

Question:
- Is there no way to get the total number of transactions (in the ark-explorer there it's included when you get the account)

zillionn:
I believe the better approach is to select start and end date.

Nasicus:
@zillionn 
True. However I don't think the [API](https://github.com/ArkEcosystem/ark-node/blob/development/modules/transactions.js#L64)  supports that yet.

Let's see what the devs say :)

alexbarnsley:
I do think the date range would be the better approach. Since transactions are ordered pretty much in date order, you could keep getting pages (like you are at the minute) until you reach a specific date. I know it's a bit more of an intensive task, but that way you can guarantee exporting gives you all the transactions you need

Nasicus:
@alexbarnsley 
Done... honestly it was much harder than I thought a little "in memory" filtering would be xD

When reviewing please pay special attention to:
- getAllTransactions
- getRangedTransactions

I wrote unit tests for both of this methods, however it would still be good if someone else could verify the logic of them!

Some screenshot on how it looks now:

image

image

alexbarnsley:
Looks amazing! I'm getting the below issue though. When I open the date picker, it moves and hides the first digit of the date. The below screenshot should have "11/29/2017" in it but is only showing "1/29/2017"

image

Other than that, I love it and once that small UI tweak is made, I shall approve!

Nasicus:
@alexbarnsley 
Hmm I think it's a angular material bug, but I'll try to fix it (I noticed it's only when the wiindow is small, I can probably override `left` css attribute)

@j-a-m-l 
Good point - I even thought myself of this. However I ... hmm let's be honest, was to lazy to do it :>
Noticed that too, thought that was by design (it's arktoshi instead of ark)

Nasicus:
Done with the 3 things !
It now looks like this when it's finished:

image

I would have preferred to leave the "progress bar", but show it as complete (which is possible) and add a "success mark" in the middle of it. However I don't think the progress circle can do that, so I let it be for now.

Nasicus:
@j-a-m-l 
You probably cannot reproduce it because I fixed it :P Well anyway it happened when the dev tools were open and the windows was  maybe like 1000px wide.

I moved the button to the button row and I remebered that I can use `md-icon` and now it loooks imho waaay better when it's finished. Also removed the now "redudant" success message: Because a picture can say more than thousands words  :trollface: 

image

j-a-m-l:
+10 instead of + 5: instead of just fixing the behaviour you've improved it a lot and added tests

PR #487, opened 2017-12-25
Fix wrong display of amount in transaction list (if sender==receiver)

Nasicus:
A minor thing I saw when testing:

If you send yourself something (which obviously apart from the darknet nobody does) the transaction explorer showed it like this:

image

However this is wrong, since you did not magically get `+3` ark and only lost the fee cost.

Correct would either be to show only the fee as a negative value or - and that's what I've done - show it like this:

image

I wonder if the background-color shouldn't be "negative" (`default-red-200`)?

### DIFF
--- a/client/app/src/components/account/templates/transaction-tab.html
+++ b/client/app/src/components/account/templates/transaction-tab.html
@@ -34,7 +34,7 @@
           </td>
           <td md-cell>{{it.date | date: 'short'}}</td>
           <td ng-if="it.recipientId==it.senderId && it.type==0" md-cell>
-            <span class="md-button selectable-text" md-colors="{'background' : 'default-blue-grey-600' }">{{ '+' + ((it.amount | convertToArkValue)) + ' / -' + (it.fee | convertToArkValue) }}</span>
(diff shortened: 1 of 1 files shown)


luciorubeens:
makes sense, but seems confusing. what do you think? @boldninja

boldninja:
Makes sense, but its confusing in any case :)

Nasicus:
Well if you guys prefer we could just show it as:

`-0.1 `  (which would be just the fee)

But in my opinion this is even more confusing and less complete/correct than the other approach.

alexbarnsley:
I actually think it should show `-3.1 / +3` showing what's gone out first, then what's come in. I'm not sure there's a right or wrong answer though. I think it at least needs a tooltip detailing what happened. E.g. `You sent 3 ARK to the same address, excluding the transaction fee` or something like that. That's just my thoughts though

Nasicus:
Well the question is also if it's worth "introducing" a new translation / toast for this, even though this probably almost never occurs :D 
Concering "-" or "+" first, I personally don't care :)

Nasicus:
done

j-a-m-l:
+5

PR #488, opened 2017-12-27
Fix peer forcing.

n4ru:
When the client data is reset or initialized for the first time, peers are gathered and cached from the seed nodes. If you change the Seed Server, it will overwrite that data with new peers after a client restart.

After various frustrations in testing relays using the "force" option, I realized it doesn't actually work. If force is enabled, the peer list is not called or refreshed, which I would assume is intended. The problem however, is the specified IP is not "forced". If force is selected, the client will simply avoid calling for a fresh list of peers, but never "forces" the desired IP as the seed node. When you use force and attempt to use the client, it will still use its previous peer list. The "force" option essentially acts as a "don't update peers anymore" checkbox and nothing more.

This fixes it by forcing the peer ip to the one set.

### DIFF
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -232,6 +232,10 @@
 
     function broadcastTransaction (transaction, max) {
       var peers = storageService.get('peers')
+      if (network.forcepeer) {
+      	var forcedPeer = network.peerseed.match(/([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})\:?([0-9]{1,5})?/)
+      	peers = [{'ip':forcedPeer[1],'port':forcedPeer[2]}]
+      }
       if (!peers) {
         return
       }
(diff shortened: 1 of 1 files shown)


n4ru:
I've added peer forcing for broadcasts as well, but the regex used will currently not support URLs or IPv6, only IPv4 addresses.

alexbarnsley:
@j-a-m-l is there a reason Travis is failing? Tests haven't been changed, and I see no problems with the actual tests. Other PRs are passing.
@n4ru could you try rebasing off master in case there's a problem from when you branched off master

FYI - looks good! Will test once Travis is happy 👌

perryhoffman:
@alexbarnsley @n4ru its the [Standard JS](https://standardjs.com/) module thats failing (similar to eslint). Running `npm run standard` locally will show the formatting rules it is complaining about.  Otherwise, you can see it on line [844](https://travis-ci.org/ArkEcosystem/ark-desktop/builds/322183625#L843) -> [853](https://travis-ci.org/ArkEcosystem/ark-desktop/builds/322183625#L853) of the travis logs!

n4ru:
Should be good now.

fix:
Actually this is NOT what it is intended to do. The force peer is a security to force retreiving data to a known peer, instead of random access to the network. However the broadcast for reliability SHOULD be down as intended, hence the name broadcast.

Closing it but thanks for taking time to propose it.

PR #495, opened 2017-12-31
Move utility functions to utility services, minor bug fixes

Nasicus:
- Fixes harcoded 'Ark' and 'Ѧ' values (only seen by accident)
- Move all Arktoshi calculations to `utilityService` and use the service everywhere
- Remove `networkService` dependency from `utilityService`
- Introduce constant for ark launch date
- Move all ark relative date calculations to utility service
- Improve all tests (instead of having to define `ARKTOSHI_UNIT` everytime, just inject our constants and use them if needed)

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1668,9 +1668,9 @@
                 let secondPhraseArktoshiVal = fees['secondsignature']
                 var secondPhraseArkVal = secondPhraseArktoshiVal / ARKTOSHI_UNIT
                 var confirm = $mdDialog.confirm({
-                  title: gettextCatalog.getString('Second Passphrase') + ' ' + gettextCatalog.getString('Fee (Ѧ)'),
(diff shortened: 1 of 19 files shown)


Nasicus:
Hmm I find `getArkDate` also not optimal since a normal date is returned which has nothign to do with ark...however I thought of a better name - which would imho bi `getDateFromArkStamp` or `getDateFromArkTimeStamp` however, both of this names are so very long... so I just your proposition anyway..

j-a-m-l:
@Nasicus what about these ideas: `fromArkDate` / `dateFromArkDate` `arkStampToDate` / `arkToDate` / `fromArkToDate` / `fromArkToRealDate`?

Nasicus:
`arkStampToDate` => nice idea, renamed the opposite method to `dateToArkStamp`. 
Thank you! 😎

j-a-m-l:
+$10

Issue #489, opened 2017-12-27
ARK and NEO addresses are too similar


dwarrier:
Hi,

I accidentally sent ARK coins to a NEO address, since the client considered it to be a valid address. I'm assuming my coins are lost for good. However, I would like to make sure others don't run into this issue in the future. Is there any way to prevent this (e.g. by giving a warning if the destination address doesn't have any transaction history?)

fluorine:
> by giving a warning if the destination address doesn't have any transaction history?

I think this is quite a good idea. It also helps users of cold storage wallets, to encourage them they check twice their cold wallet address.

j-a-m-l:
After filling the form to send ARK, there's a confirmation step with all the relevant information. Adding a new confirmation should be unnecessary.

Instead of your suggestion of triggering a warning if the destination address is empty, we could show it in that confirmation page. We could check that is an existent ARK address, or even a NEO one, and show different text depending on the case.

IMHO is a very uncommon problem, so this issue would be a not urgent enhancement. @boldninja what do you think?

dwarrier:
Either one of these solutions sounds like it would prevent this issue in the future, thanks for the prompt replies!

Nasicus:
Personally I also think it's better if we show to the user that something _could_  be wrong. I'm also active in a big facebook group and I see often people who accidentally send stuff to wrong addresses (e.g. BTC to LTC and stuff like that).

I gave this a little try and it now looks like this (when the address is not valid, you cannot even click next)
![1](https://user-images.githubusercontent.com/1086065/34477401-467cc144-ef9a-11e7-9ea3-058c7f0c317b.PNG)

If the address is valid, but has no transaction, an info text is shown:
![2](https://user-images.githubusercontent.com/1086065/34477411-576f4184-ef9a-11e7-81d8-27da4ab2de38.PNG)

What do you think? Shoul I create a PR for this or do you guys find it unnecessary.

j-a-m-l:
Yes, IMHO we should use that kind of input status + message more.

The killer feature would be checking that it's a NEO address already.

Nasicus:
Will try that. i can probably simpy use their API (new service in angular).

j-a-m-l:
This is fixed in https://github.com/ArkEcosystem/ark-desktop/pull/499

PR #499, opened 2018-01-04
Address on type checks in send dialog

Nasicus:
We now have the following states:

Invalid Address (error, send is disabled):

Address is own address (warning, send is enabled):

Address doesn't have any transaction / doesn't exist (info, send is enabled):

Address is neo address (warning, send is enabled):
=> should we display it as an error or as a warning? (and disable send button) I think warning is correct.

... and of course, the state when everything is fine:

Addresses: #489


fix:
well done @Nasicus for this NEO address pinging

What i am not sure is if same publickey gives same address, if yes i guess we could lock down the 'send' button if a NEO address is matched

Nasicus:
@fix 
Thanks.
Even if it would: Wouldn't the possibility of generating the same address be impossible small anyway? If so you're right and showing and error + disabling the button would be better.

(Will resolve the conflicts this evening)

luciorubeens:
NEO uses elliptic curve secp256r1, and ARK uses secp256k1. So the private key will be different.

Nasicus:
But getting the same addresses should still be nearly impossible?  (Sorry I'm not really "educated" in this whole area :) )
And what does this mean concerning if it should be `warning` or an `error`? :D

luciorubeens:
Well, they use a different mechanism, so the same publicKey will generate different addresses. But yes, it's possible to have the same address in both blockchains. Looks good, good job! :+1:

PR #500, opened 2018-01-05
Update fi.po

Hek3:
Started to make changes/corrections to this before realizing that the translations were made on a dedicated platform. I think that there are allot of 'out of context' translations on the original that doesn't make sense.

Anyway these are the ones that didn't seem right to me. 
For example for "a delegate" there is a Finnish word "delegaatti" or even "valtuutettu" which means basically the same, instead of "edustaja" which means "a representative".

### DIFF
--- a/po/fi.po
+++ b/po/fi.po
@@ -12,10 +12,10 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 msgid "- This also can help to prove if a file has been tampered with if the SHA256 signature does not match with the one recorded on blockchain"
-msgstr "- Myös tämä voi auttaa todentamaan onko tiedostoa peukaloitu, jos tiedoston SHA256-allekirjoitus ei vastaa lohkoketjuuun kirjattua"
+msgstr "- Tämä voi myös auttaa todentamaan onko tiedostoa peukaloitu, jos tiedoston SHA256-allekirjoitus ei vastaa lohkoketjuun kirjattua"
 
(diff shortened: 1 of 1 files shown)


luciorubeens:
Hey, Hek3, thanks for your PR. But we can't update through github, we use a platform to manage the translations: https://github.com/ArkEcosystem/ark-desktop#pinned-help-us-with-translations

Hek3:
Hey Lúcio,

Sorry, should have investigated more before started to do changes on
github. The fact that this was the case was brought up at slack, but *boldninja
*suggested me to submit the PR anyway, so I did.
Anyway this raised a different issue. Correct me if I'm wrong but at OneSky
there is no way to modify or suggest a change for translations that have
been 'Finalized'.
For example Finnish translation for "Application" is translated to
"Hakemus" which means like a form that you fill out when you apply for
something, not as a software application ("Sovellus") as it should be.
There are a bunch of similar kind misstranslations that should be fixed,
but seems there's no way to do it(?).
Also while on the topic, I noticed that in the English translation there is
a typo on the word "Finnish", it's now "Finish" witch means 'to end
something'. This will now get mistranslated in every language because of
the misspell, but technically it's correct because the misspelled word has
a meaning. I made a English to English translation to correct the typo, but
before it gets corrected (if ever) the other languages may get finalized
and then there will be no chance to correct them.

Sorry for bitching about this on a PR rejection reply email. I get that
this might not be the right forum, but atm I'm not really sure where else.

With regards,
Heke







2018-01-06 11:52 GMT+02:00 Lúcio Rubens <notifications@github.com>:

PR #459, opened 2017-12-14
Add setting to set the Date Format 

fluorine:
Feature to select the date format of the transactions, discusses in issue #450 

An option to select the Date Format has been included in the configuration menu, just after the setting for language.

Date formats added:
-  MM/DD/YYYY - Used mainly in the US
-  DD/MM/YYYY - Common in Europe
-  YYYY/MM/DD - Used in Asia

PR #505, opened 2018-01-07
Add missing space in confirm dialog

Nasicus:
Before:
image

After:
image

Obviously no need for the PR reward for this, just bothered me for a while :P
  
Edit:
Or ist this even wanted? Just saw that there's also no space for other currencies in the client... well if that's the case, it can be closed obviously.

### DIFF
--- a/client/app/src/accounts/view/validateTransactionDialog.html
+++ b/client/app/src/accounts/view/validateTransactionDialog.html
@@ -45,7 +45,7 @@
       </md-button>
       <span flex></span>
       <md-button class="md-raised md-primary" ng-click="validate.send()">
-        <span translate>Send</span> {{ul.network.symbol}}{{validate.totalAmount}} <span translate>now!</span>
+        <span translate>Send</span> {{ul.network.symbol}} {{validate.totalAmount}} <span translate>now!</span>
       </md-button>
       <md-button ng-click="validate.cancel()" style="margin-right:20px;">
         <translate>Cancel</translate>
--- a/client/app/src/components/account/templates/main-sidenav.html
+++ b/client/app/src/components/account/templates/main-sidenav.html
@@ -38,9 +38,9 @@
       <md-icon ng-if="$ctrl.ul.selected" style="pointer-events: auto" ng-click="" ng-mouseover="stats = $ctrl.ab.getStats($ctrl.ul.selected.address, contact.address)" aria-label="Stats" class="md-secondary" md-menu-align-target>
         info_outline
         <md-tooltip md-direction="bottom" class="tooltip-multiline">
-          {{'Income' | translate}}: {{$ctrl.ul.network.symbol}}{{stats.income.amount}} ({{stats.income.transactions}} {{'Transactions' | translate | lowercase}})
+          {{'Income' | translate}}: {{$ctrl.ul.network.symbol}} {{stats.income.amount}} ({{stats.income.transactions}} {{'Transactions' | translate | lowercase}})
           <br>
-          {{'Expense' | translate}}: {{$ctrl.ul.network.symbol}}{{stats.expend.amount}} ({{stats.expend.transactions}} {{'Transactions' | translate | lowercase}})
+          {{'Expense' | translate}}: {{$ctrl.ul.network.symbol}} {{stats.expend.amount}} ({{stats.expend.transactions}} {{'Transactions' | translate | lowercase}})
         </md-tooltip>
       </md-icon>
     </span>


luciorubeens:
Can you check in the entire application? Because we don't define a specific standard for this

Nasicus:
Basically we almost always have a space when `network.symbol` is used (I fixed the only other 2 places which I found), but we don't have a space if we show fiat currencies (e.g. `CHF3432.23`).
I'd find it more readable when we would always have a space.

This isn't very good readable for me:
image

zillionn:
You should leave space when you're using acronym, and no space when you use a sign:
$123 -> USD 123
€123 -> EUR 123
Ѧ123 -> ARK 123

Nasicus:
No. As I tried to explain it seems to depend on the locale. If we want to do it correctly, we probably need to do it in the same way. I will maybe try something like that.

luciorubeens:
[Position of ISO 4217 code in amounts](https://en.wikipedia.org/wiki/ISO_4217#Position_of_ISO_4217_code_in_amounts)

j-a-m-l:
Ideally we should infer the user locale to format currencies (as @luciorubeens has pointed) and dates (as [here](https://github.com/ArkEcosystem/ark-desktop/issues/450)).

Since we've an almost finished PR (https://github.com/ArkEcosystem/ark-desktop/pull/459) that adds an option for selecting the date format, we could add another one for choosing the currency format.

Nasicus:
I think javascript does it already right, one more time:

```
// us locale
Number(23).toLocaleString("en-us", {style: 'currency', currency: 'chf', currencyDisplay: 'symbol'})
// produces: "CHF23.00"

// swiss german locale
Number(23).toLocaleString("de-ch", {style: 'currency', currency: 'chf', currencyDisplay: 'symbol'})
// produces: "CHF 23.00"

// spanish locale
Number(23).toLocaleString("es", {style: 'currency', currency: 'chf', currencyDisplay: 'symbol'})
// produces: "23,00 CHF"
```
And our `formatCurrency` filter does it most likels already correctly (if the language is correctly determined?):

```
    .filter('formatCurrency', function () {
      return function (val, self, bitcoinToggleIsActive) {
        var currencyName = bitcoinToggleIsActive && self.btcValueActive ? 'btc' : self.currency.name
        var languageCode = self.language.replace('_', '-')
        var options = {
          style: 'currency',
          currency: currencyName,
          currencyDisplay: 'symbol'
        }

        var localeVersion

        if (currencyName === 'btc') {
          let value = String(val).length > 8 ? Number(val).toFixed(8) : val
          localeVersion = 'Ƀ ' + Number(value)
        } else {
          localeVersion = Number(val).toLocaleString(languageCode, options)
        }

        return localeVersion
      }
    })
```

I think what we SHOULD offer is additional to selecting the `language` we should offer to select the `locale`. Because myself for example uses the client in english, but I'd still like to get the `locale` format for currencies etc. for `swiss german` (`de-ch`).

Also we probably have to implement the same logic for the `network.symbol` (e.g. `Ѧ`) and the `network.token` (e.g. `Ark`).

Obviously we then have to use this logic in the whole app.

zillionn:
Wow @Nasicus you're so thorough :), but that's probably the right way...

Nasicus:
@zillionn 
Oh no fear I'm not. I opened this PR originally with the only intention that this bug was "bothering me", without thinking of any consquencens of changing it.
Then when I had already opened it I was suddenly like "hmm, I guess in other languages there is no space sometimes", so I started looking in the app and wrote my thoughts down again.
Then only after lucio and juan posted something regarding standards and the date format it suddenly "clicked" :P

j-a-m-l:
About this PR. Probably we should go wild and add the configuration option. I could finish https://github.com/ArkEcosystem/ark-desktop/pull/459 myself before and use a similar style for choosing the currency format.

What do you think?

Nasicus:
Which configuration option?
The correct option for me would be to make the regional-setting configurable, this would mean that the date format and the currency format don't have to be configurable (and would therefore make #459 almost obsolete).

But I could also live with the other option ;)

j-a-m-l:
In this context: configuration option == Choose your currency format

I think that choosing the locale should be enough, but I'm not sure about the preferences of other people, ie: I may rather "overriding" my locale to have a space between the currency and the amount.

Nasicus:
@j-a-m-l 
I know what you mean.
We both agree that something has to be done, but probably not in this PR.
If you prefer the currency format setting together with the date that's also fine by me.

=> Maybe we can close this and create an issue instead for this?

j-a-m-l:
@Nasicus I can't reply right now about it, but I'll do it ASAP

j-a-m-l:
I think that we could close this issue, although it has been a good discussion that would be useful for the next wallet version.

PR #507, opened 2018-01-08
Make it possible to access contacts / Read-Only accounts from Dashboard

Nasicus:
image

In the current version of the client it's not possible to...
- ... create a contact (read only account) WITHOUT having to import a generate a "real account" first (ok that's a lie, you can do it via the menu, but you don't see the contact anywhere after that)
- ... go directly to a contact (read only account) (you have to select a "real account" first and can only then select a contact on the left side)

This really ~~hurts my feelings~~ bothers me when I want to install the wallet and only add read-only accounts (for example on an "unsafe computer").

I'd like to implement this and gave this some thoughts on how to integrate it smoothly in the UI.
I don't think adding another box would be a good idea, since the space is, both in the horiztonal and in the vertical, not enough.

The only good solution I came up with is using a `md-dropdown` (only a prototype yet, we can add the same total value, same actions to the contacts, etc.).

The big advantage of this is, that the old, clean UI, can stay almost exactly as it is, except for one "arrow" icon :)

demoold

=> if there are no real accounts and only contacts, we can directly set the contacts tab as active.

- Does anyone have a better idea?
- Do you think this feature is even wanted by anyone other than me?

### DIFF
--- a/client/app/assets/app.css
+++ b/client/app/assets/app.css
@@ -314,6 +314,19 @@ account-box {
   overflow: hidden;
   opacity: 0.9;
 }
+
+account-box h2 .md-select-value {
+  border-bottom-width: 0;
+}
+
+account-box h2 md-select.md-default-theme:not([disabled]):focus .md-select-value, md-select:not([disabled]):focus .md-select-value {
+  color: white;
+}
+
+account-box .md-toolbar-tools .md-button  {
+  margin-bottom: 6px;
+}
+
 #home-box-accounts {
     min-height: 180px;
 }
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -58,13 +58,13 @@
     </div>
   </div>
 
-  <div ng-cloak="" style="height: 100%" layout="column">
+  <div ng-cloak="" style="height: 100%" layout="column" ng-controller="AddressbookController as ab" >
 
     <main-navbar account-ctrl="ul"></main-navbar>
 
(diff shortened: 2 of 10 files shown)


j-a-m-l:
I use contacts as read-only accounts too ;)

Using the dropdown is quite subtle, but is a good way to show the contacts on the dashboard. Other option would be a new box with them.

But that has 1 problem: users would have normal and read-only accounts, so both should be displayed together to calculate the sum of their balances.

Nasicus:
@j-a-m-l 
I implemented it with the dropdown. Everything works already, however, I still want to refactor stuff (trying to reduce the dependency to accountController at least a little...).
You don't need to review just yet.

Nasicus:
Ready for review :)

## Stuff done:
- The Ark Client can now be used as a read-only client (no need to ever enter a passphrase if you don't want to!)
- If you have only contacts the contacts view is selected on app startup:
readonlyonly
- Obviously if you have real accounts or if you have no accounts at all, the traditional accounts view is loaded:
readonlyonly
- When you switch the view, the balance is reloaded
- Finally removed the `refreshAcccountBalances` method from `accountController` and move it to the correct place: `accountBoxController` (since it's the only place it's used)
- Moved the method `createRefreshState` to `utilityService` since it makes more sense there

## Questions left:
- If you look into the `accountBoxController` you'll see that some stuff is translated right there (and this should also work), however I'm pretty sure this strings will now not be detected anymore as translatable strings - how do I declare them as such?

luciorubeens:
Great PR! it works like a charm 👌🏻 and I did not notice any problems with the translation

Issue #484, opened 2017-12-23
Include all active networks in config


Jarunik:
It would be good to include all active networks which are compatible with the Ark desktops.

I think we should have: Ark Mainnet, Ark Devnet, Blockpool, Kapu.

balubalubalu:
Ok, I'm going to take a look to this!

j-a-m-l:
This issue will be addressed on the new version of the desktop wallet

PR #511, opened 2018-01-10
Include Blockpool and Kapu by default

balubalubalu:
This is the list of the implemented changes:
- Move the config info of the networks to a config file.
- Implement changes to include the network of blockpool and Kapu to the menu automatically
- Implement the way to refresh the menu with new networks every time we add a new one in the config file.

Solves https://github.com/ArkEcosystem/ark-desktop/issues/484

### DIFF
--- a/client/app/config/config.js
+++ b/client/app/config/config.js
@@ -5,7 +5,7 @@ module.exports = {
     defaultHideDelay: 5000
   },
   networks : {
-		mainnet: { // so far same as testnet
+		mainnet: { // so far same as testnet 
 		  nethash: '6e84d08bd299ed97c212c886c98a57e36545c8f5d645ca7eeae63a8bd62d8988',
 		  peerseed: 'http://5.39.9.240:4001',
 		  forcepeer: false,
--- a/client/app/config/config.js
+++ b/client/app/config/config.js
@@ -3,5 +3,57 @@ module.exports = {
     logFile: false,
     level: 1,
     defaultHideDelay: 5000
-  }
+  },
+  networks : {
+		mainnet: { // so far same as testnet
+		  nethash: '6e84d08bd299ed97c212c886c98a57e36545c8f5d645ca7eeae63a8bd62d8988',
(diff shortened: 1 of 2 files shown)


j-a-m-l:
@Jarunik any more network?

Jarunik:
I think the two are fine for now to have the live once.

luciorubeens:
This PR is controversial, need to discuss before merge. I think it is not a good option, only devnet already confuses users.

boldninja:
I have to agree with @luciorubeens what happens once there are 20-30 forks of ARK? I'd move this custom forks into a kind of "folder" in manage networks and if you click there would be a list of all avail forks and when you click on it will ask you if you want to get it added to the list.

alexbarnsley:
What if we have a marketplace somewhere like networks.ark.io, where there's a link to their own homepage plus a link to load the network into the desktop application through a custom URI scheme? 🤔

fix:
I would not do that, because 
- it would look too complex for users
- would let think that blockpool, kapu or others a under quality ark blockchain


Instead we should think about something that would make the blockchain some first class citizen
somehow.

maybe some kind of build config that would make easy to compile a kapu desktop client.

or a "first start" welcome screen to choose default blockchain you are in, and hiding the others into some "advanced user" settings

j-a-m-l:
Yes, choosing the blockchain at the beginning would be probably the best way to make them "first class citizens".

What do you think about using a grid like the appearance dialogue with the name of the blockchain, an (optional) image, and a description and or link to its website? It would be used to change blockchain / network instead of the current dropdown.

As a reference:
appearance

j-a-m-l:
@fix @luciorubeens @alexbarnsley @boldninja any new feedback?

alexbarnsley:
@j-a-m-l I like the idea of a market place, either when first opening or an option in the menu to view "Network Marketplace". I think it needs to be easier for users to add a new network, without having to know IPs, etc. But we also can't be hardcoding each sidechain into the repo. Perhaps this PR should be closed and the discussion carries on on #484

PR #514, opened 2018-01-11
Only use ellipses if string is long enough

arkmoon:
The change simply ensures that the full ID is at least longer than the two individual slices so that there is no text overlap.

### DIFF
--- a/client/app/src/filters/filters.js
+++ b/client/app/src/filters/filters.js
@@ -63,6 +63,6 @@
   }])
 
   function smallId (fullId) {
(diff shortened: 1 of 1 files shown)


j-a-m-l:
Thanks for this attempt, but this is being addressed [here](https://github.com/ArkEcosystem/ark-desktop/pull/503/commits/f24ae6cef51f87caa1cf3222ff477e68242285f6).

In fact, the correct behaviour would be something that adapts the length to the screen.

PR #519, opened 2018-01-13
Fix bug 'Cannot set property 'scrollTop' of undefined'

Nasicus:
Happens when you create / import a new account which has no transactions.

image


luciorubeens:
Nice! 👍🏻

Issue #516, opened 2018-01-11
Don't Fixate Just on 1 IP At Fresh Install


boldninja:
Right now if you install ARK desktop on first start it tries to get peers list from 1 fixed IP. If that seed is down it won't be able to get peer list and not able to connect to network. Solution is to not relay on just 1 IP at fresh install, but pool / ping a few from seed servers.

Nasicus:
Maybe I don't fully understand it, but I guess you're talking about the `peerseed`:

```
mainnet: { // so far same as testnet
            nethash: '6e84d08bd299ed97c212c886c98a57e36545c8f5d645ca7eeae63a8bd62d8988',
            peerseed: 'http://5.39.9.240:4001',
            forcepeer: false,
            token: 'ARK',
            symbol: 'Ѧ',
            version: 0x17,
            slip44: 111,
            explorer: 'https://explorer.ark.io',
            background: 'url(assets/images/images/Ark.jpg)',
            theme: 'default',
            themeDark: false
          },
```` 
However isn't it a little of a "vicious cycle" problem? Because the call `http://5.39.9.240:4001/api/peers` returns the list of peers...?
Or is the solution simply to provider an array of - let's say -  10 hardcodes peerSeeds which will be static in this file?

alexbarnsley:
That's the solution. I was going to replicate the sort of process as I did on the mobile app. If main fails, it tries the array

boldninja:
@Nasicus exactly, but if http://5.39.9.240:4001/api/peers is down you won't even be able to get a list of peers on a fresh install as this is the only IP it gets peers from.

Nasicus:
@boldninja yes then I understood correctly ;) But it's kind of a vicious cylce :D (Well except for hardcoding multiple ips :) )

@alexbarnsley 
Do you wanna fix it or should I do it?

alexbarnsley:
Feel free to give it a go, and let me know if you have any trouble! I won't get chance to work on it until late tomorrow, or early next week I don't think

Nasicus:
Hmm I guess you're talking about this list:
https://github.com/ArkEcosystem/ark-ts/blob/master/src/config/index.ts

However in `arkjs` there's no such list - right?

So my questions now:

- Should I copy the same list into `ark-desktop`?
- Should I copy the same list into  `arkjs` - make a PR there, then use the updated package in `ark-desktop`? (_better_)
[...]

PR #523, opened 2018-01-15
Make ark client work if main peerseed is down on fresh-install

Nasicus:
Implemented as follows:

- The mainnet (ark) and the devnet (dark) networks are now created with the values from `ark-js`. 
  - Because of this the `peerseed` of the mainnet (`http://5.39.9.240:4001` => `http://node1.arknet.cloud:4001` ) and the peerseed of the devnet  (`http://167.114.29.55:4002` => `104.238.165.129:4002` ) are now different!
    -   I did not come up with these values myself, they are the same as we use in `ark-ts`
  - ~I used the `pubKeyHash` as the `version`. I'm not sure if this correct :P For `Ark` it appears to be the same value, but for `DARK` it changed from `30` to `0x52`. Sorry if this is wrong, I think I don't even fully understand what the difference between these values is (if there is any). But it's not the same, I think it wold be good to add the version to the network config in `ark-js` and `ark-ts`~.
    - Ok found out it's not the same. Still two questions: Can someone explain me `pubKeyHash` vs `version` (and why it's the same value in Ark) and also should the `version` als be added to `ark-js`? 
- If the app is started, the peerseed is down and there is no peer list in the storage, the peer list from `ark-js` is taken. If one of these peers could be reached, then the real peer list, of the newly picked peer is loaded and stored in the storage. So on the next run it will work without the static peer list from `ark-js`
- I do never change the `activePeerSeed` in the storage / network config. Basically this means if the activePeerSeed is down forever the first request will always fail (it's the same way, as it was before). I guess this is correct? Or should it be updated to a working peer once we got any?

ToDos:
 - [x] Increase ark-js package version / use newer ark-js version

Depends on: https://github.com/ArkEcosystem/ark-js/pull/48
Fixes: #516

### DIFF
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -324,7 +324,7 @@
         peers = tryGetPeersFromArkJs()
         isStaticPeerList = true
       } else if (index === 0) {
-        peers = peers.sort((a, b) => b.height - a.height || a.delay - b.delay)
+        peers = peers.sort((a, b) => b.height - a.height || a.delay - b.delay).filter(p => p.ip !== '127.0.0.1')
       }
 
       // check again or we may have an exception in the case when we couldn't get the static peer list from arkjs
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -10,13 +10,15 @@
    */
   function NetworkService ($q, $http, $timeout, storageService, timeService, toastService) {
     const _path = require('path')
+    const ark = require(_path.resolve(__dirname, '../node_modules/arkjs'))
+    const mainNetArkJsNetworkKey = 'ark'
+    const devNetArkJsNetworkKey = 'testnet'
 
     let network = switchNetwork(storageService.getContext())
 
     if (!network) {
       network = switchNetwork()
     }
-    const ark = require(_path.resolve(__dirname, '../node_modules/arkjs'))
     ark.crypto.setNetworkVersion(network.version || 23)
 
     const momentTimezone = require('moment-timezone')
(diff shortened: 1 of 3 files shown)


faustbrian:

Nasicus:
rebased - new arkjs was automatically taken when I used npm install

alexbarnsley:
@Nasicus change the branch in package.json that's being looked at to "master" then give it some testing. We only recently re-merged to the mainnet branch. If this causes problems in the future then we may revert back to it's own branch. It's good to keep updated though

Nasicus:
Sorry - this got lost somehow.

Will do this this evening.

Nasicus:
So it appears my fix & also everything else is still working.

I changed the `package.json` so it check out the `master` branch and then did `npm install` which auto-changed the `package.lock.json` hope this is correct this way 🤔

j-a-m-l:
@alexbarnsley do you about why is `127.0.0.1` being used in this context? (I was wondering about it too).

alexbarnsley:
I noticed this on the mobile app so possibly a desktop issue too. Basically, when fetching a list of peers from a node via the API, it returned the 127.0.0.1 address. To fix I just filtered out that IP in the list of peers. I can't say that's definitely the issue, but certainly sounds like it

Nasicus:
But is it correct that `127.0.0.1` is even contained in the response for the peers?

alexbarnsley:
It's not but it's a node issue so unlikely to be fixed before V2 as I think it's always been an issue, and not a critical one.

Mobile app check:
https://github.com/ArkEcosystem/ark-mobile/blob/master/src/providers/ark-api/ark-api.ts#L86

Nasicus:
I integrated the filtering into this PR, since I did changes there anyway.

alexbarnsley:
Nice work! I stopped my machine from connecting to node1.arknet.cloud, reset the wallet and I was still connecting to peers 👌

Issue #520, opened 2018-01-13
"Type an address or a name" but only an address is accepted.


n4ru:
Haven't actually tested with previous versions, but it looks like the client no longer allows sending to names as of the newest version - it gives the error "The address is not valid!".

Nasicus:
Just tested it. It always acted like that.

But I agree with you: You should proabably even be able to just type the name of a contact and it should be valid and be resolved.

n4ru:
I was thinking more about delegate names, but that too.

zillionn:
You can't use delegate name to send ARK, only contacts.

n4ru:
That's the problem, you should be able to since it resolves to an address just fine.

Nasicus:
I personally agree that it would be nice. HOWEVER we'd need to ensure that no conflicts occur, e.g. if you have a contact with the same name as a delegate. Then we'd need to show a warning / error message.

zillionn:
If you connect to a bad peer it could resolve you a bogus delegate's address. Always use address.

alexbarnsley:
I agree with @zillionn. It should be contacts only. If you regularly send to a delegate then you can add them as a contact, because then you've effectively approved the address

n4ru:
@zillionn I did not consider that, good catch.

@alexbarnsley Right. I guess contacts should still be resolvable by name when sending.

PR #524, opened 2018-01-15
Resolve contacts in send form & optimizations

Nasicus:
This does the following:
- On type in address field now also works for wallet labels (did only work for contacts before)
- Form cannot be sent if address is empty
- Known addresses (contacts & wallet labels) are now always resolved and displayed below the field (it always bothered me, that you selected a contact, but didn't know which one it was after you did so, because only the address was displayed). I'm not yet sure if it's optimal from a UX standpoint:
 capture
- Contacts & wallets can now be resolved without explicitly clicking on it in the dropdown (this only works if you type it 100% correctly, also it only works if the contact is non-ambiguous). Also while you're still typing the `Invalid Address` error is not shown if you're typing the name of a contact.
  resolve
  - If the contact is ambiguous for some reason (case insensitive!) an error is shown:
     ambiguous

Fixes: #520

### DIFF
--- a/client/app/src/accounts/account-transactions.controller.js
+++ b/client/app/src/accounts/account-transactions.controller.js
@@ -84,11 +84,11 @@
 
       const mergeTransactions = [...transactions, ...vm.transactions]
       // remove duplicates
-      const uniqueTransactions = mergeTransactions.filter((obj, pos, arr) => {
+      const singleTransactions = mergeTransactions.filter((obj, pos, arr) => {
         return arr.map(obj => obj['id']).indexOf(obj['id']) === pos
       }).sort((a, b) => b.timestamp - a.timestamp)
 
-      vm.transactions = uniqueTransactions
+      vm.transactions = singleTransactions
       storageService.set(`transactions-${vm.address}`, vm.transactions, true)
     }
   }
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -326,7 +326,7 @@
       }
     )
 
-    self.getAccountIcon = (account) => {
+    function getAccountIcon (account) {
       if (account.delegate) {
         return 'security'
       }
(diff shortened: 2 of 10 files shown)


Nasicus:
Also added this info to the validation page, but kept it more subtile and decided to only show it as a icon with a tooltip (I also had it as a separate row once, but find it this way less disruptive):
tooltip

Also changed the labels from `Contact / wallet` to `Contact` or `Account`.

Nasicus:
@j-a-m-l fixed conflicts ;)

Nasicus:
I changed this a little. I now like it much more than what I did before, see screenshots here:
https://github.com/ArkEcosystem/ark-desktop/issues/537

j-a-m-l:
Now it works slightly different, please, @Nasicus take a look. Probably your way is better.

Nasicus:
- rebased
- integrated @j-a-m-l changes (looks MUCH better now than before - thanks for the "input"!)

The final gif of how it looks and works:

resolve

Nasicus:
Problem is that they look almost the same as contacts.
I'd rather replace all account icons over the app with the wallet icon?

j-a-m-l:
There are 4 types:
- Ledger
- Account that are on the blockchain (at least 1 transaction) => maybe this one could be replaced by the `account wallet balance` icon.
- Account that aren't (`cloud off`)
- Delegates

I think that it's useful knowing which accounts haven't been used yet.

One way to differentiate them could be using different colours.

Nasicus:
Damn it you're right - totally forgot about these.

However I don't think it would be good to use all these different icons in the search box, because the aim of the current two icons is to clearly distinguish between contacts and accounts / wallets.

What is your current proposed solution? I must send I don't know what'd be best...

j-a-m-l:
I understand you point, but I think that it's better to maintain the familiar icons instead of using a new one.
Apart from that, it would be faster to spot a specific account thanks to the icon.

Nasicus:
At @j-a-m-l what do you say now:

image

I also added a tooltip if you hover over the icon in the dropdown:
image

Nasicus:
Also renamed `Unique` to `Single` (as discussed), see 34a21f7

j-a-m-l:
Good work!

Issue #526, opened 2018-01-16
[Bug] [Version 1.5.0] Voting doesn't work for wallets with 2nd passphrase 


vmunich:
First report: https://www.reddit.com/r/ArkEcosystem/comments/7qpaiz/im_having_trouble_with_voting_with_second/

Voting on 1.4.3: 
![](https://i.gyazo.com/c95a73b6192e6da5391b014b5cd65404.png)

Voting on 1.5.0:
![](https://i.gyazo.com/d6bc1df3f394d1cd53745adff7c9847f.png)

Bug introduced on https://github.com/ArkEcosystem/ark-desktop/pull/491

luciorubeens:
Thanks for opening the issue, I'll close this because it was already opened on #525.

Issue #525, opened 2018-01-16
Voting for Delegate with Second Passphrase Enabled


ABootleg53:
If you have created a second passphrase, there is no way to vote for a delegate through the desktop wallet. See images attached https://imgur.com/a/qBjC8 for recreation on Dev net - signing second passphrase, then the UI for "Voting" then the error. Wallet is 1.5.0 for both. Original post [here](https://www.reddit.com/r/ArkEcosystem/comments/7qpaiz/im_having_trouble_with_voting_with_second/)

PR #527, opened 2018-01-16
Fixes voting with second passphrase enabled

Nasicus:
I wonder why does the  method `accountService.getPassphrases` even exist? It's a little confusing, because apparently it just get's the passphrase from the store, but we (hopefully) never store them anyway.

Ok just checked:
There are methods for saving the passphrases, but they are not used anymore.
Wonder if we should remove these methods altogether (in a separate PR maybe)

@perryhoffman Maybe you have any feedback on this, since you know this voting-area well ;)

fixes: #526 
fixes: #525

### DIFF
--- a/client/app/src/components/account/votes-tab/templates/vote.dialog.html
+++ b/client/app/src/components/account/votes-tab/templates/vote.dialog.html
@@ -26,7 +26,7 @@
         </md-input-container>
 
         <passphrase-field ng-if="!$dialog.account.ledger" ng-model="$dialog.passphrases.first" label="Passphrase"></passphrase-field>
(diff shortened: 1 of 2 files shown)


luciorubeens:
Exactly, this passphrase-saving feature was not even used, you can remove it!

This `passphrasesArr` is confusing, but as you said let's do it in another PR.

Good job! Thanks ✌🏻

perryhoffman:
Yeah @Nasicus, @luciorubeens is right. I looked at it quickly and couldn't find the use for it during my last PR. I didn't want to rip it out of the code in my PR incase their was something using it that I may have missed! Good work 👍

Issue #518, opened 2018-01-13
Errors of POST requests are not displayed (affects sending and voting)


doublehelix24:
**Update:**
When an error in sending or voting (or any POST request really) occurs the UI does not show any message. I.e. you can click through the sending/voting wizards and after that just _nothing_ happens. No error message, no green confirmation either, just nothing.

I propose that those POST error messages should be displayed in order to help the user solve the issue.
For example, if I set my computer's clock to 1 minute in the future I get the silent failure. In the dev console I see a helpful error message in response to the POST request. Reading that would help many users whose clocks might be slightly off.

In my case the error was caused by my proxy server's settings related to the X-FORWARDED-FOR header - but I imagine any network issue (or even a temporary server issue) would be affected.
**/Update**

Voting does not work in the Windows 64 Bit 1.5.0 Version: The two wizard pages work normally but after the final "SEND A1 NOW" button the wizard disappears and nothing happens. No TX is created and no feedback, neither positive nor negative, appears.
The issue is repeatable - it never works.

Doing the _exact_ same process with the exact same wallet on the Linux 64 Bit 1.5.0 Version **does** work.

More info:
* Peer Status is green, current block height and delay of 120ms
* Windows' clock is accurate

I can provide more info and do tests.

![image](https://user-images.githubusercontent.com/17807858/34908790-f4f52e2e-f895-11e7-9377-de720b7a6f0c.png)

Nasicus:
Just tested on Devnet with Windows 10 x64. It works for me.
Can you open the Dev console (`CTRL + D` or via the edit menu) and see if there's an error?

alexbarnsley:
Just make sure the wallet is definitely connected to a peer at the time of sending the transaction

doublehelix24:
@alexbarnsley Yeah, peer status was green, block height up to date and delay was shown as 120ms (i.e. normal).

@Nasicus Will do. Need to get DEVNET ARKs first since my real wallet now _has_ voted thanks to the Linux version and I don't want to waste real ARKs.

Nasicus:
Just click on the "Wifi"-Icon on top to switch the network.
And here you can create a DARK account with 50 DARK: https://kristjank.github.io/dark-paperwallet/

doublehelix24:
Aha, the dev console shows an Internal Server Error:

dependencies.min.js:1 POST http://159.89.92.13:4002/peer/transactions 500 (Internal Server Error)

[...]

PR #528, opened 2018-01-16
Show error messages when a transaction fails

Nasicus:
The error is only shown if the transaction to the peer fails.

If any further broadcasts fail we show nothing.
I wonder if we should do something in this case?
Log somewhere (console :P)

@doublehelix24 If you want you can test it and give feedback :)

fixes: #518

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1550,6 +1550,8 @@
         })
       }
 
+      const transactionLabel = accountService.getTransactionLabel(transaction)
+
       function send () {
         $mdDialog.hide()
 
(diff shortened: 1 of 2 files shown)


j-a-m-l:
I'm faking an error inside the `broadcastTransaction` but isn't catched.

Probably its `for` loop could be changed to `Promise.all` and then check if all of them have been resolve or not.

Nasicus:
@j-a-m-l 
As I tried to explain:
When the first call was successful and only the broadcasts fail I don't show an error.
Because wouldn't it be confusing to show an error to the user if one broadcast fails, even though the transaction was still successful (at least to one peer)?

faustbrian:

PR #532, opened 2018-01-18
Delete all "save" / "stored" passphrase related stuff & fix small bug

Nasicus:
Note: it was NOT used! But when only quickly looking at the code it was confusing and made the impression we could have stored the passphrases.

As discussed in #527 

Also fixes `TypeError: Cannot read property 'price' of undefined` which happened if the peer was not yet connected and you already selected an account.

Question:
There are also some mentions of "save passphrase" related stuff in `template.pot`. However I guess I don't need to adjust that?


faustbrian:
@Nasicus could you fix the conflicts?

Nasicus:
@faustbrian done. Note that doing this was already discussed with @luciorubeens .

PR #536, opened 2018-01-22
Add support to Ark URI scheme [AIP-13]

luciorubeens:
- [x] Intercept the `ark:` protocol requests.
- [x] Use [ark-qrcode](http://github.com/luciorubeens/ark-qrcode) to validate and deserialize the URI.
- [x] Open the send dialog and fill the data (`address`, `amount` and `smartBridge`) with the scheme.
- [x] Check on all platforms.
- [ ] Write unit tests.

Edit: It will not work on Linux because we can't upgrade the [Electron to version 1.8.1](https://github.com/electron/electron/pull/10670) since the [`node-hid` does not support it yet](https://github.com/node-hid/node-hid/issues/233).

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -61,6 +61,7 @@
     utilityService
   ) {
     const _path = require('path')
+    const electron = require('electron')
 
     const self = this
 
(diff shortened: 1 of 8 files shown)


zillionn:
You mean AIP-13?

luciorubeens:
Exactly, thanks! @zillionn  😆

client/app/src/accounts/account.controller.js
```diff
@@ -98,6 +99,17 @@
 
     pluginLoader.triggerEvent('onStart')
 
+    electron.ipcRenderer.on('uri', (event, uri) => {
+      $timeout(() => {
+        const qrcodeElement = document.querySelector('ark-qrcode')
+        const scheme = qrcodeElement.deserializeURI(uri)
+
+        if (!scheme) return toastService.error('Invalid URI')
+        if (scheme && !self.selected) return toastService.error('Select an account and open the URI again')
```
zillionn: How about a dialog to select an account?

luciorubeens: would be great! however it will be pending for next versions, I will open an issue to handle this and the unit tests

PR #539, opened 2018-01-27
Improve Contact information

Nasicus:
The commit fixes that the contact information icon is not shown if the selected account is the contact itself, since it was very confusing.

----

Also I'd like to start a small discussion to this:
image

I never really understood the information here. Actually I thought they were wrong and started to fix them before I realized what they are.
For those who don't know:
- Income
   - Transactions which were sent FROM the contact TO the currently selected account (in my case from `delphin` to `Delphin` (sorry for the similar names :P )
- Expense
  - Transactions which were sent TO the contact FROM the currently selected account (`Delphin` to `delphin`)

I wonder if it would help if we would just change:
- `Income` => `Received From`
- `Expense` => `Sent To`

Though I'm not sure if this would make it clearer?
Maybe we should add a dialog which display the information on-click (we could write clearer there what's the meaning of this all).
Or maybe I'm the only one who finds this confusing anyway?

### DIFF
--- a/client/app/src/components/account/templates/main-sidenav.html
+++ b/client/app/src/components/account/templates/main-sidenav.html
@@ -35,7 +35,7 @@
     <span style="white-space: nowrap;">
       <md-icon md-font-library="material-icons">account_circle</md-icon>
       {{contact.name}}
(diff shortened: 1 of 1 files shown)


j-a-m-l:
Good work!.

About the renaming, I'll rethink this feature:

It's quite useless, but involves a lot of overhead: currently it shows inaccurate data because is using the latest history only.

We should fix first the problem of calculating the amount and the number of transactions, which it's a very common problem when you're voting, or use a different text:

"Received from the selected account: A1 (1 transaction) during the last days"
"Sent to the selected account: A30 (2 transactions) during the last days"

Issue #540, opened 2018-01-27
ArkClient.exe parameter to enable/disable Screenshot protection


fbz76:
Scenario:
When launching ArkClient.exe, the application starts by Defalut with Screenshot Protection Enabled. You can disable it opening the Application top-left Menu.
![image](https://user-images.githubusercontent.com/35872694/35473871-5a5e2356-0386-11e8-9a43-1de85772bb49.png)
Issue: 
In the case you have your server or remote computers running wallets in your own LAN, and you are used to access the server via a remote client (RemoteDesktop, VNCViewer) in your won secure environment, you are not able to view the application when Screenshot Protection is Enabled. See the below example where the black area is the ArkClient.
![image](https://user-images.githubusercontent.com/35872694/35473938-5d4960f2-0387-11e8-8152-59178424a465.png)
This means that you need to connect a physical monitor/keyboard/mouse to your remote server to deal with the ArkClient.
Solutions:
1. Having the possibility to pass to ArkClient a config parameter (i.e, "DisableScreenshotProtection") that is allowing you to start the application: "_ArkClient.exe -DisableScreenshotProtection_"
or
2. Allowing the ArkClient to register the last user choice and apply it again at next ArkClient launch.

Nasicus:
Good idea!

Implemented it via a parameter. Normal users don't need this and this way it cannot be forgotten to turn off ("you need to know what you're doing").

luciorubeens:
Seems good, however this could make the wallet vulnerable

Nasicus:
Why / how do you mean? (Assuming we do it via a parameter).
On Linux we don't even have protection implemented (or is this there ensured in another way?)

luciorubeens:
Someone can install a malicious software to change the shortcut and take screenshots. Currently the option is disabled when restarting/closing the application. But I'm just predicting the worst case scenario..

alexbarnsley:
I agree with Lucio. Someone could make a shortcut/wrapper and replace the original application to launch with the extra argument. Anyway round the protection could be used by someone externally as a way of disabling it themselves.

In Windows for example, I can create a shortcut of an exe, add the --disablescreenshotprotection argument and replace any existing shortcuts with this new one.

@luciorubeens would you say the screenshot protection is to hide accounts, or to hide passphrases, or both?

I think a different route to go down would be to install the wallet on an encrypted USB and use it at the workstation you're at @fbz76, but that's just my opinion.

[...]

PR #542, opened 2018-01-28
Make screenshot protection configurable

Nasicus:
Find it a good input and it annoyed me anyhow that the protection was always enabled while developing, so I used this opportunity to disable it by default when developing.

Since having this disabled is a quite special / technical requirement, I did it via an argument which has to be passed to the client. This way no one can accidentally disable the screenshot protection on his machine.

Resolves: #540

### DIFF
--- a/main.js
+++ b/main.js
@@ -208,6 +208,10 @@ function configureReload () {
   })
 }
 
+function shouldDisableScreenshotProtection (arugments) {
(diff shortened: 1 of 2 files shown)


faustbrian:
@luciorubeens @alexbarnsley @j-a-m-l @fix

Issue #549, opened 2018-02-03
Manage networks view incorrectly adds new network without saving network (cancel does not work)


vRobM:
Manage networks, New: 
Reproduce by adding a New network, populate seed server, click "Create", once it auto-configures, click the "Cancel" button.

The network is already added (Created) even though we did not "Save" the network.

One must then remove the network manually.

Please don't add the network until we click the "Save" button. 
This flow may need to change the "Create" button to "Test" first, and have the "Create" functionality take effect by pressing the "Save" button.

PR #551, opened 2018-02-03
Save network not directly after create was pressed

Nasicus:
This PR solves various things:

- don't add network directly after create was pressed
- an angularjs error which was displayed because `$scope.apply()` was called
- only the "active network" is saved, when save was pressed (before all networks were saved). Imho both approaches are not optimal but I find it more user friendly this way
- fix that position of settings menu changed, when there was no `currency` menu item:
 best bug

Fixes: #549

### DIFF
--- a/client/app/src/services/network.service.js
+++ b/client/app/src/services/network.service.js
@@ -51,10 +51,9 @@
     }
 
     function createNetwork (data) {
-      const n = storageService.getGlobal('networks')
-      let newnetwork = data
+      const networks = storageService.getGlobal('networks')
       const deferred = $q.defer()
-      if (n[data.name]) {
+      if (networks[data.name]) {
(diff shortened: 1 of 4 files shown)


faustbrian:

client/app/src/services/network.service.js
```diff
@@ -63,14 +62,13 @@
...
           (resp) => {
-            newnetwork = resp.data.network
-            newnetwork.forcepeer = data.forcepeer
-            newnetwork.peerseed = data.peerseed
-            newnetwork.slip44 = 1 // default to testnet slip44
-            newnetwork.cmcTicker = data.cmcTicker
-            n[data.name] = newnetwork
-            storageService.setGlobal('networks', n)
-            deferred.resolve(n[data.name])
+            let newNetwork = resp.data.network
```
j-a-m-l: Instead of `let`, using `const` would be better

PR #555, opened 2018-02-05
Add a message on the sending dialog when using a Ledger

j-a-m-l:
To avoid problems such as https://cointelegraph.com/news/newly-discovered-vulnerability-in-all-ledger-hardware-wallets-puts-user-funds-at-risk

### DIFF
--- a/client/app/src/components/account/templates/send-transaction-dialog.html
+++ b/client/app/src/components/account/templates/send-transaction-dialog.html
@@ -22,6 +22,12 @@
         <md-tab-body>
           <form name="singleForm" class="md-dialog-content">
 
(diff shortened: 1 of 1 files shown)


fix:
there is another issue, is that the receiving address might be redacted. So we need to display on device the receiving address to be sure it is not redacted.

If i remember well, the ledger app already allows this, you need to put some switch in the request to display the requested address on device.
https://github.com/ArkEcosystem/ark-desktop/blob/master/LedgerArk.js#L11

@alexbarnsley might need to confirm with ledger on their slack

fix:
see here https://github.com/ArkEcosystem/ark-ledger/blob/master/src/main.c#L752

Issue #559, opened 2018-02-11
 When adding new network and not specifying the name, it gets set to UNDEFINED


vRobM:
Perhaps make this a required field to avoid undesirable network names.

Similar to the seed server field.

The name field does have a * indicating it is required, but it is not being checked properly.

PR #562, opened 2018-02-11
Bug fixes & changes in manage networks

Nasicus:
- check for required fields when adding / updating networks
- make almost all fields required when updating network
- don't open an invalid URL if explorer of a network is not specified

As mentioned I made almost all field in mange network required, but imho this makes sense, correct me if I'm wrong:
image

Fixes: #559, #571

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -373,6 +373,10 @@
     }
 
     function openExplorer (uri) {
+      if (!self.network.explorer) {
+        return
+      }
+
       require('electron').shell.openExternal(self.network.explorer + uri)
     }
 
@@ -1285,10 +1289,18 @@
               }]
             })
           } else {
-            self.listNetworks = networkService.getNetworks()
+          self.listNetworks = networkService.getNetworks()
             toastService.success(gettext('Network removed successfully!'), 3000)
           }
         })
+      }
+
+      function canCreateNetwork (network) {
+        return network && network.name && network.peerseed
+      }
+
+      function canUpdateNetwork (network) {
+        return network && network.token && network.symbol && network.version && network.nethash && network.slip44 && network.peerseed
       }
 
       let activeNetworkIndex = 0
@@ -1307,7 +1319,9 @@
         createNetwork,
         removeNetwork,
         cancel,
-        save
+        save,
+        canUpdateNetwork,
+        canCreateNetwork
       }
 
       $mdDialog.show({
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1228,23 +1228,13 @@
       return self.network.cmcTicker || self.network.token === 'ARK'
     }
 
-    function createInvalidPeerServerToast (peerServer) {
-      toastService.error(gettextCatalog.getString('The Seed Server \'{{ url }}\' is not valid! You must enter a valid http or https URL!',
-                                                  {url: peerServer}))
-    }
-
     function manageNetworks () {
       let networks = networkService.getNetworks()
 
       function save (networkName) {
-        const network = $scope.send.networks[networkName]
-
-        if (!utilityService.isValidUrl(network.peerseed)) {
-          createInvalidPeerServerToast(network.peerseed)
-          return
-        }
-
         $mdDialog.hide()
+
+        const network = $scope.send.networks[networkName]
         delete network.isUnsaved
         networkService.setNetwork(networkName, network)
         self.listNetworks = networkService.getNetworks()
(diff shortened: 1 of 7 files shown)


alexbarnsley:
Looks good - not tested yet as had 1 possible change. Maybe in the canCreate and canUpdate methods, add a check for the URL always having "http://" or "https://". I don't think it works at all without either. Maybe a URL validator of some sort, `http://1.2.3.4:1234` = valid; `http://1.2.3.4:` = invalid, etc

Nasicus:
Right. I wanted to to that but forgot again.
Where would you show when the URL is valid? as a toast and leave the window open?
one problem is, that I cannot easily / simply set it on the `send` variable, because it's shared between all tabs. Showing a toast when clicking "Save" would fix that.

alexbarnsley:
Yeh I'm happy with a toast and leaving the window open until they resolve it 👍 this might be a good area for a few unit tests to make sure the validation continues to work, do you agree?

Nasicus:
You mean unit tests for the urlValidateMethod or for a general validate method (which doesn't exist yet )?

alexbarnsley:
I think the urlValidateMethod - checking if tests pass for valid URLs and invalid URLs. Well I guess you could mix them together couldn't you in some unit tests? I won't pretend I'm any good at writing tests though xD @j-a-m-l is our unit test master!

Nasicus:
@alexbarnsley 
Check out my last commit.
Note that I used an NPM package* and the tests may be a little redundant therefore, however I find them still nice to have.
I tested only what _we_ think should be a valid URL, obviously there are a lot of other test cases which you could have.

* But Patrick, why the hell do you use an NPM package for something that simple?
Well my friends. I had the same thought. Really. I googled it and found this [post](https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/14582229#14582229). However neither the accepted or the `138` upvoted answer work correctly 🚎

j-a-m-l:
@Nasicus this looks good. Would you mind to resolve the conflicts?

Nasicus:
@j-a-m-l  done

alexbarnsley:
@Nasicus apparently `http://sdf.:3000` is a valid URL 🤔 i've not come across it in the past as valid. I had the same with `http://:3000` or `http://:`. It still fails to add the network which I'd expect, but the error message doesn't show as an invalid URL, but for "Error: Cannot connect to peer to autoconfigure the network". 
I tried adding `it('wut', () => expect(utilityService.isValidUrl('http://:3000')).to.eq(false))` as a test and the test failed, thinking it is valid. Is it just me? 😆

Nasicus:
@alexbarnsley

Wow. Apparently not even an npm package which is only dedicated to check valid urls works correctly then.
Well I can look if I find something better when I have access to a pc again. Which will probably be on saturday.

zillionn:
Why don't you try `input type="url"` and then `form.input.$valid` ?

Nasicus:
Because I didn' even know that this exists ;)
Will try that - thanks.

Nasicus:
So I added the `input type="url"` suggested by @zillionn . However this also allows some urls which are not allowed, saw that you can use `pattern` (which is a regex) to do additional validation. I used this and now _most_ (not all) invalid urls are detected.

Also reverted my other commit.

It's not perfect, but I also don't see why I or the reviewer should invest more and more time into this tbh :P

zillionn:
> However this also allows some urls which are not allowed

For instance?

Nasicus:
> For instance?

`fasdfsdfa:2323`  or `ftp://whatever` or `http://sdf.:3000` (however this URL is still valid even with the `pattern`)

zillionn:
I believe they are all valid URIs:
https://tools.ietf.org/html/rfc3986#section-1.1.2

Nasicus:
@zillionn 
Well yes probably. But not for our use case (it must always be a http / https url), hence the additional `pattern` :)

zillionn:
OK, I'm not really sure what you're trying to do. Is this the "Seed Server" URL you're trying to validate?

Nasicus:
Yes @zillionn  :)

zillionn:
Alright then. Why don't you first check `form.input.$valid` and then try to access the API at `form.input.value + '/api/loader/autoconfigure'`? If the API returns JSON with `success: true`, you'll know the URL is valid.

zillionn:
🎉

alexbarnsley:
Hi @Nasicus. Sorry for the slow response. Also, don't kill me! I have 1 last check xD for whatever reason, putting a forward slash (/) at the end always fails. E.g. `http://107.191.39.65:4001` works, `http://107.191.39.65:4001/` doesn't. I guess here, could we get it to work with or without a forward slash. Like a browser would? 🤷‍♂️ 
I'll merge once happy with that!

Nasicus:
done

alexbarnsley:
Wouldn't do multiple forward slashes, but i'm happy with that as a fix 👍 thanks!

Issue #560, opened 2018-02-11
Long delay (visually) before anything happens when selecting a new network


vRobM:
Switching from MAINNET to DEV for example, makes one think the action did not go through, so while you click the menu again.. the delay expires and refreshed the entire screen.

Makes for poor UX.

Indicate something is happening or speed up switching.

Nasicus:
Just a side note: If you have a fast connection & the peers are not slow, you barely ever notice this. However I agree that it's still nice to show a loading indicator.

PR #563, opened 2018-02-11
Add loading indicator when switching network

Nasicus:
- implement it so that it can also be used in other places

To test:
- go to the `network.service.js` file.
- replace `return window.location.reload()` with `setTimeout(() => return window.location.reload(), 10000)` in the method `switchNetwork`.

fixes: #560


j-a-m-l:
Closed until knowing more about the issue

j-a-m-l:
I don't see any loading indicator.

Issue #565, opened 2018-02-14
Removing a Network does not switch to primary Network


vRobM:
Upon removal; it should auto switch to the first network (main net) instead of showing a zombie network that doesn't exist anymore.

PR #566, opened 2018-02-14
Switch the network if removing the active network

Nasicus:
Also fixes a bug related to the background.

```
dependencies.min.js:1 TypeError: Cannot read property 'match' of undefined
    at scope.$watch (background-style.directive.js:40)
    at Scope.$digest (dependencies.min.js:1)
    at Scope.$apply (dependencies.min.js:1)
    at dependencies.min.js:1
    at Object.invoke (dependencies.min.js:1)
    at doBootstrap (dependencies.min.js:1)
    at bootstrap (dependencies.min.js:1)
    at angularInit (dependencies.min.js:1)
    at dependencies.min.js:1
    at HTMLDocument.trigger (dependencies.min.js:1)
```

Fixes: #565

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1277,10 +1277,21 @@
         $mdDialog.show(confirm).then(() => {
           networkService.removeNetwork(network)
           if (isActive) {
-            dialogService.openLoadingDialog(self.currentTheme,
-                                            gettext('Network removed successfully'),
-                                            gettext('Network removed successfully - will now switch the network'))
-            setTimeout(() => networkService.switchNetwork(null, true), 3000)
+            $mdDialog.show({
+              parent: angular.element(document.getElementById('app')),
+              templateUrl: './src/accounts/view/switchNetworkDialog.html',
+              clickOutsideToClose: false,
+              escapeToClose: false,
+              fullscreen: true,
+              locals: {
+                theme: self.currentTheme,
+                switchNetwork: () => networkService.switchNetwork(null, true)
+              },
+              controller: ['$scope', 'switchNetwork', 'theme', ($scope, switchNetwork, theme) => {
+                $scope.switchNetwork = switchNetwork
+                $scope.theme = theme
+              }]
+            })
           } else {
             self.listNetworks = networkService.getNetworks()
             toastService.success(gettext('Network removed successfully!'), 3000)
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1284,10 +1284,12 @@
               escapeToClose: false,
               fullscreen: true,
               locals: {
-                theme: self.currentTheme,
-                switchNetwork: () => networkService.switchNetwork(null, true)
+                networkName: network,
+                switchNetwork: () => networkService.switchNetwork(null, true),
+                theme: self.currentTheme
               },
-              controller: ['$scope', 'switchNetwork', 'theme', ($scope, switchNetwork, theme) => {
+              controller: ['$scope', 'networkName', 'switchNetwork', 'theme', ($scope, networkName, switchNetwork, theme) => {
+                $scope.networkName = networkName
                 $scope.switchNetwork = switchNetwork
                 $scope.theme = theme
               }]
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -1278,12 +1278,12 @@
           networkService.removeNetwork(network)
           if (isActive) {
             dialogService.openLoadingDialog(self.currentTheme,
-                                            gettext('Network removed succesfully'),
-                                            gettext('Network removed succesfully - will now switch the network'))
-            setTimeout(() => networkService.switchNetwork(null, true), 4000)
+                                            gettext('Network removed successfully'),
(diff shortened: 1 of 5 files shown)


Nasicus:
added one more commit which fixes: #568

client/app/src/accounts/account.controller.js
```diff
@@ -1275,14 +1276,31 @@
           .cancel(gettextCatalog.getString('Cancel'))
         $mdDialog.show(confirm).then(() => {
           networkService.removeNetwork(network)
-          self.listNetworks = networkService.getNetworks()
-          toastService.success(gettext('Network removed succesfully!'), 3000)
+          if (isActive) {
+            dialogService.openLoadingDialog(self.currentTheme,
+                                            gettext('Network removed succesfully'),
+                                            gettext('Network removed succesfully - will now switch the network'))
```
alexbarnsley: "successfully" is spelt incorrectly. Also, i'm not sure how much work it would be, but maybe changing it so it's not a loading dialog but a dialog that tells them the wallet needs to reload, with a reload button. then it just reloads on close 🤷‍♂️

client/app/src/accounts/account.controller.js
```diff
@@ -1275,14 +1276,31 @@
           .cancel(gettextCatalog.getString('Cancel'))
         $mdDialog.show(confirm).then(() => {
           networkService.removeNetwork(network)
-          self.listNetworks = networkService.getNetworks()
-          toastService.success(gettext('Network removed succesfully!'), 3000)
+          if (isActive) {
+            dialogService.openLoadingDialog(self.currentTheme,
+                                            gettext('Network removed succesfully'),
```
alexbarnsley: "successfully" is spelt incorrectly

client/app/src/accounts/account.controller.js
```diff
@@ -1275,14 +1276,31 @@
...
+                                            gettext('Network removed succesfully - will now switch the network'))
+            setTimeout(() => networkService.switchNetwork(null, true), 4000)
+          } else {
+            self.listNetworks = networkService.getNetworks()
+            toastService.success(gettext('Network removed succesfully!'), 3000)
```
alexbarnsley: "successfully" is spelt incorrectly

client/app/src/accounts/account.controller.js
```diff
@@ -1275,14 +1276,31 @@
           .cancel(gettextCatalog.getString('Cancel'))
         $mdDialog.show(confirm).then(() => {
           networkService.removeNetwork(network)
-          self.listNetworks = networkService.getNetworks()
-          toastService.success(gettext('Network removed succesfully!'), 3000)
+          if (isActive) {
+            dialogService.openLoadingDialog(self.currentTheme,
+                                            gettext('Network removed succesfully'),
+                                            gettext('Network removed succesfully - will now switch the network'))
+            setTimeout(() => networkService.switchNetwork(null, true), 4000)
```
alexbarnsley: Maybe halve the timeout as felt a bit long for me, it was enough time for me to wonder if it was going to reload.

alexbarnsley: oh maybe take a look at my other comment. instead of doing a timeout, https://github.com/ArkEcosystem/ark-desktop/pull/566#discussion_r169040742

client/app/src/accounts/account.controller.js
Nasicus: Well problem is, that this is a generic loading dialog which I did either reuse here or reuse in another place :P I find for this situation it's enough like that (because almost nobody ever deletes a network) - but of course in the end it's your decision :)

client/app/src/accounts/account.controller.js
Nasicus: done

client/app/src/accounts/account.controller.js
Nasicus: done

client/app/src/accounts/account.controller.js
j-a-m-l: A simple solution would be adding a new parameter to `openLoadingDialog` that permits customizing its behaviour or content.

2 examples:
 - Using `templateUrl`, to add there the reload button.
 - Passing a `options` parameter that enables HTML instead of pure text (and change the template too).

Nasicus: I guess I just fail to see the advantage of showing a button to the user. He cannot do anything anyway - he HAS to reload because the network was deleted and the dialog is modal.

alexbarnsley: To me the UX didn't feel seamless. I'd rather a button I have to press to reload, because then I can press it in my own time. Loading screens I feel are used for processing in the background or waiting for something to happen, and not to display a message to the user, because I might be a slow reader who didn't manage to read the whole message before it disappeared.

Nasicus: Okay, that's a valid point.
I will adjust it accordingly.

Nasicus: Done - what do you think?

image

alexbarnsley: Looks good! I'll have a cheeky test Monday

alexbarnsley:
Looks good and working as expected. 👍

PR #573, opened 2018-02-19
Sort the languages alphabetically (in the application).

trigger67:
Better to sort languages correctly in the application ;)
I let English first, maybe it should also be moved.


luciorubeens:
Hey @trigger67, thanks for your PR. But if you want to sort languages **correctly**, don't restrict this to English only. You can create a filter to handle the translated list..

trigger67:
Hey @luciorubeens, the sort is not restricted to English as languages are never translated.
But you are right a filter would be better.

Nasicus:
@trigger67 
This is actually a bug - since there are even translations for them and the `gettext` method is called.
If you want you can fix it - if you don't know how or don't want to I can help you or do it myself.

trigger67:
@Nasicus 
I do not see where the gettext method is called. Can you tell me ?

There is a ng-repeat on selectAllLanguages() with:

self.selectAllLanguages = function () {
      return languages
    }

Also "languages attribute" is not an array, but an object with one attribute per language. Reflection is done on the object to set the language... This is quite complexe just for that. I assume there is a reason why it is done like this.

Also I am experienced with Angular (version >= 2) but not much (at all) with AngularJs.
So I can try to fix it, but I would probably need some help :)

Should I do some refactoring to use something like an array of an object with 2 properties (like name and isoCode) ?
Should I add a getText() for the translation ?
Should I do the ordering  in the selectAllLanguages() method ?

Thank you for your time. If it is too long to explain I agree if you fix it, but I would like to give it a try ;)

Nasicus:
Well the translate method is called for each language directly:
`en: gettextCatalog.getString('English'),`

However the problem is, that the language for the text service is set _after_ the `const languages` object is initialized (lines 140 - 142).

Pretty sure the translations will work if you just move these three lines up, before the declaration of the `languages` object.

I think your idea to changing it to an array with objects inside is good.
Because we almost never need to access the value by key directly anyway (only one time I think).
If you change it, then you can simply sort it directly after declaring it plus the setting of a new language will probably also get easier...

trigger67:
Alright, I struggled with some "refresh" of components, but now **the translation of the languages are working**. And I refactored some code, and use a list instead of an object.

Also there were a bug, previously the method selectAllLanguages was **constantly called !** Put a debugger in it and you will see. This one is fixed also.

But I did not ordered the language, because when using "gettextCatalog.getString()" to order my list, it was causing some refresh issue also. The list was not well refresh on the screen, causing some "blinks" also. Maybe you can accept the pull request, for fixing the translation issue + the bug of method called indefinitly. Or I would need some help for the "ordering" (or a good sleep)...

Nasicus:
Sounds & looks quite good already. 👍 

There are a few things which caught my eye:

If you declare the languages like you did, they will not get caught by our "translatable" extractor.
This means you either have to call `gettext('English')` (which is only a marker method and does nothing) or translate them directly - as it was before via `gettextCatalog.getString('English')`. I would do the second option.

Why?

Because then sorting will be a piece of cake, you can simply do it like this then:
```
self.languages = [{ name: gettextCatalog.getString('English'), code: 'en' },....].sort((a, b) => a.name.localeCompare(b.name))
```

Of course you also have to move these three lines up (before the declaration of the languages, or the languages will once again not be translated):

```
    self.getLanguage = function () {
      return storageService.get('language') || 'en'
    }

    self.language = self.getLanguage()
    gettextCatalog.setCurrentLanguage(self.language)
```

One more thing which comes to my mind is also, that if you do it like I said and you change the language it will be probably only translated after the app starts.
But this could also be easily fixed if you create a method for initializing the languages (the first code block which I posed) and call this method once directly (where currently the languages are initialized) and a second time when the user changes his language.

Sorry for the long post :P

trigger67:
Hi Nascius,

Thank you for the feedback !

I only read the first part of your comment and when I finished developing it I saw that I did what you describe in the 2nd half :) So I did exactly as you said and it works like a charm ;)

j-a-m-l:
Good work @trigger67!

Snd thanks, @Nasicus for helping him :-)

trigger67:
Thanks Juan ;)

PR #574, opened 2018-02-19
Fix an error in the translation name. Not a translation error, but an…

trigger67:
… error in the name of the field. So it needs a pull request.


luciorubeens:
Thanks, but we are managing translations on OneSky. I can't merge it here because it will be overwritten by the translation made there. [See more](https://github.com/ArkEcosystem/ark-desktop#pinned-help-us-with-translations).

PR #575, opened 2018-02-20
Fee override from the network

alexbarnsley:
This is in conjunction with https://github.com/ArkEcosystem/ark-js/pull/53 whereby the fees can be customised on bridgechains. With the desktop wallet, we already fetch the fee from the network, but we're now passing those fees into the create transaction methods to correctly build a transaction object.

@luciorubeens @j-a-m-l if you could take a look and have a quick test, that would be great! Thanks


alexbarnsley:
Thanks guys!

PR #558, opened 2018-02-10
Standardized menu and shortcut to match the system’s

gabrielbull:
Fixes issue #557 and standardizes application's menu to match the system.

<sub><sub><sub>My own approval.</sub></sub></sub>
&nbsp;&nbsp;&nbsp;&nbsp;↓

PR #576, opened 2018-02-21
Fix broken reload command on Windows

Nasicus:
With #558 happened some changes in the menu.

On windows (can't speak for other systems) there is now a bug that the `Reload` command doesn't work anymore via keyboard - this PR fixes this.

However I also saw that some commands which were previously there are now missing, I don't know if they were omitted or changed on purpose, so I didn't change anything yet:

- `About ArkClient` became `About Electron` 
  - pretty sure this is wrong and should be changed back right?
- `Maximize` is missing
- `Full Screen` is missing
- `Quit` was removed from `Application` and is now instead called `Close` beneath `Window`
  - I personally liked `Quit` more below application (or both)

@gabrielbull  @alexbarnsley

### DIFF
--- a/main.js
+++ b/main.js
@@ -184,7 +184,7 @@ function createWindow () {
           label: 'Learn More',
           click () { require('electron').shell.openExternal('https://ark.io') }
         },
-        { label: 'Reload App', accelerator: 'Command+R', click: function () { mainWindow.reload() } },
+        { label: 'Reload App', accelerator: 'CmdOrCtrl+R', click: function () { mainWindow.reload() } },
         { label: 'Open Dev Tools', accelerator: 'CmdOrCtrl+D', click: function () { mainWindow.webContents.openDevTools() } }
       ]
     }


alexbarnsley:
I'll do a screenshot of mine on Mac later. I had the full screen and maximize options. I ignored the stuff about Electron because whenever I run it from source, it says electron. Will build an actual release and test that. Give me an hour to do that screenshot

Nasicus:
@alexbarnsley ah I compared it to the productive version - it's very possible that the thing about Electron is not an issue then!

Nasicus:
This is the before & after on windows:

image

alexbarnsley:
For some reason the options that are missing are only available on Mac based on the #558 PR. I'm happy for you to re-add them so it matches how it was before, unless @gabrielbull can shed light on why he is only showing them for Mac?

gabrielbull:
The about menu shows the application name, when running using the `npm start`, the application name is always __Electron__. That is just how Electron works, it will get switched to the application name in the final build.

As for the per system menu differences, I used the standard template for each system. My computer is in repair so I can't test right now, but I believe it matches other Windows applications. For instance, if you open Internet Explorer and show the menu, what is shows under Application, Edit, Window and Help? Is there a quit option? If the quit option is there in other windows applications I might be wrong.

I believe it's best to use standard practices per system, which is why I used the same [template as Electron](https://electronjs.org/docs/api/menu#main-process).

Nasicus:
Great news concerning the name.

Well internet explorer has `File` as first entry, if you click it there is a `Exit` option. Same for Slack for example.

I personally find it strange that only `Minimize` is there, but no `Fullscreen` and no `Maximize`. But I doubt that anyone actually uses that....same goes for "Exit".

I guess @alexbarnsley or the other devs needs to decide what they want. I don't really care (ctrl + r works again, this is important :D )

gabrielbull:
Then I think that we should match what Slack, IE, and other apps have. So the first entry on Windows should be File, not Application, and should include the Enable Screenshot Protection, a separator and Exit (not quit).

alexbarnsley:
I think if we can match it as much as we can to what it was before, but keeping the newer way of how it's done. Otherwise people who do use those features may wonder why they're gone. I'm happy for stuff to be renamed, moved, etc. It's more the shortcuts which people may rely on.

Not sure if my answer helps xD

gabrielbull:
Sounds good, so I made another PR containing those changes. You'll have to merge this open before.

alexbarnsley:
Nice work on correcting the reload shortcut 👍

PR #593, opened 2018-05-11
Generate and import passphrase in all possible bip39 languages

air1one:
PR related to https://github.com/ArkEcosystem/ark-mobile/pull/148

To make it compatible.

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -118,6 +118,27 @@
 
     self.setLanguage()
 
+    self.getWordlistLanguage = function () {
+      return storageService.get('wordlistLanguage') || 'english'
+    }
+
+    self.setWordlistLanguage = function () {
+      storageService.set('wordlistLanguage', self.wordlistLanguage)
+    }
+
+    self.wordlistLanguages = {
+      'english': 'English',
+      'french': 'French',
+      'spanish': 'Spanish',
+      'italian': 'Italian',
+      'japanese': 'Japanese',
+      'korean': 'Korean',
+      'chinese_simplified': 'Chinese simplified',
+      'chinese_traditional': 'Chinese traditional'
+    }
+
+    self.wordlistLanguage = self.getWordlistLanguage()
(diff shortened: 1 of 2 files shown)


air1one:
Hi @alexbarnsley so here it is, tell me what you think !

client/app/src/accounts/account.controller.js
```diff
@@ -1654,7 +1675,7 @@
 
     function isBIP39 (mnemonic) {
       const bip39 = require('bip39')
-      let valid = bip39.validateMnemonic(mnemonic)
+      let valid = bip39.validateMnemonic(mnemonic) || bip39.validateMnemonic(mnemonic, bip39.wordlists[self.getWordlistLanguage()])
```
zillionn: Why not just `let valid = bip39.validateMnemonic(mnemonic, bip39.wordlists[self.getWordlistLanguage()])` ? Default is English, for other languages you'll run `validateMnemonic()` twice, which is not necessary.

air1one: Hi @zillionn thanks for your review, I explained this particular behavior on the ark-mobile PR : 
"For wallet import, entering a passphrase in english will still work, even if we select a different language in "Wordlist language". (we might want to generate wallets in a specific language but import a previous wallet which was generated in english)"

But I think I could test 1st the mnemonic in the defined language (changing the order in the "or" condition). Because if the language defined is different than English, there's a higher probability that we would import in this language.

What do you think ? And does it make sense to you to validate in English too even if we defined a different wordlist language ?

Thanks!

zillionn: I see. It's not up to me, I was just curious.

air1one:
How about "Passphrase Language" ?

luciorubeens:
Nice! Thank you :+1:

luciorubeens:
@air1one Can you replicate to your PR in the mobile repo?

air1one:
Yep done already @luciorubeens 👍

PR #600, opened 2018-06-15
[WIP] Changelly

j-a-m-l:


client/app/src/components/account/exchange-tab.controller.js
```diff
@@ -16,9 +16,58 @@
...
+      let amount
+
+      if (this.ul.network.token === 'DARK') {
+        from = 'btc'
+        amount = 0.1
+      } else {
+        to = this.ul.network.token
+
+        // Use the current address to receive the ARKs
+        if (to === 'ARK') {
```
luciorubeens: This can be removed

client/app/src/components/account/exchange-tab.controller.js
```diff
@@ -16,9 +16,58 @@
...
-  function ExchangeTabController ($scope) {
+  function ExchangeTabController ($scope, $sce) {
     this.$onInit = () => {
       this.ul = this.accountCtrl
+
+      const merchantId = 'bab9de3731aa'
+      const refId = merchantId
+
+      let from
+      let to = 'ARK'
```
luciorubeens: Hm, maybe we could use `this.ul.network.cmcTicker`

client/app/src/components/account/exchange-tab.controller.js
```diff
@@ -16,9 +16,58 @@
...
+        }
+
+        let pseudo = 500 * Math.random()
+        pseudo = pseudo < 100 ? 100 : pseudo
+
+        let price
+
+        if (this.ul.btcValueActive) {
+          from = this.ul.bitcoinCurrency.name
+          price = this.ul.connectedPeer.market.price[from]
```
luciorubeens: Now you need to use the marketService: `marketService.getPrice(from).price`

j-a-m-l:
I've simplified how the amount is calculated: Changelly, due volatility probably, offers a very different price than CoinMarketCap, so it's probably better suggesing a fixed amount (0.1BTC, 300$ or 300€).

client/app/src/components/account/exchange-tab.controller.js
j-a-m-l: This was a way to make the wallet slightly more compatible to other ARK-based chains

Issue #603, opened 2018-06-28
Privacy concerns with Changelly integration


katomm:
Hello,
as soon as the wallet opens, it looks like it does load the iframe of the Changelly widget. (Even if you do not want to use it or click the tab) 

If an user has some kind of personal firewall installed it will look like the wallet opens connections to:
 google.com
 ssl-google.analytics.l.google.com
 platform-lookaside.fbsbx.com 
 facebook.com
 yandex.ru
 doubleblick.net
 and maybe others
 
 Not something you would expect from a wallet - jeopardising users trust. A better way would be ~~removing Changelly completely~~ not loading the iframe until someone clicks on the tab.

j-a-m-l:
This was fixed by #604. In some minutes/hours it'd be available.

PR #604, opened 2018-06-30
Do not load the Changelly widget unless the user confirms it

j-a-m-l:
This implements a way to avoid privacy issues (https://github.com/ArkEcosystem/desktop-wallet/issues/603).


j-a-m-l:
Well, it's a subtle behaviour: the `localStorage.set` function, deletes the entry when its value is false-like.

I've removed it completely because it's better that the users decide every time they open the wallet.

PR #606, opened 2018-07-10
Warn about NEO addresses only when they have transactions

j-a-m-l:
On the send transaction dialog, it's more common than expected that addresses are flagged as NEO.

The problem was that our current implementation was expecting that the NEO API endpoint returned a different response when NEO addresses have transactions. Instead it was asking about the unclaimed GAS.

### DIFF
--- a/client/app/src/services/neo-api.service.js
+++ b/client/app/src/services/neo-api.service.js
@@ -1,40 +1,33 @@
 ;(function () {
   'use strict'
 
+  /**
+   * Example of NEO address with transactions: AYJXuFBfUyELoTe3wXEzPkq5RcsHsF58uW
+   */
   angular.module('arkclient.services')
     .service('neoApiService', ['$q', '$http', NeoApiService])
 
   function NeoApiService ($q, $http) {
-    const ark = require(require('path').resolve(__dirname, '../node_modules/arkjs'))
     const baseUrl = 'https://neoscan.io/api/main_net/v1'
 
-    /*
-      returns {"unclaimed": <value>, "address":"<address>"}
-    */
-    function getUnClaimed (address) {
-      const deferred = $q.defer()
-      $http.get(baseUrl + '/get_unclaimed/' + address)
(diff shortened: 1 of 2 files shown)


Nasicus:
@j-a-m-l  in mobile I implemented it differently:
https://github.com/ArkEcosystem/mobile-wallet/blob/master/src/providers/neo-api/neo-api.ts

I don't know how fast your API request is, by the time I made my mobile changes my change there was very fast.
Maybe we can do the same call in both places.

Back then I didn't create a PR here because it's not allowed anymore ;)

luciorubeens:
Nasicus is alive 🎉

boldninja:
Is he or did someone took over his cool identity :thinking:

j-a-m-l:
Thanks @Nasicus for you feedback.

That implementation may work in almost all cases, but what about when an address is completely emptied? Checking that an address has interacted with the blockchain through transactions would ensure that the address exists.
I'm not a NEO expert, so if that's not right, please, correct me.

Nasicus:
@j-a-m-l  tbh I don't remember anymore :P 
According to my comment it should work, but I don't know anymore - I 'd need to test it too :)

```
    // we use the getBalance call, because it's fast, if the address exists (i.e. has any transactions), the address is returned
    // we check if it's a real address (and not "not found") and return the result
```
I just thought that we may want to unify the both places, so the same code  is in place (can also be your solution).

@luciorubeens 
😎 
@boldninja 
🕶 

PS: Was in holiday on may, and since then I was doing an own little side project (not blockchain related) plus worlcup + playing a lot of games and stuff ;) I will come back once I find time and motivation 💃 ;)

boldninja:
proud

Nasicus:
@j-a-m-l 

Made some [tests ](https://www.getpostman.com/collections/5e6a3c91156e771a5429)with postman.

Turns out both calls behave as expected.
However: Your call is way faster then the one I use! So we should definitely go with your solution 👍

PR #611, opened 2018-07-17
Fix the signing of transactions using the Ledger

j-a-m-l:
Seems that https://github.com/ArkEcosystem/ark-js/pull/85 provoked a problem when signing transactions with the Ledger.

### DIFF
--- a/client/app/src/accounts/account.controller.js
+++ b/client/app/src/accounts/account.controller.js
@@ -309,7 +309,6 @@
         return
       }
       if (!self.ledgerAccounts && self.ledger && self.ledger.connected) {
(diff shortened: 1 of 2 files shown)


Sbphillips19:
Still having issues with this with my ledger, receiving the following in a red box in the lower left hand corner "TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object."

j-a-m-l:
@Sbphillips19 if you're having a problem, could you create a new issue with the most detailed explanation? Mind that things, such as operating system and wallet version are important.

PR #652, opened 2018-12-04
fix: make it possible to input small numbers into the input field

ItsANameToo:
## Proposed changes

You couldn't input 0.0000001 as it would be turned into 1e-7:

schermafbeelding 2018-12-04 om 20 26 58

Now it's fixed and possible to input the correct amount:

schermafbeelding 2018-12-04 om 21 01 43

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Input/InputCurrency.vue
+++ b/src/renderer/components/Input/InputCurrency.vue
@@ -272,7 +272,10 @@ export default {
     updateInputValue (value) {
       // Ignore empty and not valid values
       if (value && this.checkAmount(value)) {
-        this.inputValue = value.toString().replace(',', '.')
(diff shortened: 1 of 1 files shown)


zillionn:
You could also restrict it to only positive numbers, because if balance is `0` and click `Send All` it shows `-txfee`.

ItsANameToo:
Good point, will probably do that in a separate PR tho

j-a-m-l:
Wouldn't it be better doing both things here? :thinking: To avoid checking the same thing 2 times

ItsANameToo:
No, I get paid per PR for bounties and I don't think that change has to do with this PR as it would probably be a check on the switch and nothing to do with populating the value of the input field

PR #653, opened 2018-12-04
feat: make it possible to remove the set name of a wallet

ItsANameToo:
## Proposed changes

It wasn't possible to remove the name of a wallet, as you would get an error saying that `''` was already taken as a wallet name. This PR makes it possible to rename a wallet to `''` again so it reverts to the original address (or delegate / contact name).

## Types of changes

## Checklist


j-a-m-l:
@ItsANameToo where is it failing? I've created a wallet with a name and I can rename it to "" (empty).

ItsANameToo:
Add a wallet without a name, that one will be stored as `''`, then add another wallet with a name, then try to rename that second wallet to one without name and it will fail as there is already a wallet with `''` available in the store.

PR #655, opened 2018-12-04
style: follow the ESlint suggestions

j-a-m-l:
## Proposed changes
These are the changes that `yarn lint:fix` is suggesting.

## Types of changes


ItsANameToo:
Why do we suddenly change from style? In the old repo these "suggestions" were not allowed

ItsANameToo:
OK I take that back as I probably didn't look well 🤔

ItsANameToo:
but if I run `yarn lint:fix` I have no changes, so I don't understand

ItsANameToo:
What do you guys think of the components being renamed to CamelCase with this PR @alexbarnsley @luciorubeens? (@faustbrian?) Even `<router-link>` is renamend to `<RouterLink>` which seems odd as that's not used in the Vue docs

luciorubeens:
Honestly, I prefer the kebab-case for builtin components. But it looks like CamelCase will become standard after they've added the [new rule in eslint](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/component-name-in-template-casing.md). So looks good to me.

dated:
> Honestly, I prefer the kebab-case for builtin components. But it looks like CamelCase will become standard after they've added the [new rule in eslint](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/component-name-in-template-casing.md). So looks good to me.

vuejs styleguides on the matter for those interested https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended

PR #660, opened 2018-12-05
feat: persist unconfirmed transactions

### LINKED ISSUE #644: Persist unconfirmed transactions, and clear if past the core expired time
**Describe the bug**
If a transaction is sent and the fee is too low, it never disappears from the app. Also, unconfirmed aren't persisted so a reload of the app will remove all unconfirmed.

**Expected behavior**
- Persist and show unconfirmed after reload
- Remove unconfirmed if past the core pool expiry time

alexbarnsley:
## Proposed changes

(ignore the branch name - i'm considering this a feature instead of a fix)

The app now has a hardcoded 45 minute value for how long it takes for transactions to expire. This is set in core, and there's no endpoint that provides this value. I thought about adding a network modal field, but I didn't want it to confuse users when adding custom networks. It can be added in the future if it causes issues.

To sum up, an unconfirmed transaction is stored in the local database. This remains until the transaction is processed and received, or the 45 minute period ends. At this point it is removed from the local database. If a wallet is in view with a transaction that expires, it will get a ~strikethrough~ and remain the unconfirmed grey.

Resolves #644

## Types of changes

## Checklist

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -1,5 +1,6 @@
 exports.APP = {
-  website: 'https://ark.io'
+  website: 'https://ark.io',
+  transactionExpiryMinutes: 45
 }
 
 exports.NETWORKS = [
--- a/__tests__/unit/services/synchronizer/wallets.spec.js
+++ b/__tests__/unit/services/synchronizer/wallets.spec.js
@@ -27,6 +27,8 @@ describe('Services > Synchronizer > Wallets', () => {
             return walletUpdate(action, data)
           } else if (action === 'transaction/deleteBulk') {
             return transactionDeleteBulk(action, data)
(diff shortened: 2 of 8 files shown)


alexbarnsley:
spelt "persist" wrong in my commit message 🤦‍♂️

config/index.js
```diff
@@ -1,5 +1,6 @@
 exports.APP = {
-  website: 'https://ark.io'
+  website: 'https://ark.io',
+  transactionExpiryMinutes: 45
```
j-a-m-l: Maybe this can be set as a network property, so, when core provides the configuration, the transition would be straight

Issue #632, opened 2018-12-03
Total Balance doesn't account for Ledger wallet's balances


vmunich:
**Describe the bug**
The Total Balance on the top of the wallet doesn't account for Ledger wallet's balances.

**Expected behavior**
It should be a sum of the balance of every wallet loaded on the Desktop Wallet.

**Screenshots**
![screenshot](https://i.gyazo.com/3f567dec3837fc2783b80533b61aeff0.png)

**Server (please complete the following information):**
 - OS: OSX 10.14.1

sleepdefic1t:
#634 also notes Ledger address balances as ```0```.

PR #665, opened 2018-12-06
fix: include the Ledger wallets when computing the total balance

j-a-m-l:
## Proposed changes
This PR includes the balance of the Ledger wallets on the wallets page and the profiles page. On the profiles page it only includes the balance of the wallets of the current network.

Resolves https://github.com/ArkEcosystem/desktop-wallet/issues/632.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/store/modules/profile.js
+++ b/src/renderer/store/modules/profile.js
@@ -1,4 +1,4 @@
-import { map } from 'lodash'
+import { map, uniqBy } from 'lodash'
 import crypto from 'crypto'
 import BaseModule from '../base'
 import ProfileModel from '@/models/profile'
@@ -14,6 +14,22 @@ export default new BaseModule(ProfileModel, {
     balance: (state, _, __, rootGetters) => id => {
       const wallets = rootGetters['wallet/byProfileId'](id)
       return wallets.reduce((total, wallet) => {
+        return total + wallet.balance
+      }, 0)
+    },
+    balanceWithLedger: (state, _, __, rootGetters) => id => {
+      let wallets = rootGetters['wallet/byProfileId'](id)
+
+      // Only include the wallets of the Ledger that are on the same network than the profile
+      const profile = rootGetters['profile/byId'](id)
+      if (profile.networkId === rootGetters['session/network'].id) {
+        wallets = [
+          ...wallets,
+          ...rootGetters['ledger/wallets']
+        ]
+      }
+
+      return uniqBy(wallets, 'address').reduce((total, wallet) => {
         return total + wallet.balance
       }, 0)
     }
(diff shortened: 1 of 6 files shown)


src/renderer/store/modules/profile.js
```diff
@@ -16,6 +16,22 @@ export default new BaseModule(ProfileModel, {
       return wallets.reduce((total, wallet) => {
         return total + wallet.balance
       }, 0)
+    },
+    balanceWithLedger: (state, _, __, rootGetters) => id => {
+      let wallets = rootGetters['wallet/byProfileId'](id)
+
+      // Only include the wallets of the Ledger that are on the same network than the profile
+      const profile = rootGetters['profile/byId'](id)
+      if (profile.networkId === rootGetters['session/network'].id) {
```
alexbarnsley: Might it be simpler to just do this based on current profile only? Maybe move it to the session store instead? Ledger only loads wallets for the current profile, not all profiles. It reloads wallets when switching profile

j-a-m-l: Yes, this should be changed, but I'm not sure if I'd keep it or not, because I was thinking on a different approach: on the profiles page, show the balance of all the profile networks that are available. What do you think @alexbarnsley?

alexbarnsley: I thought it already showed the balance of each network already? 🤔: 

image

alexbarnsley: Saying that, is this method used when working out those totals when including the ledger? If so, it makes sense to keep this method

j-a-m-l: It is included to display the total of all the wallets on the wallets page and, on each profile of the profiles page (the total of all profiles is calculated differently).

About the total of each network, I'm talking about having 2 Ledger wallets on mainnet with 200ARK, and 3 Ledger wallets with 3000DARK. If you are using a profile that connects to mainnet, you would see 200ARK, but not the other 3000DARK. If you change to a profile on devnet, it would be the reverse.
Ideally, the wallet should display both balances.

alexbarnsley: Yeh we'll get this merged. However, I don't think it's a good idea to load all ledgers for all networks. Some people have a lot of ledger wallets which could result in a very slow load of wallets when plugged in. It's unfortunately already slow because of having to do an API request for each wallet to determine if it's a cold wallet or not 🤔

j-a-m-l: Yes, that's my main concern, but the `Synchronizer` could be improved further and, after that, keeping the wallets in sync would be much more efficient, although the problem would be on the initial load. I'll try to have that changed next week.

PR #670, opened 2018-12-06
fix: validate fields when they get populated by aip13

### LINKED ISSUE #662: Transactions from AIP13 aren't filled in properly
**Describe the bug**
I have the following AIP13 compliant link: https://jsfiddle.net/0msew8n3/

When I click on it and tell my browser to open it on Ark Desktop Wallet, the Transfer fields are filled in, but the `Next` button is still disabled, even after entering my passphrase:

If I delete a character from the `Recipient` field and type it again, the `Next` button gets enabled:

**To Reproduce**
1. Click on the `Send Transaction` link on https://jsfiddle.net/0msew8n3/
2. Open it with the Ark Desktop Wallet

**Expected behavior**
The next button should be enabled after typing my passphrase/password.

ItsANameToo:
## Proposed changes

fixes #662, also see that issue for a link to test it

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
+++ b/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
@@ -141,12 +141,12 @@ export default {
 
   data: vm => ({
     form: {
-      amount: vm.schema.amount || '',
+      amount: '',
       fee: 0,
       passphrase: '',
       walletPassword: '',
-      recipientId: vm.schema.address || '',
-      vendorField: vm.schema.vendorField || ''
+      recipientId: '',
+      vendorField: ''
     },
     isSendAllActive: false,
     showEncryptLoader: false,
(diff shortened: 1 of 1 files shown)


vmunich:
Yay 😄

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -195,6 +195,17 @@ export default {
   },
 
   mounted () {
+    // Note: we set this here and not in the data property so validation is triggered properly when fields get pre-populated
+    if (this.schema) {
+      this.form = {
```
luciorubeens: Nested properties must be updated by the `$set` method:

```
this.$set(this.form, 'amount', this.schema.amount)
```

https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats

luciorubeens: But since you are defining the whole object it is probably already reactive 🤔

ItsANameToo: I can change it to this:

schermafbeelding 2018-12-06 om 15 54 07

Also gets rid of setting some properties again while they don't change 🤷‍♂️

PR #677, opened 2018-12-07
chore: add a new version  (`test:e2e:build`) of the script to run E2E tests

j-a-m-l:
## Proposed changes
Basically, this would be useful when executing E2E tests without changing the files of `src`.

## Types of changes

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,9 @@
     "pack:main": "cross-env NODE_ENV=production webpack --progress --colors --config .electron-vue/webpack.main.config.js",
     "pack:renderer": "cross-env NODE_ENV=production webpack --progress --colors --config .electron-vue/webpack.renderer.config.js",
     "postinstall": "electron-builder install-app-deps",
-    "test": "npm run test:unit && npm run test:e2e",
-    "test:e2e": "npm run pack && jest --config __tests__/e2e.jest.conf.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -38,8 +38,9 @@
     "pack:main": "cross-env NODE_ENV=production webpack --progress --colors --config .electron-vue/webpack.main.config.js",
     "pack:renderer": "cross-env NODE_ENV=production webpack --progress --colors --config .electron-vue/webpack.renderer.config.js",
     "postinstall": "electron-builder install-app-deps",
-    "test": "npm run test:unit && npm run test:e2e",
-    "test:e2e": "npm run pack && jest --config __tests__/e2e.jest.conf.js",
+    "test": "npm run test:unit && npm run test:e2e:build",
+    "test:e2e": "jest --config __tests__/e2e.jest.conf.js",
+    "test:e2e:build": "npm run pack && npm run test:e2e",
```
alexbarnsley: I'm not sure this command reads well. To me, if i was running `yarn test:e2e:build` i'd expect it only to build, not to test as well. What would be the purpose of running e2e tests multiple times in a row without building in between?

alexbarnsley: Okay, that makes sense as it's only related to the src folder changes, not tests. Could we reword the command though, perhaps `test:e2e:full` since it does both build **and** run?

PR #688, opened 2018-12-08
feat: disable 'sendall' when user focuses amount

### LINKED ISSUE #682: Send All
Couple issues with the `Send All` switch when sending:
- When click it again/turn it off, the total amount stays. It should reset the amount.
- When the switch is ON and I change the amount, after clicking on the next field, the amount is changed to total amount. When change the amount, it should turn off the switch.

kalgoop:
## Proposed changes

Send all option should be disabled when user puts the cursor on amount field.
I think, this fixed #682 . 
However, I don't agree with @zillionn that amount should be reset when sendAll is disabled.

## Types of changes

## Checklist

Also, I have changed name of 'onSendAll' function to 'setSendAll', since this function may be used from other places than send all button.


alexbarnsley:
@kalgoop 	I sort of agree with @zillion. Could you make it so that when disabling send all, it resets the amount if the amount === the value it would be if send all was enabled please. E.g. if i enable send all and then change the value manually, disabling shouldn't reset the value. if i enable send all, don't change the value and then disable send all, it should reset the value. Let me know if that doesn't make sense please.

kalgoop:
What do you mean by reset ?
Reset to 0 or reset to value that existed before enabling 'send all'

alexbarnsley:
I think reset the field to an empty value so it's more obvious that it's been reset. Both empty and 0 give the same error so don't need to worry about that

zillionn:
IMO reset should be the initial value. Example:
- I click on a `Pay with ARK` link to pay somebody `8.50 ARK`
- Wallet will populate all the fields for me
- Enabling `Send all` will set the amount to my total amount
- Disabling `Send all` should reset to the initial value - `8.50`, not to `0`

alexbarnsley:
I'm happy with that @zillionn, I didn't consider the URI's loading in data to the send form.

kalgoop:
ok. will push after some time

alexbarnsley:
I think keep it as the second point. whatever the initial value was, keep that if resetting. If the form was opened with no amount, reset to no amount. If the form started with an amount, reset to that amount. but only reset if the amount didn't change between enabling and disabling send all

kalgoop:
I'm unable to set field to blank value due to validators.

kalgoop:
I'm unable to set field to blank value (or 0) due to validators.

alexbarnsley:
Maybe try changing the validators so you can

Zillion:
You got the wrong Zillion, you want Zillionn I think. Just noticed I have a
number of emails from you guys.

Kind regards
Zillion with only one n

alexbarnsley:
sorry about that @Zillion with one n 😅

zillionn:
@Zillion Give me your GitHub password and we solve this :wink:

kalgoop:
@alexbarnsley setting amount to blank isn't possible due to number type in v-model.
I was trying since hours to fix this, but couldn't. 

Btw, why do you think current implementation isn't better?
I feel current implementation is better.
Why user would prefer blank field after enabling and disabling sendAll ?
I think there will be more users that would want sendAll amount there after enabling and disabling sendAll with initial blank amount.

kalgoop:
@faustbrian don't you think we should take feedback from users if they feel above is intuitive to them ?

kalgoop:
@alexbarnsley don't you think we should take feedback from users if they feel above is intuitive to them ?

zillionn:
@kalgoop Switching on/off should not break your form and should return to the previous value. Current implementation is broken, If you change the amount while the switch is on, it MUST go off, because currently it stays on and you can't change the amount.

Most people will probably never use the switch, but here is an example:
- You click on `ARK Pay` link, popup will open and most of the fields populated including some amount, then you click `Send All` just to see the total amount in your wallet, because there is no such info in that popup, or misclick it or just out of curiosity, and then what, how you get back the amount from the `ARK Pay` link?! You have to close the popup, search for that exact same link and click it again.

alexbarnsley:
@kalgoop 	is the PR as you envisage it should work? If so I will test and decide from there

kalgoop:
I had tested the code while pushing the pr. And it was as I wanted. 
I can't really test until day after tomorrow, if the conflicts caused some bug.

kalgoop:
@alexbarnsley tested it again today, working as intended.
" If the amount is initially empty, sendAll enabled and disabled -> amount remains the sendAll amount only
If the amount has some value initially, sendAllenabled and disabled -> amount resets to initial value "
Also, when you focus the amount textbox, sendall is disabled.

alexbarnsley:
when focusing on the amount field and enabling/disabling send all, you're getting an error

image

this doesn't happen for me on develop.

kalgoop:
@alexbarnsley fixed all errors.
With this PR - 
-When user toggles off the sendAll, amount is set to previous amount (even if previous amount had errors)
-When user focuses amount, sendAll is disabled, and the amount remains the same (as with sendAll)

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -39,13 +39,14 @@
         :wallet-network="walletNetwork"
         class="flex-1 mr-3"
         @blur="ensureAvailableAmount"
+        @focus="setSendAll(false,false)"
```
alexbarnsley: There should be a space after the comma `(false, false)`

alexbarnsley:
I don't think it should disable Send All on focus. It should be when the value is changed

kalgoop:
@alexbarnsley I dont agree. When user focuses, he'll come to know that sendAll has been disabled.
If sendAll is disabled on change, then even on focus sendAll is enabled and user may not even try changing the field.

Anyways, lint error and conflicts have been resolved.

alexbarnsley:
Well it seems you disagree with everything I say so I'm going to ask others to review this.

kalgoop:
@alexbarnsley well that's rude. I have given an explanation why I disagree.

alexbarnsley:
How is that rude? I gave my opinions to your changes and you've disagreed with them. I can't move forward with this PR with a stalemate between us, so I will get other people's feedback on your changes.

kalgoop:
@alexbarnsley "disagree with everything I say ".
Okay, leave that.

j-a-m-l:
So guys, how we move forward this PR?

The behaviour described by @ItsANameToo it's what I would do.

alexbarnsley:
@j-a-m-l we'll wait 14 days for a response and then close if there isn't one.

PR #695, opened 2018-12-10
Bug: Fix sorting of Wallet Transactions

kalgoop:
Sorting by different columns wasn't working correctly earlier.
This PR fixes the same.

## Proposed changes

- arguments have been corrected for the onSortChange
- __sortTransaction function added
- __sortTransaction called from fetchTransactions

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletTransactions/WalletTransactions.vue
+++ b/src/renderer/components/Wallet/WalletTransactions/WalletTransactions.vue
@@ -26,7 +26,7 @@
 </template>
 
 <script>
-import { at } from 'lodash'
+import { at, orderBy } from 'lodash'
 import mergeTableTransactions from '@/components/utils/merge-table-transactions'
 import TransactionTable from '@/components/Transaction/TransactionTable'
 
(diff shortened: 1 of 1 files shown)


kalgoop:
1. Oh. Didn't know about it.
2. Didn't realize earlier that transaction table too had the same problem. (when I made PR for delegate table)

alexbarnsley:
You can still modify a PR once you've created it if it hasn't been merged yet.

kalgoop:
I prefer keeping 2 fixes that are not on the same page apart in different PRs - 
- I don't think gitHub allows merging partial commits easily
- easier rollback, in case one of them fails
- easier to track

Anyways, different organisations have different rules. Above were my self-made rules from understanding of the teachings in my university course

dated:
> I don't think gitHub allows merging partial commits easily

You can cherry pick commits. If you need to change stuff, you can add the `--no-commit` flag.

alexbarnsley:
I would agree with you if they were both large commits, but they were both related to the same method on 2 pages. That means they should be in the same PR. There's no point doing 1 PR per page for the same change. The whole point of pull requests is to batch fixes/features/changes into the same request, because it's all related to each other and can be referenced or reverted (if needed) in one go.

- You can add to Pull Requests after they're made. You can cherry pick (`git cherry-pick <commit hash>`) from different branches in command line
- Unless it's a major change, it should just get additional fixes, not rolled back
- I'd say it's harder, 2 commits in 2 different PRs means requiring looking in 2 places. If it's discovered there is a bug with one of them, there could be a bug with the other. But other people may not know about the other because it was done in a completely different PR

alexbarnsley:
@dated 	:danstop:

PR #703, opened 2018-12-10
chore: remove yarn installation from circleci

dated:
## Proposed changes

Removes the yarn installation step from CircleCI, since yarn comes with the *-browsers images by default and the updated version isn't being used in the subsequent steps anyway.

## Types of changes

## Checklist


alexbarnsley:
tests are passing so must be good

PR #704, opened 2018-12-10
Add minimum fee notice

kalgoop:
## Proposed changes

User must be aware that setting fee to minimum may lead to transaction getting rejected after some time.

## Types of changes

## Checklist


alexbarnsley:
We was going to do this, however it's not just at minimum fee it can be rejected with so we decided to delay it to see what happened on mainnet. @boldninja it was us discussing it. Got any thoughts?

kalgoop:
At minimum the probability is maximum so ...
The ones using ark first time may be tempted to use minimum fees for small transactions. They should be notified regarding the option drawback.
Whereas the ones entering fees on their own, will probably have better knowledge of fees system of ark.

Another way is to display warning for all fees less than average.

alexbarnsley:
I like your last idea, display the warning for any fees below average and you've got yourself a deal 👌

kalgoop:
Everything now works as intended.
I was locally editing previous version of develop branch. I was messing up trying to get the additions on the updated branch.

PR #671, opened 2018-12-06
feat: set screenshot protection by environment variable

luciorubeens:
## Proposed changes

And also disable it in development mode by default.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

PR #705, opened 2018-12-10
fix: set screenshot protection upon app launch instead of sidebar

ItsANameToo:
## Proposed changes

#671 introduced a way to disable screenshot protection on launch, but it was set in the appsidemenu. This means that it got only triggered after the sidemenu got openend, which resulted in still not being able to take screenshots in dev mode when you hadn't touched the sidemenu yet. This PR moves it towards the `App.vue` file so it gets set immediately upon launch.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -146,6 +146,8 @@ export default {
         // We only set this if the env variable is 'false', since protection defaults to true
         // Since it's not a boolean, we can't do status !== false, since that would disable protection with every env var that's not 'true'
         this.$store.dispatch('session/setContentProtection', !(status === 'false'))
+      } else {
+        remote.getCurrentWindow().setContentProtection(true)
       }
     })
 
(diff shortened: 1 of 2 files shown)


src/renderer/App.vue
```diff
@@ -139,9 +139,16 @@ export default {
...
+      // Environments variables are strings
+      const status = process.env.ENABLE_SCREENSHOT_PROTECTION
+      if (status) {
+        // We only set this if the env variable is 'false', since protection defaults to true
+        // Since it's not a boolean, we can't do status !== false, since that would disable protection with every env var that's not 'true'
+        this.$store.dispatch('session/setContentProtection', !(status === 'false'))
+      }
     })
 
-    remote.getCurrentWindow().setContentProtection(this.hasProtection)
```
alexbarnsley: if i don't set `ENABLE_SCREENSHOT_PROTECTION` at all, does this mean it will be off by default now? 🤔

ItsANameToo: 🤔

PR #714, opened 2018-12-11
fix: check correctly if the balance is enough for creating the second signature

j-a-m-l:
## Proposed changes
The message about "not enough funds" was displayed always.

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/components/Input/InputFee.spec.js
+++ b/__tests__/unit/components/Input/InputFee.spec.js
@@ -267,7 +267,7 @@ describe('InputFee', () => {
       })
 
       describe('when the balance is bigger than the fee', () => {
-        fit('should return an empty message', () => {
+        it('should return an empty message', () => {
           // NOTE: Balance is in arktoshi, while fee i in units
           wrapper.vm.wallet_fromRoute.balance = 20e8
           wrapper.vm.fee = 1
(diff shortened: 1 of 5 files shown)


ItsANameToo:
Not "always" but you are right that it didn't correctly calculate it as `2000 ark` would trigger the error if fees were larger than `2`. Good catch

__tests__/unit/components/Input/InputFee.spec.js
```diff
@@ -239,4 +245,39 @@ describe('InputFee', () => {
...
+          expect(wrapper.vm.insufficientFundsError).toEqual('TRANSACTION_FORM.ERROR.NOT_ENOUGH_BALANCE')
+
+          wrapper.vm.wallet_fromRoute.balance = '10000'
+          wrapper.vm.fee = '20e8'
+          expect(wrapper.vm.insufficientFundsError).toEqual('TRANSACTION_FORM.ERROR.NOT_ENOUGH_BALANCE')
+        })
+      })
+
+      describe('when the balance is bigger than the fee', () => {
+        fit('should return an empty message', () => {
```
dated:
```suggestion
        it('should return an empty message', () => {
```
Leftover? :)

j-a-m-l: Thank you!

PR #729, opened 2018-12-12
refactor: use semver for release check

dated:
## Proposed changes

Uses the `semver` package to determine if there is a new version available. Right now a false positive is displayed, since 2.0.0 (latest release) and 2.1.0 (package.json) are not equal, but the former is not higher than the latter.

## Types of changes

## Checklist


dated:
:thinking:

PR #737, opened 2018-12-13
Fix disappearing div backgrounds

kalgoop:
screenshot from 2018-12-13 22-54-08

As you can see, the background of the div holding contacts disappears on decreasing the width of window.

## Types of changes

## Checklist


ItsANameToo:
That's according to design

PR #739, opened 2018-12-13
refactor: chart buttons hover background color

dated:
## Proposed changes

There is currently no visual feedback when hovering over the chart header buttons.

nohover -> hover

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/MarketChart/MarketChartHeader.vue
+++ b/src/renderer/components/MarketChart/MarketChartHeader.vue
@@ -49,3 +49,13 @@ export default {
   }
 }
 </script>
+
+<style scoped>
+.MarketChartHeader__button {
+  transition: all 0.3s;
+}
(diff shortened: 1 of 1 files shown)


j-a-m-l:
In your example, `Day` should change too

dated:
Since `Day` is the currently active button that is not necessary imho.

j-a-m-l:
Since there is only 1 active button, maybe it's better changing the opacity on hover on active and inactive buttons, instead of using the same style.

dated:
Doesn't look so good 😕

transitions

The style of the PR is in alignment with how the buttons are handled on the explorer.

j-a-m-l:
If you don't differentiate the active button and the hover, that's confusing.

About looking good, I'd say that it's almost the same. Push those changes, and I'll review and approve them.

PR #741, opened 2018-12-13
bug: fix multiple profile creation on multiple clicks

kalgoop:
On the profile creation page, I have noticed that when in the end you have to press 'Done'. If you press that button multiple times quickly, you'll get multiple profiles created with the same name.
This problem is more visible on slow networks as it gives more time for multiple clicks.

A user may press Done multiple times if the profile doesn't load instantly.
## Proposed changes

in the create function, profile name is revalidated if its unique

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -135,6 +135,7 @@ export default {
       const currentProfileId = this.$store.getters['session/profileId']
       await this.$store.dispatch('session/reset')
       await this.$store.dispatch('session/setProfileId', currentProfileId)
+      await this.$store.dispatch('ledger/reset')
 
       this.isReady = true
 
--- a/src/renderer/components/App/AppSidemenu/AppSidemenuSettings.vue
+++ b/src/renderer/components/App/AppSidemenu/AppSidemenuSettings.vue
@@ -64,6 +64,24 @@
       </MenuOptionsItem>
 
       <MenuOptionsItem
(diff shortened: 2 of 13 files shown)


JeremiGendron:
@kalgoop I found the bug, I'll submit a PR to your branch now.

kalgoop:
@JeremiGendron nice catch ! I think this would fix the issue. I'll test this tomorrow morning and rollback my commit if your commit is enough to ensure non duplicate profile names

dated:
Why not disable the button once clicked? `ReloadButton` for reference.

JeremiGendron:
```js
        <button
          v-if="isNextVisible || isNextEnabled"
          :disabled="!isNextEnabled || isLastItemClicked"
          class="MenuStepItem__footer__next-button blue-button"
          @click="emitNext(isLastItem)"
        >
```
@dated already in place

JeremiGendron:
@kalgoop could work better (event listeners should be triggered after setting isLastItemClicked, not before):
```js
    emitNext (isLastItem) {
      if (isLastItem) {
        this.isLastItemClicked = true
        this.$emit('next')
      }
      if (!this.lastItemClicked) {
        this.$emit('next')
      }
    }
```

kalgoop:
Everything seems to be fine. I tried really hard to click the Done button multiple times but failed. :smile:

PR #745, opened 2018-12-14
URI waits for activeWallet

### LINKED ISSUE #738: URI scheme
When I click on a link starting with `ark:`, the wallet opens up and shows an error message:
`Select a wallet and open the URI again`

**Describe the solution you'd like**
Is it that hard to set that URI to some variable and show a message like `Please choose a wallet.`. Then when I choose a wallet just check if there is such variable and open and populate the `Send` popup?

kalgoop:
Resolves #738 

## Proposed changes

setInterval with interval of 1000 (may be changed) ms and wait for active wallet, and open the uri after that.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

I have checked on my level. But I was unable to associate 'ark://' on my laptop.


ItsANameToo:
I believe a better solution would be to have a modal with more information than just using an error notification to let someone know they should select a wallet. So you would be greeted with a modal that prompts you to select a wallet instead of seeing an error, that's actually not an error but a hint, that you need to select a wallet for the URI 🤔

zillionn:
I believe this is the right way: #751

alexbarnsley:
I disagree @zillionn. To me it would be a strange user experience to navigate the wallet as normal and then have the send form popup automatically once i clicked on a wallet. What if i clicked the wrong wallet? What if i wanted to cancel before the popup showed up? I think a simple popup with 2 dropdown boxes for Profile and Wallet doesn't seem like a major problem to me, and it then clearly shows the users it's purpose.

zillionn:
@alexbarnsley I agree with you that a popup is more user friendly. If you have time to create it, do it!

One more thing.
https://github.com/ArkEcosystem/desktop-wallet/blob/7dfef473526e93f5783a0aeb15939c52f7222e62/src/renderer/App.vue#L206
Not sure how to fix this, but this works only when the wallet is already open. If the wallet is closed, the `ark:` link opens the wallet but no alert messages.

alexbarnsley:
I will see if i can replicate that @zillionn. What OS are you running?

We will be doing the dropdown update as we would like to release again soon with a couple of fixes anyway that were done after 2.1.0. As a result, I will close this PR since I'll be working on it now, thanks @kalgoop

zillionn:
@alexbarnsley macOS Mojave 10.14.2

PR #746, opened 2018-12-14
refactor: wallet transactions & delegate table sorting

dated:
## Proposed changes

Imho sorting should not trigger a potential change of the dataset, but only manipulate the order of the data already present. The current implementation fetches the delegates in the `onSortChange` method.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletDelegates/WalletDelegates.vue
+++ b/src/renderer/components/Wallet/WalletDelegates/WalletDelegates.vue
@@ -218,8 +218,8 @@ export default {
     },
 
     onSortChange (sortOptions) {
-      const columnName = sortOptions[0].field
       const sortType = sortOptions[0].type
+      const columnName = sortOptions[0].field
       this.__updateParams({
         sort: {
           type: sortType,
@@ -227,7 +227,7 @@ export default {
         },
         page: 1
       })
-      this.fetchDelegates()
+      this.delegates = this.__sortDelegates()
     },
 
     reset () {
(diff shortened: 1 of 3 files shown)


dated:
Should probably extended to `WalletTransactions.vue` as well.

j-a-m-l:
I'm not sure about this. IIRC, we passed from sorting the rows locally on the initial implementations to doing it remotely.

About the pointer, please, include it on other PR.

dated:
added to #735

dated:
all it does is refetching the delegates before sorting them, which delays the action unnecessarily.

```
const { delegates, totalCount } = await this.$client.fetchDelegates({
  page: this.queryParams.page,
  limit: this.queryParams.limit
})
this.delegates = this.__sortDelegates(delegates)
```

alexbarnsley:
There a few issues with this:

1 - Going to page 2 and sorting resets the page to 1, but shows page 2's sorted data
2 - If i sort date ascending on page one it sorts locally. When i then go to page 2, it shows page 2 of my very early transactions, because sorting is also done on the API
3 - Assuming you removed the API sorting, your change means I may never be able to see my very first transactions if I have a lot

I agree with @j-a-m-l that this should be done remotely, as I would expect to see my first transaction on page 1 if i sorted locally.

dated:
Yes, i hadn't looked into the transaction fetching properly. This PR was originally intended only for the delegates table, the changes to the transactions were added later and are obviously wrong.

The delegate table has no remote sorting however, should that be added? If not I'd still say that the re-fetching of the delegates is superfluos.

PR #778, opened 2018-12-17
feat: add vote details to WalletAddress.vue

dated:
## Proposed changes

Adds the delegate name to `WalletAddress.vue` if the transaction shown is a vote/unvote transaction.

## Types of changes

## Checklist


src/renderer/store/modules/delegate.js
```diff
@@ -25,6 +25,20 @@ export default {
...
+    },
+
+    byPublicKey: (state, _, __, rootGetters) => publicKey => {
+      const network = rootGetters['session/network']
+
+      if (!network || !state.delegates[network.id]) {
+        return false
+      }
+
+      const delegates = Object.values(state.delegates[network.id]).filter(delegate => {
```
alexbarnsley: Could possibly just do `find` here (I returned false on failure in case you specifically want to return false on failure):

```
return Object.values(state.delegates[network.id]).find(delegate => {
  return delegate.publicKey === publicKey
}) || false
```

src/renderer/store/modules/delegate.js
```diff
@@ -25,6 +25,20 @@ export default {
       }
 
       return state.delegates[network.id][address]
+    },
+
+    byPublicKey: (state, _, __, rootGetters) => publicKey => {
+      const network = rootGetters['session/network']
+
+      if (!network || !state.delegates[network.id]) {
```
alexbarnsley: Could probably also check to make sure publicKey is set. Unlikely to happen but saves the loop in case it ever is a scenario in the future elsewhere

PR #780, opened 2018-12-17
fix: showPublicKey bug with address import

JeremiGendron:
## Proposed changes

#771 doesn't set the activeWallet, which means if you switch to a second wallet the fix is voided. This is fixed in this PR and refactored a little bit, so as to store the id of the wallet only and not the whole object.

The other change I've implemented here is to watch for changes in the currentWallet's publicKey in the `WalletSidebar` in order to check whether the wallet has one. If the wallet has no publicKey, it can cause a bug where the heading doesn't get reset.

1. The wallet heading reset is still misbehaving
 - Before
before_heading_noset

 - After
after_heading_set

2. The publicKey bug mentioned above (imported wallet with address only)
 - Before
before_publickey_noreset

 - After
after_publickey_reset

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletHeading/WalletHeading.vue
+++ b/src/renderer/components/Wallet/WalletHeading/WalletHeading.vue
@@ -39,7 +39,7 @@ export default {
 
   watch: {
     currentWallet () {
-      if (this.activeWallet !== this.currentWallet.id) {
+      if (this.activeWallet && this.activeWallet.id !== this.currentWallet.id) {
         this.resetHeading()
       }
     }
--- a/src/renderer/components/Wallet/WalletHeading/WalletHeading.vue
+++ b/src/renderer/components/Wallet/WalletHeading/WalletHeading.vue
@@ -22,7 +22,7 @@ export default {
   },
 
   data: () => ({
-    activeWalletId: null
+    activeWallet: null
   }),
 
   computed: {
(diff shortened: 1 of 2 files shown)


ItsANameToo:
Activewallet is set in `resetHeading()` tho: https://github.com/ArkEcosystem/desktop-wallet/blob/8cbdcb0268f7d8385a4ebbbcad4999de8c35e408/src/renderer/components/Wallet/WalletHeading/WalletHeading.vue#L55

ItsANameToo:
Also: I'm unable to reproduce what you show in your gif. With latest develop I can't get the heading to reset itself at random 🤷‍♂️

JeremiGendron:
@ItsANameToo I admittedly didn't review the changes that well, let me see if I can reproduce in develop.

src/renderer/components/Wallet/WalletHeading/WalletHeading.vue
```diff
@@ -39,7 +39,7 @@ export default {
 
   watch: {
     currentWallet () {
-      if (this.activeWallet && this.activeWallet.id !== this.currentWallet.id) {
+      if (this.activeWallet !== this.currentWallet.id) {
```
ItsANameToo: You should add `&& this.activeWallet.id` again, as it currently doesn't check against an id anymore

JeremiGendron: cool, on it!

PR #822, opened 2018-12-21
fix: how the tooltip is hidden when scrolling

j-a-m-l:
## Proposed changes
Solves https://github.com/ArkEcosystem/desktop-wallet/issues/767

@dated @JeremiGendron would you mind to try this branch?

## Types of changes
- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist
- [x] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation
- [x] Lint and unit tests pass locally with my changes

Issue #767, opened 2018-12-15
Entire window scrollable when popover is scrolled out of view


JeremiGendron:
**Platform (please complete the following information):**
 - OS: Lubuntu 18.10 (QT)
 - Version: Develop

**Describe the bug**
When scrolling the popover out of view, the entire frame can be scrolled and leave a scroll-sized white area in the direction where the popover went.

**How to reproduce the bug**
Steps to reproduce the behaviour:
1. Go to Wallet Transactions List
2. Hover on any field with a popover tooltip (From, To, Amount) of the last transaction (works best with 20 + per page)
3. Scroll up multiple times
4. See big scrollbar on right and bottom of screen and white area when scrolling all the way to the bottom

**Expected behaviour**
No scrollbars appear, no white is visible.

**Screenshots**
![hover-tx-sent](https://user-images.githubusercontent.com/23587429/50048133-57d03980-0092-11e9-848c-5a8e3a6f60ad.png)
![scroll-after-hover](https://user-images.githubusercontent.com/23587429/50048134-5acb2a00-0092-11e9-90e5-71774c1259f8.png)

dated:
![scroll](https://user-images.githubusercontent.com/6547002/50048184-b8c53500-00c5-11e9-9998-be06fbfa8367.gif)

The tooltip should probably be dismissed upon scrolling.

JeremiGendron:
@dated good catch

JeremiGendron:
v-tooltip appends to `<body>` by default, which can be changed with `container: '__CSS_SELECTOR__'` in v-tooltip binding data. Also, `boundariesElement` might be of use.

dated:
This can probably be closed.

ArkEcosystemBot:
@JeremiGendron This issue has been closed. If you wish to re-open it please provide additional information.

PR #824, opened 2018-12-21
fix: set tooltip container on transaction table

dated:
## Proposed changes

Partially solves #767 and replaces #822.

This sets the tooltip container to `.TransactionTable` and adds a property to `WalletAddress`, which allows to set a custom tooltip container.

tooltip

@j-a-m-l & @JeremiGendron can you confirm this works for you?

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Transaction/TransactionShow.vue
+++ b/src/renderer/components/Transaction/TransactionShow.vue
@@ -134,7 +134,7 @@
         <span
           v-show="!transaction.isExpired"
           v-tooltip="{
-            content: $t('TRANSACTION.CONFIRMATION_COUNT', [transaction.confirmations]),
+            content: $t('TRANSACTION.CONFIRMATION_COUNT', { confirmations: transaction.confirmations }),
             trigger: 'hover'
           }"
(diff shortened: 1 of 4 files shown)


j-a-m-l:
This doesn't solve the problem, @dated. Thanks anyway for trying it.

dated:
Which problem our we trying to solve? I am unable to replicate the issue described by Jeremi, but this solves the problem of the tooltip being shown above the app footer.

Imho adding an event listener to detect the scroll event and dismiss the tooltip is overkill.

dated:
To clarify, this PR fixes this behaviour:

zillionn:
Can't you listen for the scroll only when you show the tooltip and `onScroll` hide the tooltip and remove the listener?

dated:
If anything such behaviour should be added to the Akryum/v-tooltip package.

dated:
https://github.com/Akryum/v-tooltip/issues/145

j-a-m-l:
@dated I missunderstood your intention: I though you were trying to fix that v-tooltip issue.

PR #826, opened 2018-12-21
refactor: maxlength of vendorField

kalgoop:
## Proposed changes

add maxlength attribute to vendorfield to prevent user from entering more than 64 characters instead of displaying an error

## Types of changes

## Checklist


alexbarnsley:
This doesn't work. You've removed methods but not where they're called

image

kalgoop:
App was working as intended. However, I didn't check the console.
Will add commit to fix this.

kalgoop:
@alexbarnsley removed those calls to non-existant functions

alexbarnsley:
There's still errors in the console which aren't present on the develop branch:

image

You should really check the logs for errors

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -427,9 +416,7 @@ export default {
           return this.walletNetwork.apiVersion === 1 // Return true if it's v1, since it has a static fee
         }
       },
-      vendorField: {
-        maxLength: maxLength(64)
-      },
+      vendorFielf: {},
```
j-a-m-l: `vendorField` or, since your are not using it anymore, remove it.

kalgoop: Actually, if I don't include validation for vendorField, I see a lot of errors due to that in console

j-a-m-l: Yes, that's is because you are using the Vuelidate field (`$v.form.vendorField`). Since it's not performing any validation, you could have replaced it with `form.vendorField` and removed the `vendorField` validation.

j-a-m-l:
@kalgoop displaying some feedback when the field has reached the 64 limit would be very useful; otherwise users might not realized that some characters are not included when pasting text or they could think that something is wrong.

j-a-m-l:
@kalgoop, are you going to finish this PR? We would wait some days more and, if no action is taken, we would close it.

kalgoop:
@j-a-m-l will do today. Sorry for the delayed response.

kalgoop:
@j-a-m-l no errors now and everything works as intended.

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -244,15 +244,11 @@ export default {
     vendorFieldLabel () {
       return `${this.$t('TRANSACTION.VENDOR_FIELD')} - ${this.$t('VALIDATION.MAX_LENGTH', [64])}`
     },
-    vendorFieldError () {
-      if (this.vendorFieldIsInvalid) {
-        return this.$t('VALIDATION.TOO_LONG', [this.$refs.vendorField.label])
+    vendorFieldHelperText () {
+      if (this.form.vendorField.length === 64) {
+        return this.$t('VENDORFIELD.LIMIT_REACHED', [64])
```
j-a-m-l: Better use `VALIDATION.VENDOR_FIELD.LIMIT_REACHED`

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -244,15 +244,11 @@ export default {
     vendorFieldLabel () {
       return `${this.$t('TRANSACTION.VENDOR_FIELD')} - ${this.$t('VALIDATION.MAX_LENGTH', [64])}`
     },
-    vendorFieldError () {
-      if (this.vendorFieldIsInvalid) {
-        return this.$t('VALIDATION.TOO_LONG', [this.$refs.vendorField.label])
+    vendorFieldHelperText () {
+      if (this.form.vendorField.length === 64) {
```
j-a-m-l: Instead of hardcoding `64` here:

1. add that value to each network configuration of `config/networks/*.json`
```json
"vendorField": {
  "maxLength": 64
}
```

2. add a new key on `config/index.js`:

```js
exports.VENDOR_FIELD = {
  defaultMaxLength: 64
}
```

kalgoop:
@j-a-m-l I am unable to figure out why session_network.vendorField is undefined.

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -242,17 +242,13 @@ export default {
       }
     },
     vendorFieldLabel () {
-      return `${this.$t('TRANSACTION.VENDOR_FIELD')} - ${this.$t('VALIDATION.MAX_LENGTH', [64])}`
+      return `${this.$t('TRANSACTION.VENDOR_FIELD')} - ${this.$t('VALIDATION.MAX_LENGTH', [this.session_network.vendorField.maxLength])}`
```
j-a-m-l: You could replace this `this.session_network.vendorField.maxLength` with a computed property `vendorFieldMaxLength`, here and in other places.

src/renderer/i18n/locales/en-US.js
```diff
@@ -155,6 +155,9 @@ export default {
     PUBLIC_KEY: {
       INVALID_LENGTH: 'The public key must be 66 characters long'
     },
+    VENDORFIELD: {
```
j-a-m-l: `VENDOR_FIELD`.

src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -242,17 +242,13 @@ export default {
...
     },
-    vendorFieldError () {
-      if (this.vendorFieldIsInvalid) {
-        return this.$t('VALIDATION.TOO_LONG', [this.$refs.vendorField.label])
+    vendorFieldHelperText () {
+      if (this.form.vendorField.length === this.session_network.vendorField.maxLength) {
+        return this.$t('VENDORFIELD.VENDOR_FIELD.LIMIT_REACHED', [this.session_network.vendorField.maxLength])
```
j-a-m-l: Fix the i18n key.

kalgoop:
@j-a-m-l this.walletNetwork.vendorField is still undefined.
I wonder why only this.walletNetwork.vendorField.maxLength gives error while this.walletNetwork.constants.activeDelegates doesn't.

j-a-m-l:
Another reason might be that the network you are using doesn't have that field (because it was created before starting this PR). Try deleting all the data of the application.

kalpitk:
@j-a-m-l Just deleted all app data. Everything works perfect.

kalpitk:
Works fine after deletion of app data

kalgoop:
No errors in console after app data deletion

j-a-m-l:
Since these changes should work on previously installed apps, we need to be sure that nothing would break, so:

1. Go back to the `develop` branch and create a profile
2. Checkout this branch and set the default value, on network model, of `vendorField` to `VENDOR_FIELD.defaultMaxLength`.
3. Check if that profile works.

If not, we would need to create a DB migration that, in case that `vendorFileld` doesn't exist, creates the value. You could see an example here: https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/renderer/store/migrations/2.2.0%20-%20initial%20migration.js

kalgoop:
@j-a-m-l Now I am using VENDOR_FIELD.defaultMaxLength if this.walletNetwork.vendorField.maxLength gives undefined.

src/renderer/i18n/locales/en-US.js
```diff
@@ -155,6 +155,9 @@ export default {
     PUBLIC_KEY: {
       INVALID_LENGTH: 'The public key must be 66 characters long'
     },
+    VENDOR_FIELD: {
+      LIMIT_REACHED: 'You can enter maximum \'{0}\' characters only'
```
j-a-m-l: @kalgoop remove the \\`

kalgoop:
@j-a-m-l done

j-a-m-l:
Thanks @kalgoop.

PR #829, opened 2018-12-21
fix: capture any error when sending transactions and warn about broadcasting

j-a-m-l:
## Proposed changes
This PR fixes 1 problem introduced on https://github.com/ArkEcosystem/desktop-wallet/pull/820, which ignores that the API response could include other errors.

Apart from that, it displays a warning when the transaction is only broadcasted.

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/components/Transaction/TransactionModal.spec.js
+++ b/__tests__/unit/components/Transaction/TransactionModal.spec.js
@@ -58,8 +58,39 @@ describe('TransactionModal', () => {
         response.status = 200
       })
 
+      describe('when the response does not include errors', () => {
+        beforeEach(() => {
+          response.data.errors = null
+          response.data.data = {
+            invalid: [],
+            accept: []
+          }
+        })
+
+        it('should return `true`', () => {
+          expect(wrapper.vm.isSuccessfulResponse(response)).toBeTrue()
+        })
+      })
+
+      describe('when the response includes errors', () => {
+        beforeEach(() => {
+          response.data.errors = {
+            tx1: [{ type: 'ERR_LOW_FEE' }]
+          }
+          response.data.data = {
+            invalid: [],
+            accept: []
+          }
+        })
+
+        it('should return `false`', () => {
+          expect(wrapper.vm.isSuccessfulResponse(response)).toBeFalse()
+        })
+      })
+
       describe('when the response does not include invalid transactions', () => {
         beforeEach(() => {
+          response.data.errors = null
           response.data.data.invalid = []
         })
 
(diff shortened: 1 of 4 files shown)


j-a-m-l:
@zillionn I'm going to add more changes and tests to this component, how would you rephrase the broadcast warning?

zillionn:
Maybe something like this:
`Transaction was successfully sent but it may not be accepted/forged because the fee (Ѧ0.00000300) is too low.`

I'm also experiencing something else. I'm using custom peer `95.179.150.18:4003` and I'm pulling the data from it, but when I send a tx, it is sent to a different peer `167.114.29.34:4003`:
screenshot 2018-12-21 at 18 00 22
screenshot 2018-12-21 at 18 00 33
screenshot 2018-12-21 at 18 00 43

zillionn:
I guess the problem is with `alternativeWallet`, it's always set in `onBuilt`:
https://github.com/ArkEcosystem/desktop-wallet/blob/c013b91f4e0c2b9907309ad4bb2937404b5c69bb/src/renderer/components/Transaction/TransactionModal.vue#L103-L107

dated:
The wallet looks for the best peer and sends to that.

zillionn:
@j-a-m-l Also your code is wrong and breaks what I did:
https://github.com/ArkEcosystem/desktop-wallet/blob/c013b91f4e0c2b9907309ad4bb2937404b5c69bb/src/renderer/components/Transaction/TransactionModal.vue#L192

If my tx is broadcasted but not accepted:
> 10|ark-core-relay  | [2018-12-21 16:30:23][DEBUG]                      : Transaction bb2f15321ddfa051decd081e98984a070100e35160185e5b140d63aae543e017 eligible for broadcast - fee of 0.000003 DѦ is greater than minimum fee (0.00000253 DѦ)
10|ark-core-relay  | [2018-12-21 16:30:23][DEBUG]                      : Transaction bb2f15321ddfa051decd081e98984a070100e35160185e5b140d63aae543e017 not eligible to enter pool - fee of 0.000003 DѦ is smaller than minimum fee (0.00759 DѦ)
10|ark-core-relay  | [2018-12-21 16:30:23][INFO]                       : Received 1 transaction (accept: 0 broadcast: 1 excess: 0 invalid: 1).

the above code should return `true` because the tx is broadcasted, but it returns `false`, because `data.invalid.length != 0` and `errors != null`.
screenshot 2018-12-21 at 18 34 57

I believe my check is the better approach:
https://github.com/zillionn/desktop-wallet/blob/94733b6d1a3d8116ffd142285629c437f925b3ef/src/renderer/components/Transaction/TransactionModal.vue#L170-L172

PR #833, opened 2018-12-21
fix: range of ports

kalgoop:
## Proposed changes

- fix range of ports (65535 was excluded)
- use regex to test if numeric

## Types of changes

## Checklist


kalgoop:
@dated you are right. I think numeric function is of not much use and can be dropped off. regex in code appear more readable (not always). And it'll avoid import of a module.
Yeah it doesn't change things much.

faustbrian:
Use the numeric validation rather then a regex and avoid unnecessary changes like that in the future as they are personal preference.

kalgoop:
@faustbrian reverted. Will keep in mind

PR #832, opened 2018-12-21
perf: findLatestTransaction

kalgoop:
## Proposed changes

instead of sorting and then picking the latest transaction, loop over all transactions to find the latest transaction.
Reduces complexity from O(nlogn) to O(n)

## Types of changes

- [x] Refactoring (improve a current implementation without adding a new feature or fixing a bug)

## Checklist

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation
- [x] Lint and unit tests pass locally with my changes

PR #878, opened 2018-12-28
perf: sortBy / orderBy

dated:
## Proposed changes

- replace sortBy with orderBy in findLatestTransaction:
  replacement for https://github.com/ArkEcosystem/desktop-wallet/pull/832, uses `orderBy` so complexity should stay at O(logn) instead

- replace sortBy with orderBy in merge-table-transactions.js:
  uses `orderBy` instead of `sortBy` so reversing the list is unnecessary

- replace orderBy with sortBy ContactAll.vue:
  since contacts with the same name are not allowed, ordering by address as well is unnecessary

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/utils/merge-table-transactions.js
+++ b/src/renderer/components/utils/merge-table-transactions.js
@@ -1,4 +1,4 @@
-import { sortBy, uniqBy } from 'lodash'
+import { orderBy, uniqBy } from 'lodash'
 
 /**
  * This utility is used to merge the transactions that have been fetched using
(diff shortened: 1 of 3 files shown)


kalgoop:
@dated how using order by keeps O(logn). I think it will take same O(nlogn) as earlier implementation. Sorting/ordering takes O(nlogn) and picking first element takes O(1). So overall it is O(nlogn)

Instead if you traverse all the elements of transactions to find max timestamp it takes O(n) only. (As in #832 )

dated:
You are correct, that was a typo on my end.

kalgoop:
One of your commit actually reverts the #840 .

PR #881, opened 2018-12-28
[WIP] [DO NOT MERGE] feat: shown placeholder message if wallet is a known wallet

ItsANameToo:
## Proposed changes

This will make the wallet show an icon (waiting for the design from Oleg) when a wallet is a `known wallet`. The way it is set up makes sure that a known wallet will also get the icon if it has been imported as a contact with a different name, so if you have for example `binance` as a contact, it will still show the icon but with the contact name you have given it.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletAddress.vue
+++ b/src/renderer/components/Wallet/WalletAddress.vue
@@ -53,6 +53,12 @@
     <span v-else-if="type === 8">
       {{ $t("TRANSACTION.TYPE.DELEGATE_RESIGNATION") }}
     </span>
+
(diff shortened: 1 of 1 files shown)


dated:
If feasable we could even show a customized icon per address, e.g. the binance logo for binance. With a generic fallback icon if none is defined. Thoughts?

faustbrian:
I would keep it generic, it just adds overhead and no value if you use wallet specific icons.

ItsANameToo:
@faustbrian it's finished now

PR #910, opened 2018-12-31
fix: make time consistent in wallet

### LINKED ISSUE #908: Verify Timedate Status & Add 24h/12h Preference
Verify timedate status. Is the desktop showing timedates in local system time, or always GMT? If always GMT make local system time.

Also let's add a preference in the gear menu for "Use 24h time format" with switch, so if 12h time format is desired, it can be used. I noticed that the explorer shows 12h time format, so desktop wallet should be consistent

ItsANameToo:
## Proposed changes

Makes time in wallet consistent with the one used on the explorer (and consistent in the wallet itself by using `moment` everywhere instead of vue's `$d` function). It also makes use of the "Multiple Locale Support" so it should show 12AM/PM for locales where that is the standard. 

Resolves #908 

Edit: added a dropdown in the profile where you can choose between default, 12h and 24h times. I’ve also removed the `capitalize` function from lodash as that would capitalize the first letter and make all others lowercase, which results in issues with store functions like `setBip39Language` as it would try to use `setBip39language`

## Types of changes

## Checklist


src/renderer/components/Transaction/TransactionTable.vue
```diff
@@ -246,9 +245,7 @@ export default {
 
   methods: {
     formatDate (value) {
-      moment.locale(window.navigator.userLanguage || window.navigator.language)
-
-      return moment(value).format('L HH:mm:ss')
+      return this.formatter_date(value)
```
ItsANameToo: @alexbarnsley in case this should be in the 24h format like you said in the issue, we can pass it `L HH:mm:ss`

alexbarnsley: I think its worth adding an option in the Profile Edit/Create screen to choose. Default it to locale I guess and people can change it that way

PR #912, opened 2018-12-31
fix: update alternative currency when no input

dated:
## Proposed changes

The alternative currency amount is not being set to 0 when there is no input.

@ItsANameToo this seems to work as originally intended, i am unable to reproduce the same bugs as in #882. Can you confirm?

## Types of changes

## Checklist


ItsANameToo:
Seems to work 👍

PR #915, opened 2018-12-31
fix: do not show tx notifications for contacts

ItsANameToo:
## Proposed changes

Transaction notifications would also show for contacts, which it shouldn't do
 
## Types of changes

## Checklist

### DIFF
--- a/src/renderer/services/synchronizer/wallets.js
+++ b/src/renderer/services/synchronizer/wallets.js
@@ -184,55 +184,57 @@ class Action {
   }
 
   displayNewTransaction (transaction, wallet) {
-    let message = {}
-    switch (transaction.type) {
-      case config.TRANSACTION_TYPES.SECOND_SIGNATURE: {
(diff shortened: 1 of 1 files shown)


src/renderer/services/synchronizer/wallets.js
```diff
@@ -184,55 +184,57 @@ class Action {
...
 
   displayNewTransaction (transaction, wallet) {
-    let message = {}
-    switch (transaction.type) {
-      case config.TRANSACTION_TYPES.SECOND_SIGNATURE: {
-        message = {
-          translation: 'SYNCHRONIZER.NEW_SECOND_SIGNATURE',
-          options: {
-            address: truncateMiddle(wallet.address)
+    if (!wallet.isContact) {
```
alexbarnsley: I think instead of all the indenting, just do the below at the top:

```
if (wallet.isContact) {
  return
}
```

Issue #902, opened 2018-12-30
Import Wallet Page Adjustments


doubled1c3:
1. Remove the "use the address only" and "use the passphrase only" elements. I can tell there wants to be a 'both' aspect, to verify that the address and passphrase match. But that aspect is not necessary if we modify the passphrase autotext. I'll explain further soon. Also a switch is not the best choice as graphical element. Should be a button that stays lit.

See below graphic I mocked up for the change, including some addtl text elements:

![image](https://user-images.githubusercontent.com/32085788/50551291-2c715100-0c33-11e9-9360-5f6cc4393119.png)

2. Change "The 'Passphrase' is not a BIP39 Passphrase" to "This is not a valid BIP39 passphrase"

3. Upon successful entry of a BIP39 passphrase, the orange autotext becomes green autotext, showing the corresponding address. This will allow a visual check and handle the 'both' feature described in item 1. 

See below mockup:

![image](https://user-images.githubusercontent.com/32085788/50551417-cede0400-0c34-11e9-8815-bea8ea0a8e13.png)

4. Add explainer text to Encryption section:

The ARK Desktop Wallet has an additional feature that lets you set an encrypted password as well, for easier management. If your computer dies, you cannot use the encrypted password to access your funds from a different machine. You will need the passphrase. Warning: Storing your passphrase can bring risks.

See mockup below:

![image](https://user-images.githubusercontent.com/32085788/50551538-ef0ec280-0c36-11e9-9add-9c86fb6db3db.png)

**Discussion:** Regarding 'storing your passphrase can bring risks,' I would like more clarity on what are the risks. I'm going to take a stab in the dark at this. When an an encrypted password is set for a wallet, it creates an encrypted file on the user PC. If a hacker were to obtain the encrypted file, it could theoretically be bruteforced. Thus, if the encrypted password were "Fluffy1987!" it would be more at risk than if the encrypted password were "ThisPhraseIsAmazingAndWillTake1,234,567,890YearsToFigureOut!" Is my understanding of this correct?

5. Change ghost text "Wallet password (optional) to "Set encrypted password (optional)"

6. Confirmation intro text. Change "Wallet configuration Optionally, choose a name for you wallet and you are done" to "Wallet name Choose an optional wallet name to complete the wallet import process."

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

ArkEcosystemBot:
[...]

PR #918, opened 2019-01-02
[WIP] refactor(ui): Change Wallet Import page

JeremiGendron:
## Proposed changes

 - "Choose how to import" Label, followed by:
 - Select either Address or Passphrase (blue-button's)
 - Corresponding address: [text-green (address)] if BIP39
 - (Talk about password encryption @ step 2)
 - Modify step 3 header and Wallet name input label

Ref: #902

## Types of changes

## Checklist

## Screenshots

shot1
shot2
shot3
shot4
shot5
shot6
shot7

## Further comments
These are tentative changes, please review.

@doubled1c3 please check out the screen shots and provide feedback.

Other team members as well!

### DIFF
--- a/src/renderer/components/Input/InputAddress.vue
+++ b/src/renderer/components/Input/InputAddress.vue
@@ -112,6 +112,11 @@ export default {
       required: false,
       default: false
     },
+    errorNoValue: {
+      type: Boolean,
+      required: false,
+      default: true
+    },
     value: {
       type: String,
       required: true
@@ -132,7 +137,7 @@ export default {
     error () {
       let error = null
 
-      if (!this.isDisabled && this.$v.model.$dirty && !(this.hasSuggestions && this.isFocused)) {
+      if (!this.isDisabled && this.$v.model.$dirty && !(this.hasSuggestions && this.isFocused) && this.errorNoValue) {
         if (!this.$v.model.required) {
           error = this.$t('INPUT_ADDRESS.ERROR.REQUIRED')
         } else if (!this.$v.model.isValid) {
@@ -150,7 +155,7 @@ export default {
     },
 
     isInvalid () {
-      return this.$v.model.$dirty && !!this.error
+      return this.$v.model.$dirty && !!this.error && this.errorNoValue
     },
 
     model: {
--- a/src/renderer/components/Passphrase/PassphraseInput.vue
+++ b/src/renderer/components/Passphrase/PassphraseInput.vue
@@ -117,6 +117,11 @@ export default {
       type: Number,
       required: true
     },
+    errorNoValue: {
+      type: Boolean,
+      required: false,
+      default: true
+    },
     value: {
       type: String,
       required: false,
@@ -133,7 +138,7 @@ export default {
   computed: {
     error () {
       if (this.$v.model.$dirty) {
-        if (!this.$v.model.required) {
+        if (!this.$v.model.required && this.errorNoValue) {
           return this.$t('VALIDATION.REQUIRED', [this.label])
         } else if (!this.$v.model.isValid) {
           return this.$t('VALIDATION.NOT_VALID', [this.label])
--- a/src/renderer/pages/Wallet/WalletImport.vue
+++ b/src/renderer/pages/Wallet/WalletImport.vue
@@ -27,7 +27,6 @@
             @next="!useOnlyAddress ? moveTo(2) : moveTo(3)"
           >
             <div class="flex flex-col h-full w-full justify-around">
-
               <!-- TODO check duplicate here when db store is available -->
               <InputAddress
                 ref="addressInput"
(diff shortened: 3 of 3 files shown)


doubled1c3:
WOW this is amazing, clean intuitive. Massive score :)

alexbarnsley:
@olejegcord can you review the screenshots above please. I know I sent you the above address/passphrase screenshot and you wanted something different

doubled1c3:
That functionality is retained through the greentext upon successful entry of passphrase, so the user can have a quick look to confirm. @alexbarnsley

alexbarnsley:
And the problem there is when you said "quick look". The purpose before was to make sure the entire address the user had matched the passphrase. Just "looking" to compare addresses is prone to user error.

michaelthecryptoguy:
- [ ] "Choose how to import" Label, followed by:
- [ ] Select  either Address or Passphrase (blue-button's)
- [ ] Modify step 3 header and Wallet name input label
- [ ] Corresponding address: [text-green (address)] if BIP39

> That looks excellent

michaelthecryptoguy:
> * [x]  "Choose how to import" Label, followed by:
> * [x]  Select  either Address or Passphrase (blue-button's)
> * [x]  Modify step 3 header and Wallet name input label
> * [x]  Corresponding address: [text-green (address)] if BIP39
> 
> > That looks **Excellent** `Very Clean!!`

Can you see check marks here, or are they only on my screen?

zillionn:
Why you have to choose anything to import a wallet? Can we just have both fields visible all the time and a text `and/or` between them?
Wallet address: ______________
and/or
Passphrase: _______________

JeremiGendron:
Glad to hear some feedback, will listen for further instructions.

doubled1c3:
@JeremiGendron @alexbarnsley @zillionn 

I went off the starting method of "this/that/show/hide" and proposed improvements to that. We can have both address and passphrase visible and remove the show/hide switches, but it would be a bit more work to make the text fields smart enough to handle the cases:

A. Enter only address, next button becomes clickable, but passphrase can still be entered too
B. Enter only passphrase, if address blank, autocomplete address field, if address not blank, run validation check and display error if problem

This method is good too.

I just feel the switches are kind of wonky and removing the switches and making the text fields smarter, or having two easy button choices, gives a clear workflow of what they are doing and why they are doing it

I settled on the buttons because it does show the user there are two ways to import, which are different from each other, and the user will understand what each of them means. I have spoken to users who imported watch-only and thought for some reason they could control their funds. So for that reason the two method buttons combined with the explainer text (@JeremiGendron can you add explainer text according to mockup https://user-images.githubusercontent.com/32085788/50551291-2c715100-0c33-11e9-9360-5f6cc4393119.png) is pretty nice.

Alex regarding your comment re: validation, the two always visible fields with no switches handles that validation case a bit better, but I do believe the current solution of displaying the greentext is sufficient for confirmation the address is familiar to them...

alexbarnsley:
@doubled1c3 who is "them"? If we're to make it as safe as possible for first time users, surely we should be doing everything we can to make sure they don't lose any tokens? Users who have literally just started out in the crypto/blockchain space are nervous and we should give them a categorical "this passphrase matches this address". It's not like banks, which you seem to be comparing with, it's a 34 character address with both letters and numbers that they have to make sure is correct. I don't want people importing wallets with an address which they believe is correct for their wallet, which isn't.

doubled1c3:
@alexbarnsley Naturally, I agree.

We modify based on dropping the switches, making both boxes always visible, but making them smarter.

a. We drop the Address/Passphrase buttons
b. We drop the Only/Only switches
c. We make both fields always visible
d. If user enters only address, next button become clickable, for watch-only mode
e. If user enters only passphrase, address field autopopulates on successful BIP39 entry
f. If user enters address and then passphrase, it performs a validation against the address the user entered, on successful passphrase entry.
g. We drop the "Corresponding Address:" greentext. It would now be handled through autocompleting the address field.
h. Change main intro text on this step: 

image

becomes

Import wallet
"Enter an address to import in watch only mode, or enter the passphrase to perform a full import and control your funds. Optionally, you may enter both address and passphrase to perform a validation check before importing."

At this point the user can import via address, wallet, both and perform a legit verification, and there are no more switches or buttons to deal with, clean

maybe @JeremiGendron can throw the changes in and we can check out the result from there

olejegcord:
Thank you for telling us your idea, we appreciate it and will take some things into implementation. In this moment we have already updated the UI and UX on the "Wallet Import" page, it will be available in the new updates in the fast time. :)

zillionn:
Use Passphrase (if you own the wallet and want to set encrypted password for faster sending):
`______________________________________________________`
or Address (if you don't own the wallet and want to import it in Watch Only mode):
`______________________________________________________`

If passphrase - populate the `Address` field.

@doubled1c3 It should always populate the address even if not BIP39 passphrase, because some users use custom passphrases. However, checking if the passphrase is BIP39 is a must have, maybe a yellow text below the passphrase field:
`The passphrase you've entered is not a valid BIP39! Please double check your passphrase if you're not using a custom passphrase.`

PR #919, opened 2019-01-02
refactor: new profile creation page

luciorubeens:
## Proposed changes

By Oleg.

captura de tela 2019-01-02 as 07 53 05

TODO:

- [x] Make the instructions side as blocks.
- [x] Change network selection.
- [x] Change theme selection.
- [x] Change the style of the selected background.
- [ ] Optional avatar.
- [ ] Progress bar.

## Types of changes


zillionn:
Oleg, why this gap between the black and white screen? Doesn't it look better if there is no gap and rounding between them?

olejegcord:
The entire purse is stacked with blocks that are separated from each other. At the moment, the information unit has a dark color, exactly the same as on the dark version of the wallet. If the blocks are merged, then on the dark version all the blocks will merge into one dark spot and the user's eyes will not be able to separate the information from the block with filling in the information

zillionn:
Sorry, I'm not sure I understand you. This is what I meant:

profile

Gap between the left menu bar and the content makes sense. But imo separating the content with a gap looks like they are 2 different pages, even though they're both about creating a profile.

JeremiGendron:
I like it a lot

zillionn:
However, black and white looks sick and imo combining them looks like the black is the cover of a book (the info) and white is the content. Very nice!

olejegcord:
and now imagine you've switched your wallet to a dark theme and there's no more separation between black and white and there's only dark and the same color

zillionn:
Well I'm not sure about the correct colors in dark mode, but I'm certain you can make it look good!

However, I'm using dark theme and these two are pretty distinguishable to me:
screenshot 2019-01-02 at 15 08 42

It's up to you, I'm just giving my opinion.

olejegcord:
Thank you for telling us your idea, we will take it into account in future updates and maybe even use it :)

src/renderer/i18n/locales/en-US.js
```diff
@@ -468,15 +468,18 @@ export default {
       STEP2: {
         INSTRUCTIONS: {
           HEADER: 'Network selection',
-          TEXT: 'Choose the network of this profile.'
+          TEXT: 'Choose the network of this profile.',
```
ItsANameToo: I'm not seeing this text anywhere in the stepper

src/renderer/i18n/locales/en-US.js
```diff
@@ -468,15 +468,18 @@ export default {
...
+          CUSTOM_NETWORK_EXPLAIN: 'You can select your custom network or choose any of the available networks.'
         },
         NAME: 'Profile name',
         TITLE: '2. Network'
       },
       STEP3: {
         INSTRUCTIONS: {
           HEADER: 'Appearance',
-          TEXT: 'Customize this application selecting one of our themes and backgrounds.'
+          TEXT: 'Customize this application selecting one of our themes and backgrounds.',
```
ItsANameToo: not seeing this one either in the UI, also maybe add 'by' to the sentence: `... this application by selecting ...`

src/renderer/pages/Profile/ProfileNew.vue
```diff
@@ -178,22 +155,22 @@
 </template>
 
 <script>
+/* eslint-disable */
```
ItsANameToo: why is this needed?

src/renderer/pages/Profile/ProfileNew.vue
```diff
@@ -277,24 +242,12 @@ export default {
...
-          id: network.id,
-          title: network.description,
-          textContent: network.title,
-          svg: `networks/${network.id}.svg`
-        }
-      })
-    },
-    first () {
-      return this.defaultNetworks[0]
+      return NETWORKS.map(network => network)
```
ItsANameToo: can't this simply be `return NETWORKS`?

ItsANameToo:
Also the right part takes up a lot of width when the screen becomes smaller:

From:
schermafbeelding 2019-01-03 om 18 46 38

To this on a slightly smaller screen:
schermafbeelding 2019-01-03 om 18 46 44

And if you reduce the screen width, the `application language` no longer fits on a single line:

schermafbeelding 2019-01-03 om 18 48 25

src/renderer/pages/Profile/ProfileNew.vue
luciorubeens: For some reason, the length appears empty when reloading the window.

olejegcord:
You haven't seen the entire adaptive version. On the definite sizes, the information block is removed.

ItsANameToo:
@olejegcord this is on an even smaller screen: 

schermafbeelding 2019-01-03 om 19 07 08

olejegcord:
This view should be displayed only on the mobile device, preferably on the size of the phone

src/renderer/i18n/locales/en-US.js
luciorubeens: https://github.com/ArkEcosystem/desktop-wallet/blob/c8a4ef5193389b82c6ff3cd42a8cc0cb35e75977/src/renderer/pages/Profile/ProfileNew.vue#L12

ItsANameToo: I don't see it in the UI anywhere tho 🤔 unless I overlooked it

ItsANameToo: ah I see, it's in the left part of the screen

luciorubeens:
@ItsANameToo fixed, take a look

alexbarnsley:
Looks sexy! It isn't showing extra networks for me though 🤔 

image

image

luciorubeens:
@alexbarnsley That seems a bug with [`Object.values` on vuex bindings](https://forum.vuejs.org/t/object-values-on-vuex-chrome-v65/29603/2). I solved by removing the computed cache.

alexbarnsley:
Looks good! I think we can address the responsiveness from the other comments in another PR

PR #921, opened 2019-01-02
misc: ui tweaks and typos

dated:
## Proposed changes

- instead of just showing the address when renaming a wallet, show the currently set name if available
- renames `isSuccessfulRsponse` to `isSuccessfulResponse`
- renames `doesNotExists` to `doesNotExist`
- disables the chart button of the currently selected period

## Types of changes

## Checklist


src/renderer/components/Wallet/WalletRenameModal.vue
```diff
@@ -1,10 +1,11 @@
 <template>
   <ModalWindow
     :title="isNewContact ? $t('WALLET_RENAME.TITLE_ADD') : $t('WALLET_RENAME.TITLE')"
+    container-classes="WalletRenameModal"
     @close="emitCancel"
   >
     <div class="flex flex-col justify-center">
-      <p>{{ $t('WALLET_RENAME.ADDRESS_INFO', { wallet: wallet.address }) }}</p>
+      <p>{{ $t('WALLET_RENAME.ADDRESS_INFO', { wallet: walletName }) }}</p>
```
alexbarnsley: Looking at it now, i wonder if it would be better to bold the wallet name / address here so it's more obvious. Because of the `...` and brackets, it felt a little hard to read

PR #930, opened 2019-01-04
fix: set default fees on transfer screen so send all properly works

ItsANameToo:
## Proposed changes

Reported to me in slack by tk0n: the send-all button didn't take into account the 0.1 fee on v1 networks. After looking into it, this seems to be caused by the fees being set on submit, instead of when the form is created. So while using the form, it defaults to a fee of 0 if you haven't touched the inputFee component (which is not available on v1) and will incorrectly calculated the amount you can send. This also seems to be an issue on v2 networks if you don't interact with the inputFee component as it will also default to 0 in that case.

## Types of changes

## Checklist


faustbrian:
Then it is time to start adding the first e2e test for that.

ItsANameToo:
nooo, ok fine

ItsANameToo:
So looking at this, an e2e test will not be the best approach since it will require a whole workflow to get to the proper page; meaning that half the wallet will be tested just to get to that transaction screen.

Unit test should work, but the `TransactionFormTransfer` component requires a lot of other components, store things, other defined functions like bip38worker, etc and will take a lot of time to properly mock everything. For now I'll leave the PR as is, and might look into the test if I have time for that or if someone is willing to help out with it.

Issue #928, opened 2019-01-04
Display transactions as "sent to yourself"


dated:
**Is your feature request related to a problem? Please describe.**
When sending a transaction from address A to address B, both owned by the user, the dashboard displays such transactions as either incoming or outgoing, when in fact they are both (from the dashboards perspective that is).

**Describe the solution you'd like**
The dashboard should show such transactions with a another, third icon.

**Screenshots**
![image](https://user-images.githubusercontent.com/6547002/50690105-763faa00-102c-11e9-94a1-8752dd553907.png)

**Additional Context**
This requires some refactoring in how transactions are merged, as currently multiple transactions with the same ID are overwritten, when instead the `isSender` / `isReceiver` properties should be combined.

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

PR #934, opened 2019-01-05
[WIP] feat: different display for transaction sent to self

dated:
## Proposed changes

Implements #928.

Also fixes:

 - watch only wallets were not flagged as such

image

The screenshot shows four of the possible scenarios.

1) tx sent from other wallet to own wallet
2) tx sent from own wallet to the same wallet
3) tx sent from own wallet to other own wallet
4) tx sent from own wallet to other wallet

## Types of changes

## Checklist


zillionn:
IMO you should completely remove the +/- signs on last transactions page, because I have wallets that are not mine - watch only. Like the explorer's homepage https://explorer.ark.io/
Maybe use only `±` when sender and recipient is the same.

dated:
I've changed it as such that the new prefix and icon get applied only to transactions where sender === recipient or to transactions where both wallets are owned and have not been imported as watch only. The watch only flag was never set however during the wallet import, so you'd need to reimport those wallets.

boldninja:
I agree with @dated here is also for visually impaired (color) so if they can see somehow real representation (in - out - not yet confirmed) without the need to hover over to see the actual confirmations etc... would be good as well.

dated:
I did only introduce the yellow dot for neither exclusively incoming (green) or outgoing (red) ;)

Yellow was a poor choice, i agree with that.

Issue #938, opened 2019-01-06
Custom peer refresh button unintuitive


JeremiGendron:
**Describe the bug**
When connected to a custom peer, it isn't evident that the refresh button will disconnect from the peer and fetch the best one.
 
**How to reproduce the bug**
Steps to reproduce the behaviour:
1. Go to 'Network'
2. Click on 'Connect custom peer'
3. Enter peer details and connect
4. See no change in refresh button

**Expected behaviour**
Visual queue of what's going on should be more obvious.

PR #939, opened 2019-01-06
feat: Clarify custom peer disconnection

JeremiGendron:
## Proposed changes

 - Implement v-if/else-if/else switch for peer/custom/disconnecting
 - Provide "Disconnect from peer" instead of "Connect to best" when connected to custom peer
 - Automatically refresh (and show through refresh button rotation) when user disconnects from peer
 - Use the cross svg with matching styling to the refresh button
 - hover:bg-theme-option-button-hover hover:text-grey-light

Looks and feels very nice. Planning to add more functionality to the custom peer interface as mentionned in the Ref below. I think we should provide a timeout window for reconnecting when a custom peer disconnects which the user can manually terminate (alongside the presented disconnect button).

Fixes: #938
Ref: #981

## Types of changes

## Checklist

## Further comments

image

image

After click
image

After disconnect
image


alexbarnsley:
I think it looks good. My only tweak would be to add a hover-over to the reload button. Possibly to match the reload button hover on the import page, but the dark mode colour 🤔

JeremiGendron:
alright, I'll do that now @alexbarnsley

alexbarnsley:
thanks @JeremiGendron 👌

JeremiGendron:
And light theme
image

ItsANameToo:
@JeremiGendron just a note, it's better if you leave out the stuff between brackets after the commit keywords like `feat` and `refactor`, since they don't add any value in this repo. It's used in core to denote the package that is worked on, but here we don't have different packages

JeremiGendron:
@ItsANameToo got it, thanks!

src/renderer/components/App/AppSidemenu/AppSidemenuNetworkStatus.vue
```diff
@@ -196,6 +224,7 @@ export default {
 
   methods: {
     async refreshPeer () {
+      console.log('clicked')
```
alexbarnsley: debug

alexbarnsley:
Just a debug line that needs removing then can be merged 👌

PR #943, opened 2019-01-07
refactor: ledger throw errors

alexbarnsley:
## Proposed changes

Also fixes issue with accountIndex validation

## Types of changes

## Checklist


src/renderer/store/modules/ledger.js
```diff
@@ -308,9 +308,8 @@ export default {
         })
       } catch (error) {
         logger.error(error)
+        throw new Error(`Could not get wallet: ${error}`)
```
luciorubeens: Should not it display the error for the user with the alert component?

alexbarnsley: The 2 possible reasons it could get thrown are:
- "Ledger not connected" or "Ledger disconnected"
- "accountIndex must be a Number"

The former will already show to the user when a Ledger has disconnected from the wallet. The latter is more of a background/internal issue which wouldn't make much sense to the user, so in theory should never show as it would be more a bug. I did it this way for the case of reloadWallets, whereby it was returning false for `ledgerWallet`, which could have been handled incorrectly.

alexbarnsley:
thanks @luciorubeens :lala:

PR #947, opened 2019-01-09
feat: optional profile avatar

luciorubeens:
## Proposed changes

This PR introduce the ability not to choose an avatar for the profile, instead the first letter of the profile name will be displayed.

button-letter

## Types of changes


boldninja:
@luciorubeens did we add this switch also on Profile creation page?

zillionn:
Looks very good, not sure what's the point of the switch, but it's not that important I guess.

I kind of don't like the description on the left side:
`Enter your name or nickname and select your preferred language and default currency.`
How about?
`Enter your name/nickname, select preferred language and default currency.`

zillionn:
@luciorubeens I meant that you can just remove the switch and show all avatars but the default one to be the first letter. And `Avatar for your profile` could become `Choose a profile image` or something.

j-a-m-l:
I think that @zillionn idea would be the best approach, @luciorubeens. In fact, it could be used to include the network icon as another avatar, as other users have suggested.

luciorubeens:
@j-a-m-l @alexbarnsley Changes implemented, can you take a look?

luciorubeens:
@j-a-m-l Agreed. Done

PR #950, opened 2019-01-09
fix: do not display the not-voting icon before knowing if the wallet is voting

j-a-m-l:
## Proposed changes
When navigating to 1 wallet, or switching from 1 to other, the orange dot that warns about that the wallet is not voting, is displayed even when the wallet is voting while fetching data.

That was very annoying, and this PR changes the behaviour and it doesn't show it until checking on the server that the wallet is not voting.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletHeading/WalletHeadingPrimaryActions.vue
+++ b/src/renderer/components/Wallet/WalletHeading/WalletHeadingPrimaryActions.vue
@@ -112,7 +112,7 @@ export default {
 
   watch: {
     // Never show the not-voting icon until knowing if the wallet is voting or not
-    'currentWallet.address': () => {
+    'currentWallet.address' () {
       this.showNotVoting = false
     },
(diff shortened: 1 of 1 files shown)


src/renderer/components/Wallet/WalletHeading/WalletHeadingPrimaryActions.vue
```diff
@@ -109,6 +110,17 @@ export default {
     }
   },
 
+  watch: {
+    // Never show the not-voting icon until knowing if the wallet is voting or not
+    currentWallet () {
```
luciorubeens: Watching this property may not be the best approach, as this reacts randomly. This will works better:

```suggestion
    'currentWallet.address' () {
```

j-a-m-l: Thanks @luciorubeens

luciorubeens:
Sorry, the arrow function that I suggested broke the component context (`this`). I just pushed the fix.

PR #955, opened 2019-01-11
fix: check if contact address is already present

dated:
## Proposed changes

Adds validation to the new contact form, disabling the first steps `Next` button and the `Done` button and showing a helper text if the address or name has already been added as a contact.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/pages/Contact/ContactNew.vue
+++ b/src/renderer/pages/Contact/ContactNew.vue
@@ -30,6 +30,8 @@
               <InputAddress
                 ref="addressInput"
                 v-model="schema.address"
+                :is-invalid="$v.schema.address.$invalid"
+                :helper-text="addressError"
                 :pub-key-hash="session_network.version"
                 class="my-3"
               />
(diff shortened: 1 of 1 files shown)


dated:
That's intended, I didn't see a problem with wanting to add a wallet as a contact. It checks for already present contacts only.

dated:
So while i wouldn't forbid it, a notice would be appropriate that reminds the user that the address has already been added as a wallet.

j-a-m-l:
The current implementation doesn't allow duplicating wallets and contacts. Even in your PR and error toast is triggered after clicking on "Done", so it should be forbidden.

Keep in mind that adding an existing wallet as a contact doesn't add any value and it promotes confusing behaviours. For instance: which labels would be used to display it on the transaction table?

dated:
Noted! I'll change it accordingly, and differentiate between address already present as wallet and address already present as contact in the helper text.

j-a-m-l:
I like your idea of showing a different message if the contact address already exists as a wallet, but:
 - Change the message when there is a contact already, instead of using "The wallet address X already exists"
 - Use the error (red) colour on the messages for those messages
 - Check that the contact name is not a wallet name yet. This is not forbidden but it should be to not confuse users, or, at least, a message (on grey) should be displayed.

PR #956, opened 2019-01-11
refactor: multi-wallet endpoints

alexbarnsley:
## Proposed changes

This implements the use of the multi-wallet search endpoints. These are currently only available on devnet, but it will automatically switch between the search endpoint and the "1 request per wallet" endpoints, depending on if it's available. It does this check on each profile change, by doing a spoof search request to see if it can search for many at once.

This has also been incorporated into the ledger wallet loading, so it batch loads 10 wallets, then gets the data for all 10, and keeps doing so until it hits a cold wallet.

**This needs extensive testing on mainnet and devnet since it is important changes to how we get data**

## Types of changes

## Checklist


j-a-m-l:
@alexbarnsley in a brief test I didn't detect any problem. Any suggestion about most bug-prone things to try?

alexbarnsley:
@j-a-m-l I also couldn't reproduce any issues. But I'd say the most bug prone is ledger usage with many wallets. It will still be a problem on a mainnet profile but should be faster on devnet. My tests loaded 100 ledger wallets in 35 seconds on devnet. Took 50 seconds on mainnet.

But I guess as long as everything works, it's all good

alexbarnsley:
@luciorubeens great, thanks! Interesting, possibly something we can review. The stuff in the watch is mostly for checking for new transactions and updating the confirmation times. However, that may not be necessary when we start caching all transactions for wallets 🤔

PR #965, opened 2019-01-15
 refactor: display app version in the window title

luciorubeens:
## Proposed changes

Change the release filename, add version to the window title and fixes the productName.

## Types of changes


alexbarnsley:
My approval is based on the above being correct. Merge when you're happy @luciorubeens

PR #973, opened 2019-01-16
fix: Provide chart tooltip when hovering on x-axis

JeremiGendron:
#966


luciorubeens:
Finally! Thanks @JeremiGendron

PR #986, opened 2019-01-17
fix: persist wallet layout

alexbarnsley:
## Proposed changes

The change to layout between grid and tabular wasn't persisting between reloads. This fixes this, along with some renaming as "layout" was too ambiguous.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/pages/Contact/ContactAll.vue
+++ b/src/renderer/pages/Contact/ContactAll.vue
@@ -6,7 +6,7 @@
 
         <div class="flex items-center">
           <button
-            v-if="!hasGridLayout"
+            v-if="!hasWalletGridLayout"
             class="ContactAll__CreateButton"
             @click="createContact"
           >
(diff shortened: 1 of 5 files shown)


src/renderer/pages/Contact/ContactAll.vue
```diff
@@ -176,18 +176,18 @@ export default {
...
 
-    sessionLayout: {
+    walletLayout: {
       get () {
-        return this.$store.getters['session/layout']
+        return this.$store.getters['session/walletLayout']
       },
       set (layout) {
-        this.$store.dispatch('session/setLayout', layout)
+        this.$store.dispatch('session/setWalletLayout', layout)
```
j-a-m-l: Why not adding an action to the profile store and use it in both contacts and wallets?

alexbarnsley: This is used in both the wallets and contacts. Contacts are still technically wallets, just not yours, so I thought this was a better name than just "layout". We could split it into `setWalletLayout` and `setContactLayout` in the future if people have different requirements for each I think

j-a-m-l: Thanks, @alexbarnsley, fo the explaination.

PR #991, opened 2019-01-18
feat: different style for names provided by the network & ui fixes

dated:
## Proposed changes

image

When a wallet or contact has no name assigned, but a name can be retrieved from the network, it is now shown in a different color and a tooltip shows up when hovering the name.

Also fixes some of the issues described in #997:

- merge wallets and ledger wallets in wallet list and wallet sidebar, so that there are no duplicates in the list
- respect numerical order when ordering by wallet name (e.g. 1 -> 2 -> 10 instead of 1 -> 10 -> 2)
- add ledger badge to wallet sidebar

And also:

- makes the whole wallet/contact card clickable instead of only the identicon and name
## Types of changes

## Checklist


dated:
ready for review, yea :)

dated:
Will be fixed by EOD!

j-a-m-l:
Now, everything looks OK, except
 - Making the whole card clickable: I like it, but selecting the balance or name implies clicking accidentally on it, so I wouldn't include this change. What do you think @alexbarnsley @luciorubeens?
 - The create contact card now is misaligned

dated:
I fixed the create contact card, making it clickable as a whole as well.

j-a-m-l:
@dated On mouse over the links should be underlined.

dated:
Only when hovering the name like before or when hovering the whole card (this felt a bit off, hence i didn't add the text-decoration)?

j-a-m-l:
At least when hovering the names.

When hovering the whole card would be another clue that clicking on it could be used to navigate to the wallet/contact, although I'm not completely sure if it would be the best decision aesthetically and for UX.

dated:
Has this been decided on yet?

j-a-m-l:
@alexbarnsley @luciorubeens what's you opinion about https://github.com/ArkEcosystem/desktop-wallet/pull/991#issuecomment-456052224 ?

dated:
I added/changed the following:

- removed click event on card name and balance, so they can be highlighted and copied
- added ledger badge on the wallet cards
- if sidebar is slim
  - added ledger tooltip
  - made ledger badge square

Issue #993, opened 2019-01-18
Can't properly select balance in wallet view


Azure-Infinity:
**Platform (please complete the following information):**
 - OS: Windows 7
 - Version: Desktop Wallet 2.2.0 (existed with previous V2 wallets)

**Describe the bug**
The wallet ID icon (not the avatar) seems to be getting in the way while trying to select your balance or address.

**How to reproduce the bug**
Steps to reproduce the behaviour:
1. Open ARK Desktop Wallet.
2. Click the "My Wallets" icon on the left side.
3. Click on any of your wallets.
4. Try to manually select your total balance or ARK address on the top.

**Expected behaviour**
I expect to be able to manually highlight my ARK address and also my total ARK balance without making the wallet select everything.

**Screenshots**
![capture](https://user-images.githubusercontent.com/46818126/51391700-ac9c1f80-1b2a-11e9-895e-16dd26c938f1.JPG)

j-a-m-l:
@Azure-Infinity I've changed, on #994, the text to be selectable (part of other widget was hiding the text).

PR #994, opened 2019-01-18
fix: allow selecting the balance on the wallet page heading

j-a-m-l:
## Proposed changes
Resolves https://github.com/ArkEcosystem/desktop-wallet/issues/993

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/components/Profile/ProfileRemovalConfirmation.spec.js
+++ b/__tests__/unit/components/Profile/ProfileRemovalConfirmation.spec.js
@@ -1,0 +1,101 @@
+import { shallowMount } from '@vue/test-utils'
+import useI18nGlobally from '../../__utils__/i18n'
+import { ProfileRemovalConfirmation } from '@/components/Profile'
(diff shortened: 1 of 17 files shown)


alexbarnsley:
Maybe it's worth going a little further than this. Using the `user-select` CSS property on the whole identicon, the name and address (since that can be copied with the button).

j-a-m-l:
@alexbarnsley what do you mean? The text is already selectable. Do you refer to change it on other parts of the application?

alexbarnsley:
Sorry I was clear, preventing the selection on the whole identicon, the name and address. `user-select: none`. There's probably a tailwind class for it

dated:
there is, `select-none`

j-a-m-l:
OK, now I know the reason, but I think that it's better to be practical in that sense and allow users to select the name and address, just in case they want to use it for whatever reason.

I've changed the badges of the identicons and the placeholder of the contacts page.

PR #1006, opened 2019-01-21
refactor: new wallet selection styles and ui

dated:
## Proposed changes

This PR increases the UI/UX of the wallet selection on the new wallet page by:

- making the whole list element clickable instead of only the address, thus increasing the clickable area
- reducing the opacity as well of unselected wallets instead of only changing the color of the address

wallets

[the transition delays are caused by my 🥔]

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/pages/Wallet/WalletNew.vue
+++ b/src/renderer/pages/Wallet/WalletNew.vue
@@ -70,29 +70,25 @@
             </div>
 
             <TransitionGroup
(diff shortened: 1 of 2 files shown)


j-a-m-l:
I like the style on the light theme, but on the dark theme the addresses are very hard to read. Using more opacity would solve that.

dated:
image

Issue #992, opened 2019-01-18
List view


zillionn:
**Is your feature request related to a problem? Please describe.**
Personally I don't like the icons in the action row, it's also not clear at all what they do. You have to click to find out.

**Describe the solution you'd like**
At least add tooltips `Rename` and `Delete` when hover the icons.

zillionn:
Also imo the button `+ Create contact` should always be visible and just remove that card `Create contact` when in card view.

olejegcord:
> Edit icon should be the same as in the profile.
> 
> <img alt="screenshot 2019-01-18 at 13 26 55" width="475" src="https://user-images.githubusercontent.com/3462869/51384347-bde22d80-1b24-11e9-8c69-2c0e973bf7fc.png">

**Using one edit icon throughout the wallet, you enter the user into a strong misconception by yourself without realizing it.** 
In this case, the icon suggests that only the name of the profile is being edited and no more

j-a-m-l:
We will made some of those modifications.

PR #1007, opened 2019-01-21
feat: adapt the contacts page to new design

j-a-m-l:
## Proposed changes
This PR updates the design of the contacts page.
Solves https://github.com/ArkEcosystem/desktop-wallet/issues/992 (some of the requested points were rejected).

## Types of changes

## Checklist


dated:
```
const numeric = value ? parseFloat(value.toString().replace(',', '.')) : 0
```

This will fail for numbers that have the following format: `1,234.56`

Something like the following should do the trick (untested)

```
const numeric = value ? parseFloat(sanitizeInput(value)) : 0
```
with
```
sanitizeInput (value) {
  const parts = value.toString().replace(',', '.').split('.')
  return parts.slice(0,-1).join('') + '.' + parts.slice(-1)
}
```

`toString()` can probably be omitted.

src/renderer/pages/Contact/ContactAll.vue
```diff
@@ -1,27 +1,46 @@
...
-                class="text-center"
-              />
-            </span>
-
-            <span class="flex items-center h-10 px-4">
-              {{ $t('PAGES.CONTACT_ALL.CREATE_CONTACT') }}
-            </span>
-          </button>
+  <div class="ContactAll flex flex-col">
+    <div class="ContactAll__header flex flex-row space-between bg-theme-feature rounded-lg px-10 py-2 mb-3">
```
alexbarnsley: Can we change this to py-5 instead of py-2, I think the identicon looks a little close to the edge:

py-2:
image

py-5:
image

dated:
It might be worthwhile to think about a design where the top bar is not always present. The page in itself is pretty self explanatory and I find the bar a bit superfluous just for the `create` button.

src/renderer/pages/Contact/ContactAll.vue
j-a-m-l: I've changed it. I was feeling the same way, but I tried to follow the original design.

j-a-m-l:
@dated it might be a little superfluous now, but we could add, in the near future, more buttons, such as a way to filter contacts, etc.

src/renderer/pages/Contact/ContactAll.vue
j-a-m-l: @olejegcord could you confirm this change?

olejegcord: Please make "I" more bold 2-3 times.
"py-5" - ok =)

j-a-m-l:
Increased the font weight of the "i" and conflicts resolved

Issue #999, opened 2019-01-20
Pasted amount of ARK not included in transaction


bakkiepleur:
**Platform:**
 - OS: Windows 10
 - Version Desktop Wallet 2.2.0

**Describe the bug**
When you paste an amount of ARK it won't get included in a transaction. In my case, I copied the amount of ARK from my calculator app (default windows 10 calculator) and pasted it in the 'amount' inputfield of the transaction pop-up. After signing the transaction on my ledger I reviewed the transfer pop-up and I noticed that it only contained my transaction fee.

**How to reproduce the bug**
Steps to reproduce the behaviour:
1. Go to your wallet you want send ARK from
2. Click on the 'send' button
3. Paste the amount of ARK from your clipboard in the 'amount' inputfield
4. Click the 'next' button
5. Sign the transaction on your ledger
6. In the transfer pop-up your amount will be 0.00 ARK (the pasted amount did not get through).

**Expected behaviour**
I expected the pasted amount of ARK to be in my transaction.

j-a-m-l:
On Linux, I'm trying to paste `122,01`, and it's truncated to `122`

j-a-m-l:
@bakkiepleur when you pasted the amount on the amount input, does it keep the right number?

I've tried it on a Windows 10 VM and pasting the amount and sending the transactions worked flawlessly.

bakkiepleur:
@j-a-m-l 

**Did it keep the right number?**
In the inputfield where I pasted it in, yes.

**Was the amount changed after clicking on "Next"?**
Yes, it went to 0,00 in the next pop-up.

**Which amount did you used?**
0,07771999 ARK

j-a-m-l:
Thanks, the screenshot is not necessary.

dated:
Did the pasted amount include 'ARK'? I adjusted my comment in the PR to take care of that as well.

bakkiepleur:
@dated no, just 0,07771999

bakkiepleur:
@j-a-m-l I will try and record a video tonight (in +- 8 hours from now). Are you on Slack where I can send it to you?

j-a-m-l:
@bakkiepleur yes, as `jaml`. You could send it there, although I recommend using something like https://cloudconvert.com/mkv-to-gif to convert videos to GIF. That way we could include the GIF here, so it can be visible for everyone.

bakkiepleur:
[...]

PR #1009, opened 2019-01-22
fix: pasting or writing numbers with colon (",") on the send transfer inputs

j-a-m-l:
## Proposed changes
This PR fixes, at least, 1 of the problems of https://github.com/ArkEcosystem/desktop-wallet/issues/999

## Types of changes

## Checklist


dated:
```
const numeric = value ? parseFloat(value.toString().replace(',', '.')) : 0
```

This will fail for inputs that have the following format: `1,234.56`

Something like the following should do the trick (untested)

```
const numeric = value ? parseFloat(sanitizeInput(value)) : 0
```
with
```
sanitizeInput (value) {
  const parts = value.toString().replace(',', '.').split('.')
  return parts.slice(0,-1).join('') + '.' + parts.slice(-1)
}
```

`toString()` can probably be omitted.

j-a-m-l:
Yes, @dated, I was aware of that (and that we would need to support `1.233.233,12`), but I'm waiting to get more feedback from https://github.com/ArkEcosystem/desktop-wallet/issues/999, to include more cases if necessary, such as `100 000.00` or `100_000.00`.

zillionn:
```js
Number(
  amount.toString()
   .split(' ').join('')
   .split(',').join('')
   .split('_').join('')
)
```

dated:
That would result in `123,45` to become `12345`. The above replace call with the regex should cover all cases outlined by @j-a-m-l.

zillionn:
I see but I don't understand which program is generating `123.45` as `123,45`? There is no such string/number... Comma is used for thousands only.

dated:
This ist highly dependent of localisation.

boldninja:
Some use commas for thousands (a lot of EU) and some use dots (US/UK).
For example:
In Slovenia: 100.200,30
In English: 100,200.30

zillionn:
If that's the case, then I guess you need to use something like this throughout the whole app:
https://github.com/globalizejs/globalize

You have `Application Language` in the profile settings and you should follow that setting. Even remove the `Time Format` setting.

j-a-m-l:
It looks like it has more support for some things than https://github.com/kazupon/vue-i18n, but, since we aren't having troubles with localization yet, I'd stick with the current library to not make a very big refactor without any big reason.

About following the language setting, keep in mind that people could copy and paste amounts from sites on other languages.
In some countries, people are used to both time formats and it's just a matter of preference.

zillionn:
Didn't look at the code, but if I put `1,234` how much is that `1234` or `1.234`?

j-a-m-l:
It depends on the locale: if you locale uses "." as decimal separator, it would assume 1.234, if not, it would use 1234

zillionn:
Oh, you're checking the locale, nice!

PR #1015, opened 2019-01-23
feat: allow expanding and collapsing the wallet/contact sidebar

j-a-m-l:
## Proposed changes
This feature is the first part of changes that add more functionality to the wallet sidebar. This PR is focused solely on expanding / collapsing the sidebar.

## Types of changes

## Checklist


alexbarnsley:
I think it looks good 👌 I get a weird "bug" where the bottom border of each wallet goes white when it's expanding though.

Regarding the filter stuff, I think it's worth either removing it all from this PR and adding it all together in its own PR, or adding the filter stuff to this PR. I think it's best to have commits/PRs merged as production ready, whereas this one technically wouldn't be since the filter button isn't ready yet.

A third option could be that once this PR is approved, don't merge to develop but branch off this branch and do the filter stuff. Then it can be merged into here and then all merged to develop. This means they can be both reviewed independently without either getting too large

j-a-m-l:
@alexbarnsley I think it's a good idea. I've removed the filters part, but I'd like to resume it after this PR is merged.

About that "bug", I don't see any weird thing, but, by your explanation, I guess that it's because the bottom border is only applied on the expanded version.

luciorubeens:
@j-a-m-l I've fixed conflicts and approved, now you can merge to your branch.

j-a-m-l:
Thanks @luciorubeens

dated:
How about

sidebar

The framerate isn't ideal but you get the idea.

j-a-m-l:
@dated I like the idea. If you want to create a PR, I'll review it. If not I'll implement it myself

PR #1016, opened 2019-01-24
fix: inconsistencies between network versions

alexbarnsley:
## Proposed changes

This PR fixes a few things:
- Broadcasting when there's no peers
- Finding peers on v1 networks
- Dealing with v1 endpoints on v2 networks (possibly added as a result of API headers being different - E.g. Phantom's vendor header for API version)
- Getting peers from a peer that uses https
- Transaction timestamp based on current network's Epoch, not Ark's

## Types of changes

## Checklist


alexbarnsley:
:crazypuppy: thanks darling

src/renderer/store/modules/peer.js
```diff
@@ -100,7 +100,7 @@ export default {
      */
     best: (_, getters) => (ignoreCurrent = true, networkId = null) => {
       const peers = getters['bestPeers'](undefined, ignoreCurrent)
-      if (!peers || !peers.length) {
```
dated: why was this removed? 🤔 i guess it still works as intended, but the modified condition does not stop the execution as intended.

PR #1021, opened 2019-01-25
feat: add sidebar animation

dated:
## Proposed changes

Adds an animation to the sidebar upon expanding or collapsing and fixes the "white border issue" caused by the transition on the bottom border. I deemed the fading out and in necessary since there is no proper way to transition the flex-direction property.

sidebar

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletSidebar/WalletSidebar.vue
+++ b/src/renderer/components/Wallet/WalletSidebar/WalletSidebar.vue
@@ -6,7 +6,7 @@
       'WalletSidebar--collapsed': !isExpanded,
       'WalletSidebar--expanded': isExpanded
     }"
-    class="WalletSidebar justify-start pt-0 overflow-y-auto"
+    class="WalletSidebar justify-start pt-0 overflow-y-auto menu-transition"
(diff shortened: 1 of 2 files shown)


j-a-m-l:
@dated, could you try fading out / in only the namd and balance, so the identicon is always visible?

dated:
I can give it a go, but i think it will look a bit funky as well since the navigation items are changing in height, hence the identicons will be moving around in two dimensions.

dated:
With the identicons visible it just doesn't look smooth and good. The change of the height and the bottom border makes the transition rather clunky. A positive sideeffect of this approach is also that the transition is only applied to the parent container and not to every single menu Item.

j-a-m-l:
@dated could you upload a GIF of that new animation?could you upload a GIF of the original animation, but faster?

dated:
Faster:
faster

Identicon always visible:
no-identicon

alexbarnsley:
yeh i think it looks better showing the identicons. since the only thing that should fade is anything that changes 🤔

dated:
your call ;)

PR #1023, opened 2019-01-25
feat: loading screen when broadcasting to many peers

alexbarnsley:
## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Modal/ModalLoader.vue
+++ b/src/renderer/components/Modal/ModalLoader.vue
@@ -29,7 +29,7 @@ import Loader from '@/components/utils/Loader'
 import ModalWindow from './ModalWindow'
 
 export default {
-  name: 'ModalLoading',
+  name: 'ModalLoader',
 
   components: {
     Loader,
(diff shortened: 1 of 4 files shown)


alexbarnsley:
Well the reason I wanted it is because during testing I hit some bad peers on devnet which meant the entire broadcast took a good few minutes to finish, which may have the same result on a bad internet connection. Not showing anything may give the impression it somehow failed to broadcast the transaction at all. The modal itself can be closed after 15 seconds when a close button opens, which I think should be plenty of time to get broadcast unless there are issues. 
I have also added a timeout to the broadcast of 3 seconds per peer which helped reduce the total time down.

This is for the less tech-savvy who may not necessarily trust software. To prevent the whole "It doesn't look like it's doing anything. Has it broadcast or...?", only for the "Successfully broadcast" to be sent a few minute later. It's also to prevent sending a new transaction in case they thought the previous one failed.

luciorubeens:
Got it, makes sense. My concern is with the message displayed:

`It looks like something went wrong. You may close the window but this could cause adverse effects.`

Don't you think the user who does not know what's going on will be worried?

alexbarnsley:
@luciorubeens 	I've updated the message to "It looks like it's taking a while to broadcast your transaction. You may close the window and it will continue to send in the background."

Issue #803, opened 2018-12-20
Delegate Voting easier for "Outside 51"


michaelthecryptoguy:
**Feature Request**
An easier method to find/vote for someone outside the 51 delegates 
**Describe the solution you'd like**
No Alternative or Solution at this time

kalgoop:
We can put up a input field in some modal where user can enter name of delegate he wants to vote (like it was in old version of desktop wallet)

doubled1c3:
Recommending upgrading priority of this beyond "nice to have." A text box for autocompleting a delegate name for voting keeps playing field fair... if a new delegate appears and people want to vote for them but they have zero votes, the pagination is nearly endless to locate that delegate, placing an undue inconvenience on voting for a new delegate vs an incumbent forger... @j-a-m-l

michaelthecryptoguy:
I do agree, although I Know you guys are spending Alot of time breaking and shattering apart v1. Slowing glueing the pieces back together, all for one small piece not to fit right......

j-a-m-l:
@doubled1c3 Yes, since the number of delegates is going to grow over time, this would become more necessary over time.

I'd suggest to accomplish it some kind of filtering on the whole table, instead of autocomplete (which would be a good feature too), because that characteristic allow users to search / locate delegates even when they are not interesting on voting them.

zillionn:
Maybe just make the username textfield and filter on key up:

<img width="989" alt="screenshot 2019-01-07 at 18 10 04" src="https://user-images.githubusercontent.com/3462869/50778810-88794c80-12a7-11e9-8068-b7528a8c0504.png">

doubled1c3:
@zillionn Score. Maybe rename label to "Search delegates..."? I dunno

michaelthecryptoguy:
I don't think that would be a good idea, because that would "somewhat" imply that ark is associating itself with delegates

alexbarnsley:
I'm not sure I agree @michaelthecryptoguy, I don't think it gives any sort of impression. Even so, I don't think there's anything wrong with making it easier to vote since that's a key part of the network. The thing we don't want to associate with is the reward sharing delegates do to their voters, which is entirely optional to them and their proposals.

boldninja:
I like zillionn's approach will ask Oleg for input.

dated:
[...]

PR #1027, opened 2019-01-26
feat: make voting easier with delegate modal

dated:
## Proposed changes

#803 

vote

## Types of changes

## Checklist


dated:
This PR depends on https://github.com/ArkEcosystem/desktop-wallet/pull/1023 and won't work as intended without those changes.

dated:
There is one edge case which currently isn't handled correctly, i.e. when the the vote/unvote transaction gets broadcast and confirmed before the broadcast is finished. The vote details won't be loaded correctly in that case. If you want to try it out in the meantime @alexbarnsley 😉

alexbarnsley:
🤔 	yeh will have a test either today or tomorrow 👌

dated:
the above edge case is handled correctly now.

j-a-m-l:
@dated is this already finished?

dated:
If you can give it a try that'd be great. I'll have another look at it to see if somethings missing.

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+        delegate.name = `${truncate(object.username, 25)} (${this.wallet_truncate(object.address)})`
+
+        return delegate
+      })
+
+      const results = _.orderBy(delegates, (object) => {
+        return object.name || object.address.toLowerCase()
+      })
+
+      return results.reduce((map, delegate, index) => {
```
j-a-m-l: I'd use a different name for `map`; something like `acc`, `all`, `final`, `delegates`, etc.

dated:
I'm unable to reproduce what you are saying unfortunately, the wallet shows as processing after the vote has been sent successfully until the actually confirmed vote transaction has been received.

dated:
Thanks for that, I'll fix that ASAP.

dated:
I'll mark this as WIP again, I want to properly refactor the vote handling as currently there is no persistent state of the yet to be processed votes.

dated:
Can you have another look @j-a-m-l?

dated:
No, it's an error caused by my changes during the loading of a profile into the session. I'll add proper storage of the unconfirmed vote(s) to the profile to fix this and to handle the case where the user would vote for a wallet, wait for the transaction to be sent and close the wallet before the actual confirmation has been received.

j-a-m-l:
@dated there is a `TypeError`:

```
TypeError: Cannot read property 'filter' of undefined
    at Action._callee5$ (webpack-internal:///./src/renderer/services/synchronizer/wallets.js:439:77)
    at tryCatch (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:62:40)
...
```

dated:
For backwards compatibility, when loading a profile, the `unconfirmedVotes` are now assigned if not present.

j-a-m-l:
@dated are you going to include here #1097?

dated:
If this PR is fine I'd fix that in a separate PR @j-a-m-l.

src/renderer/store/modules/session.js
```diff
@@ -186,6 +194,11 @@ export default {
       const profile = rootGetters['profile/byId'](profileId)
       if (!profile) return
 
+      if (!profile.unconfirmedVotes) {
+        profile.unconfirmedVotes = []
+        dispatch('profile/update', profile)
```
j-a-m-l: `[vuex] unknown local action type: profile/update, global type: session/profile/update` on the dashboard

src/renderer/i18n/locales/en-US.js
```diff
@@ -883,7 +894,11 @@ export default {
     VOTE_DELEGATE: 'Vote Delegate {delegate}',
     UNVOTE_DELEGATE: 'Unvote Delegate {delegate}',
     VOTED_FOR: 'You voted for delegate {delegate}',
-    WALLET_VOTED_FOR: 'This wallet voted for delegate {delegate}'
+    WALLET_VOTED_FOR: 'This wallet voted for delegate {delegate}',
+    NO_VOTE: 'You haven\'t voted yet, click on the button to the right to select a delegate',
```
j-a-m-l: Maybe "This wallet is not voting. Click on any delegate or search it to vote"?

src/renderer/i18n/locales/en-US.js
```diff
@@ -883,7 +894,11 @@ export default {
...
+    AWAITING_VOTE_CONFIRMATION: 'Your {type} is awaiting confirmation',
+    LOADING_VOTE: 'Loading vote details...',
+    SELECT_DELEGATE: 'Select delegate'
```
j-a-m-l: "Search delegate"

dated:
> I'd show the select/search button always because it could be used to find delegates

@j-a-m-l that makes sense. I'll work on it later.

src/renderer/i18n/locales/en-US.js
dated: Changed to `This wallet is not voting - use the search or click on any delegate to cast your vote`, does that sound ok?

j-a-m-l: Uhm, I'm not completely sure; users could think that clicking on a delegate would cast their vote directly.

Something like this "This wallet is not voting. To view more information and vote, use the search or click on any delegate"?

j-a-m-l:
Thanks, @dated

src/renderer/i18n/locales/en-US.js
dated:
```suggestion
    NO_VOTE: 'This wallet is not voting. To view more information and vote, use the search or click on any delegate',
```

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+        return this.dropdownValue || this.inputValue
+      },
+      set (value) {
+        this.updateInputValue(value)
+        this.$emit('input', value)
+      }
+    },
+
+    suggestions () {
+      if (!this.currentProfile) return []
```
alexbarnsley: Try to keep if statements consistent with curly brackets as it helps with readability

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+      this.$nextTick(() => this.closeDropdown())
+    },
+
+    onFocus () {
+      this.isFocused = true
+      this.$emit('focus')
+    },
+
+    onEnter () {
+      if (!this.dropdownValue) return
```
alexbarnsley: Try to keep if statements consistent with curly brackets as it helps with readability

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+    },
+
+    async updateInputValue (value) {
+      this.inputValue = value
+      // Inform Vuelidate that the value changed
+      this.$v.model.$touch()
+    },
+
+    __setSuggestion (value) {
+      if (!this.hasSuggestions) return
```
alexbarnsley: Try to keep if statements consistent with curly brackets as it helps with readability

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+        this.$refs.input.setSelectionRange(this.inputValue.length, this.dropdownValue.length)
+      })
+    }
+  },
+
+  validations: {
+    model: {
+      required,
+      isValid (value) {
+        return !!this.$store.getters['delegate/byUsername'](value) ||
```
alexbarnsley: I'd possibly do a `search` getter to do what is done in WalletDetails#onConfirmSelect and use that instead of piping here

src/renderer/components/Input/InputDelegate.vue
```diff
@@ -0,0 +1,362 @@
...
+
+<script>
+import { required } from 'vuelidate/lib/validators'
+import { ButtonModal } from '@/components/Button'
+import ModalQrCodeScanner from '@/components/Modal/ModalQrCodeScanner'
+import { MenuDropdown } from '@/components/Menu'
+import Cycled from 'cycled'
+import InputField from './InputField'
+import truncate from '@/filters/truncate'
+import _ from 'lodash'
```
alexbarnsley: It might be worth loading specifically which lodash modules we want here instead of all of them

dated:
A `Switch Vote` functionality is something I had in mind for a future iteration 👍

boldninja:
:waitasec: I thought that will be in this one :trollface:

dated:
:danstop:

src/renderer/components/Input/InputDelegate.vue
dated: Something like 77f2cc78de65e3df85aaa2c43070aadef9c3ed16?

boldninja:
Guess need to revert tier 1 :trollface:

dated:
Just subtract from it!

boldninja:
I'm kidding :colbert:, but if you can make switch vote in next couple of days in separate PR would be cool before next release.

src/renderer/components/Input/InputDelegate.vue
alexbarnsley: bootiful!

alexbarnsley:
all looking good! just a message I think at the bottom of the delegate modal when already voting elsewhere, which will be replaced by the "switch votes" option when you get round to it 👌

alexbarnsley:
I hate to say it @dated 😩 but I'm getting issues when i hit the "unvote" button in the footer. I'd screenshot but my clipboard is messed up. It's like it can't find the delegate. The modal pops up to unvote but all it says is "Unvote Delegate arkx" 🤔 will try a restart in the morning and do a screenshot then if you still need one. Double check as it may be my local data, but I believe it was working earlier

dated:
I broke that in one of the recent commits, it's fixed now @alexbarnsley.

dated:
There is a little issue still that's most likely caused by the api cache and the way `fetchWalletVote` works: after voting the outdated information is loaded. A proper fix for this will be possible once local storage of the transactions is implemented.

dated:
Alternatively one could leverage the fact, that the wallet vote is stored on the wallet itself, which would eliminate the need to touch the transactions at all.

alexbarnsley:
🤔 	well we have access to the delegate the wallet is voting for don't we. as you said in your last comment (which may be saying what i'm saying), the wallet API stores the delegate that wallet is voting for `vote: "publickey"`

dated:
yea, we are talking about the same

alexbarnsley:
Okay, great. Are you happy with this PR? If so, I'll give it a test

dated:
If you are happy with me changing the vote fetching in a future iteration, then yes 😄 the API call was previously fired everytime `isAwaitingConfirmation` changed, so when a vote transaction was received in under 8 seconds it fetched the cached response, which did not include the last vote. If i'm not mistaken, right now it should fail (= show outdated information) only when all of the following happens in a timeframe of 8 seconds: open wallet -> select delegate -> input passphrase & fee -> send/receive transaction

PR #793, opened 2018-12-19
feat: implement address navigation directly in the wallet

luciorubeens:
## Proposed changes

Implement the ability to open any address directly in the wallet.

Fixes #811

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

PR #1035, opened 2019-01-28
fix: watch only wallet import and secondary actions

dated:
## Proposed changes

This fixes the following:

- when importing a wallet as watch only, the respective flag isn't set on the wallet
- the secondary action buttons to register a delegate and second signature were visible for watch only wallets
- when navigating to an address that is neither wallet nor contact, it should not be loaded as a contact and the secondary action buttons should not be toggeable

## Types of changes

## Checklist

### DIFF
--- a/__tests__/unit/components/Wallet/WalletHeading.spec.js
+++ b/__tests__/unit/components/Wallet/WalletHeading.spec.js
@@ -11,7 +11,17 @@ localVue.use(Vuex)
 const i18n = useI18n(localVue)
 
 const store = new Vuex.Store({
-  state: {}
+  state: {},
+  modules: {
+    wallet: {
+      namespaced: true,
+      getters: {
+        byAddress: () => address => {
+          return sampleWalletData
+        }
+      }
+    }
+  }
 })
 
 const sampleWalletData = {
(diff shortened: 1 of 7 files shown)


alexbarnsley:
🤔 i originally fixed this myself but we reverted for some reason before the first release. Possibly to simply allow the option of importing by address, but still allow sending. I believe there are some people that import by address and send transactions from those wallets. Perhaps a better option may be a toggle when importing, to mark the wallet as watch only instead. I know what you've done is how it works in the mobile wallet though. I'll see what the others think as they originally raised the point to revert it

dated:
Weil might a clear definition of `watchOnly` in that case. As you say this behaviour mimics that of the mobile wallet.

This however covers the case of people adding their ledger addresses as 'watch only'.

dated:
But i get where you are coming from, importing a wallet through either address or passphrase (without encryption) results in the same object being created.

luciorubeens:
We don't need to implement the same behavior as the mobile when importing an address, since on the desktop it is more convenient to paste or type the passphrase. And the variable `isWatchOnly` was introduced [here](https://github.com/ArkEcosystem/desktop-wallet/pull/793), I did this because it's exactly a contact, except it was not imported. Thanks and sorry my late reply.

dated:
I'll have another look at the the changes to the action buttons, as those should not be shown when 'visiting' an address which is neither wallet nor contact.

PR #1044, opened 2019-01-30
refactor: correctly sort wallets and contacts in overview and sidebar

dated:
## Proposed changes

This fixes the sorting by name of the wallets and contacts in the grid view, table view and sidebar.

It also changes the implementation of the `sortByProp` (now `sortByProps`) util, which now uses [natural sorting](https://en.wikipedia.org/wiki/Natural_sort_order) by default and allows for multiple properties.

## Types of changes

## Checklist


dated:
As discussed with Alex, wallets without a name (custom or from the Network) will stay at the top of the list.

dated:
Does that happen with these changes @zillionn ?

zillionn:
No, they're not.

dated:
Thanks for the feedback, I'll take a look asap!

dated:
should be fixed now @zillionn, stupid typo 😉

dated:
Done as requested @j-a-m-l

dated:
Better?

j-a-m-l:
Thanks @dated. Now it's a lot easier to read and it's also easier to grasp what should be expected when sorting.

j-a-m-l:
@dated I just realized that the Ledger wallets are also visible when navigating through the contacts page, so the sidebar is showing both, contacts and owned Ledger wallets. Would you mind to change it?

dated:
I've made some changes to how the wallets are fetched, I can't test it myself right now unfortunately.

src/renderer/components/Wallet/WalletSidebar/WalletSidebar.vue
```diff
@@ -195,11 +195,14 @@ export default {
   }),
 
   computed: {
+    fetchContactsOnly () {
```
j-a-m-l: I'd rename it to `hasContactsOnly`, `includesContactsOnly` or something like that, using "fetch" looks like that method / computed property request the data.

dated: i changed it to `hasContactsOnly`

PR #1059, opened 2019-02-04
feat: load x ledger wallets

alexbarnsley:
## Proposed changes

This feature allows Ledger users to load X amount of Ledger wallets. The reason for this is, for example, some people may want to see a larger amount of wallets. The other reason is for users to access their Phantom tokens where the Airdrop took place on Ledger wallet 10 for example (which may also happen in the future with other chains).
this also improves the loading of ledger wallets and has the ability to stop any currently ongoing process.

## Types of changes

## Checklist


j-a-m-l:
@alexbarnsley I'd not merge this PR until having the Ledger profile settings, when I'd move this setting to that tab, what do you think?

alexbarnsley:
@j-a-m-l I wouldn't consider this a setting. It's a task to load X ledger wallets as per a request from @boldninja. If the ledger wallets aren't cached then the list of ledger wallets resets on profile change or wallet reload

ItsANameToo:
Some issues while using it:

1. I can select `0` wallets to load (it will load 1)
schermafbeelding 2019-03-01 om 16 59 11

2. Text not centered on smaller screen
schermafbeelding 2019-03-01 om 16 58 45

3. When I load 3 wallets, then turn caching on, then want to load 1 wallet; it will load 3 wallets again
4. Maybe "add ledger wallets" is not the right name for the button, as it can also be used to remove ledger wallets? E.g. you are shown 10 and then set it to 3 with that button, you'll have removed 7. I don't know what to call it yet though.

alexbarnsley:
you wasn't supposed to find issues 😭

ItsANameToo:
Looks good. Only thing I run into is that if I have a ledger connected it will spam `[16:40:51.098] Error: Ledger device: Incorrect length (0x6700)` messages. It stops when I open the ark app on the ledger 🤔

Issue #1057, opened 2019-02-03
bug: wallet balance not showing for non-contacts / non-owned wallets


ItsANameToo:
**Platform (please complete the following information):**

macOS

**Describe the bug**

When you navigate to another wallet by using the transaction table, it will show a 0 balance for any addresses that you don't own / don't have as a contact instead of the actual balance of that address. Reported by @alessiodf in slack.

**How to reproduce the bug**

1. import an address with transaction
2. click on a transaction to navigate to a wallet that you don't own in the wallet and don't have added as a contact
3. look at the balance, it will be 0

**Expected behaviour**

Should show the correct balance

ArkEcosystemBot:
@ItsANameToo Thanks for opening this issue, a maintainer will get back to you shortly!

ArkEcosystemBot:
@j-a-m-l @alexbarnsley @luciorubeens - please review this in the next few days. Be sure to explicitly select labels so I know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

ArkEcosystemBot:
@ItsANameToo This issue has been closed. If you wish to re-open it please provide additional information.

PR #1071, opened 2019-02-12
fix: show the right balance on wallets not owned by the user or contacts

j-a-m-l:
## Proposed changes
Resolves https://github.com/ArkEcosystem/desktop-wallet/issues/1057

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Wallet/WalletHeading/WalletHeadingInfo.vue
+++ b/src/renderer/components/Wallet/WalletHeading/WalletHeadingInfo.vue
@@ -210,6 +210,7 @@ export default {
         return
       }
 
+      this.lazyWallet = {}
       this.lazyWallet = await this.$client.fetchWallet(this.currentWallet.address)
     }
   }
--- a/src/renderer/components/Wallet/WalletHeading/WalletHeadingInfo.vue
+++ b/src/renderer/components/Wallet/WalletHeading/WalletHeadingInfo.vue
@@ -140,7 +140,7 @@ export default {
     alternativeBalance () {
       let balance = this.currentWallet ? this.currentWallet.balance : null
 
-      if (balance === null) {
+      if (!balance) {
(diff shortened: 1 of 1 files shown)


dated:
when switching from wallet A (owned) with `balance > 0` to wallet B (owned) with `balance === 0` it will still show the balance from wallet A i believe.

j-a-m-l:
Thanks @dated

dated:
🤔 does this work? `wallet_fromRoute` still returns a 'fresh' wallet with no balance, doesn't it?

```
const freshWallet = () => {
  return WalletModel.deserialize({
    address,
    name: '',
    profileId: '',
    isContact: true,
    isWatchOnly: true
  })
}

return this.$store.getters['wallet/byAddress'](address) || freshWallet()
```

j-a-m-l:
@dated what advantages do you think that has the `freshWallet` approach? I'm not sure about setting default values for some attributes, such as `isContact`, etc.

dated:
I meant that `this.currentWallet` will always fall back to the freshWallet object above, since the looked up wallet is not stored.

j-a-m-l:
Yes, but I'm not sure if it's better transitioning from wallet A, to wallet Fresh, to wallet B, instead of A -> B. It could provoke fast UI changes or things like that. I mean, the current fix tries to not provoke side effects, but maybe I could review your idea later to see how it works on every scenario.

dated:
The snippet above is from the actual implementation of `wallet_fromRoute`. I was merely wondering If this PR actually fixes the problem at hand, which is showing the balance of unowned (or unknown through contacts) wallets.

j-a-m-l:
I didn't find any failing case :man_shrugging:, but you are free to try it; maybe I have missed something.

luciorubeens:
Fixed via #1094

PR #1075, opened 2019-02-13
chore: upgrade to node 11 and fix all failing unit tests

luciorubeens:
## Proposed changes

- Fix all failing unit tests
- Upgrade to node 11 and used the [Intl polyfill](https://github.com/andyearnshaw/Intl.js) to properly run tests
- Refactor e2e tests and add more specs

## Types of changes

- [x] Docs (documentation only changes)

## Checklist


__tests__/unit/mixins/currency.spec.js
```diff
@@ -153,11 +153,11 @@ describe('Mixins > Currency', () => {
     it('should work with big quantities', () => {
       let amount = Math.pow(10, 12) + 0.01
 
-      expect(format(amount, { currencyFrom: 'network' })).toEqual('× 1,000,000,000,000.01')
+      expect(format(amount, { currencyFrom: 'network' })).toEqual('× 1,000,000,000,000.00999424')
```
j-a-m-l: What could be the reason to this? I mean, did something change on Node 11? If that's the case we should try to fix this behaviour or, at least, document it.

luciorubeens: It's related to the [Intl polyfill](https://github.com/andyearnshaw/Intl.js):

> Node.js 0.12 has the Intl APIs built-in, but only includes the English locale data by default

Probably a precision error, but this polyfill is used only for the tests, in the browser will be used the native api, returning the correct data.

You added a note related some lines above: https://github.com/ArkEcosystem/desktop-wallet/blob/develop/__tests__/unit/mixins/currency.spec.js#L77

j-a-m-l: That comment was about the format: since Node was including English only, it formats every currency the same way.

PR #1079, opened 2019-02-15
feat: add the time format setting to the profile creation

j-a-m-l:
## Proposed changes
That setting was on the profile edition, but not on the profile creation

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/store/modules/session.js
+++ b/src/renderer/store/modules/session.js
@@ -145,7 +145,7 @@ export default {
       state.avatar = 'pages/new-profile-avatar.svg'
       state.background = null
       state.currency = MARKET.defaultCurrency
-      state.timeFormat = 'default'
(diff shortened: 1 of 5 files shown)


src/renderer/store/modules/session.js
```diff
@@ -145,7 +145,7 @@ export default {
       state.avatar = 'pages/new-profile-avatar.svg'
       state.background = null
       state.currency = MARKET.defaultCurrency
-      state.timeFormat = 'Default'
+      state.timeFormat = 'default'
```
luciorubeens: In other places was written capitalized:

```suggestion
      state.timeFormat = 'Default'
```

j-a-m-l: @luciorubeens thanks

PR #763, opened 2018-12-15
[on hold] refactor: remote sorting of delegate table

dated:
## Proposed changes

Adds remote sorting to the delegates table. This PR depends on arkecosystem/core#1731 and will be tested once possible.

## Types of changes

- [x] Refactoring (improve a current implementation without adding a new feature or fixing a bug)

## Checklist

- [x] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation
- [x] Lint and unit tests pass locally with my changes

PR #1080, opened 2019-02-15
feat: add remote sorting to delegate table

dated:
## Proposed changes

Adds remote sorting to the delegate table. Replaces #763

## Types of changes

## Checklist


src/renderer/services/client.js
```diff
@@ -137,20 +137,23 @@ export default class ClientService {
    * @param {Object} [query]
    * @param {Number} [query.page=1]
    * @param {Number} [query.limit=51]
+   * @param {String} [query.orderBy='rank:asc']
    * @return {Object[]}
    */
-  async fetchDelegates ({ page, limit } = {}) {
+  async fetchDelegates (options = {}) {
```
j-a-m-l: @dated , I'm curious, why do you prefer using `options` instead of destructuring the object here?

dated: Just for conformity in regards to the other methods, such as `fetchWalletTransactions`.

j-a-m-l:
Thanks, @dated

PR #1081, opened 2019-02-16
misc: configure the new Ark Shield address

j-a-m-l:
## Proposed changes
The address changed, so this PR, adds the new one, but it keeps the former too.

## Types of changes
- [x] Other... Please describe: configuration

## Checklist


j-a-m-l:
@alexbarnsley @luciorubeens @boldninja, should we remove the old address? Should we rename it to "Ark Shield (old)" instead?

dated:
the old address was completely removed on the explorer https://github.com/ArkEcosystem/explorer/pull/561, which makes more sense imho as the old one won't be used again as it seems.

j-a-m-l:
Maybe, but some users might find it useful knowing the former address. @boldninja what do you think?

PR #1102, opened 2019-02-25
chore: stage linted files before commit

dated:
## Proposed changes

Adds the `git add` command to the lint-staged config, to include linted files in commits.

## Types of changes

- [x] Other... Please describe: config

## Checklist

### DIFF
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,3 @@
+{
+    "*.{js,vue}": ["npm run lint:fix", "git add"]
+}
--- a/package.json
+++ b/package.json
@@ -226,12 +226,5 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  },
-  "lint-staged": {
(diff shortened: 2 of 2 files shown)


j-a-m-l:
@luciorubeens @alexbarnsley do you like the idea?

dated:
The reason behind this is that it's very easy to push un-linted files with the current setup.

j-a-m-l:
I understand, but maybe others, don't like the idea of pushing code that has been altered by a linter, without being reviewed: currently you have to accept manually the suggestions, so you are aware of the changes.

Since this change would affect the workflow of all the devs, asking to the rest of maintainers is a must.

dated:
> Since this change would affect the workflow of all the devs, asking to the rest of maintainers is a must.

Sure! Just wanted to explain myself :D the `core` repository uses the setup proposed here, but it's a matter of personal preference in the end.

PR #1108, opened 2019-02-28
refactor: synch dynamic fees only when `InputFee` is active

j-a-m-l:
## Proposed changes
These changes avoid making requests at the beginning to fetch the fee statistics of every network, and then, periodically to have them updated.
Now, only the current network is checked, but just in the case that `InputFee` component is active, because they are only used there.

This also improves the `Synchronizer` service, so, from now on, we could disable completely some actions until needed.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/components/Input/InputFee.vue
+++ b/src/renderer/components/Input/InputFee.vue
@@ -119,7 +119,8 @@ export default {
 
   data: () => ({
     chosenFee: 'AVERAGE',
-    step: 1e-8
+    step: 1e-8,
+    fee: 0
   }),
 
   computed: {
@@ -151,7 +152,6 @@ export default {
       if (this.feeChoices.MAXIMUM === this.feeChoices.AVERAGE) {
         return +this.fee === this.feeChoices.AVERAGE
       }
-
       return false
     },
     isAdvancedFee () {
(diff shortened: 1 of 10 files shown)


src/renderer/components/Input/InputFee.vue
```diff
@@ -203,50 +224,28 @@ export default {
...
-    /**
-     * Even if the network provides a fees higher than V1, it will be corrected
-     */
-    prepareFeeStatistics () {
-      let { avgFee, maxFee } = this.$store.getters['network/feeStatisticsByType'](this.transactionType)
-
-      avgFee = avgFee * 1e-8
-      this.$set(this.feeChoices, 'AVERAGE', avgFee < this.maxV1fee ? avgFee : this.maxV1fee)
+    // Fees should be synchronized only when this component is active
+    this.$synchronizer.focus('fees')
```
luciorubeens: I think the `call` method is enough, these values are not updated with this frequency:

```suggestion
    this.$synchronizer.call('fees')
```

j-a-m-l:
@luciorubeens I've updated the minimum fee part. It was intended to support getting the minimum fee from the statistics, but, since we aren't doing it, it wasn't necessary at all.

About the other change, the idea was keeping the fees as updated as possible only when the `InputFee` is visible (when the component is destroyed, the synchronizer action is paused again). Possibly the majority of users wouldn't use it, but I've added that to not leave any loose ends.

PR #1109, opened 2019-03-01
refactor: improve address list style when creating a new wallet

luciorubeens:
## Proposed changes

- Improve the design when hovering the new address
- Change checkmark flag
- Update refresh address button

## Types of changes


luciorubeens:
@j-a-m-l True that, I did not notice the transitions when refreshing the addresses. Take a look now.

PR #1120, opened 2019-03-11
feat: allow using a pool of background workers

j-a-m-l:
## Proposed changes
This PR introduces the possibility of running jobs using a pool of threads.

Related to that, all the request (except some that are tied to the JavaScript API Client) have been moved to be executed through 1 new service (`src/renderer/services/http.js`) that can send them on the main thread (using `axios`) or on the background.

I've tried to measure the difference between the previous approach and the this one, and the different is very small, so I've disabled the requests on the background by default.

## Types of changes

## Checklist


dated:
Upgrade babel-jest to fix the bug in node v11.11 :wink:

src/renderer/services/background-http-client.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    if (payload.params) {
+      config.params = payload.params
+      delete payload.params
+    }
+
+    config.data = payload
+
+    return this.backgroundClient ? this.backgroundClient.request(config) : axios(config)
```
luciorubeens: This is confusing because it is not declared anywhere in the class, perhaps adding a method (`setClient`) that overwrites the variable.

j-a-m-l: Changed and explained in the comments.

Mrmules:
Could you please help me out.  I cannot sign into my account.  It does not
recognize my email address even though I keep getting multiple emails from
you on a regular basis.  What procedure will I need to get sign back into
ARK/ github??

j-a-m-l:
@Mrmules as you can read at the bottom of the message that you posted, you've received that email because you're subscribed to this PR notifications. There you could see a link to unsubscribe.

PR #1143, opened 2019-03-19
fix: close transaction show modal when clicking on addresses and navigate to transactions tab always

j-a-m-l:
## Proposed changes
Changed the transaction modal to, when clicking on an address, close itself and navigate to that address, to the transactions tab instead of the current tab (which could be delegates, for example).

## Types of changes

## Checklist


alexbarnsley:
Doesn't seem to be working for me 🤔 I click on a wallet, go to delegates, click on the name and it closes the popup and loads the new wallet, but I'm still on the delegates tab. What's interesting is that I cannot then click on a delegate again, I have to go to the wallet list and back into the wallet. It's strange. No errors in devtools. I guess I can't click a delegate because I'm not on one of my wallets - if this is intended then perhaps we need to disable the pointer cursor on mouse over.

The popup does close though when clicking an address from the transactions list which is working 👌 although the tooltip doesn't seem to disappear which I just noticed (it moves to the top-left):

image

j-a-m-l:
@alexbarnsley this change was intended initially for the transaction modal (`TransactionShow`), but it looks like you were trying the vote modal. I've changed it too.

About the tooltip, I've fixed that, although it's possible to see it during some ms. while it's closing. I guess that it's related to how the library works internally.

IMHO, this section should be changed to not be so confused, with examples like the 1 you said (I can click on the delegate on my wallet, but I can't click on other wallet) or others such as displaying the contacts in some occasions and my wallets on others.

ItsANameToo:
Seems to work as intended now (closes modal and navigates to transaction tab). I agree that the delegates tab is confusing as you can't interact with it. IMO it might be useful to hide that tab when we're not viewing a wallet that we own as it happens with every navigation action, thoughts @alexbarnsley ?. Not necessarily related to this PR so this can be merged

alexbarnsley:
If you're happy I'm happy 👌

PR #1146, opened 2019-03-21
feat: add option to hide the wallet button text

### LINKED ISSUE #1141: Option to hide text on wallet buttons
The wallet buttons take up a lot of horizontal space, which makes it rather difficult to add more buttons to the header. This issue proposes to make the display of the text underneath each button optional and show tooltips instead.

image

**Describe alternatives you've considered**
Completely revamp how actions in the wallets page are shown.

dated:
## Proposed changes

Adds the option to hide the text from the buttons on the wallet page, to allow for a more condensed view and make room for additional action buttons. Closes #1141.

image

image

image

Also adds a ledger svg icon.

## Types of changes

## Checklist


j-a-m-l:
We talked about it, but it looks like nobody wrote here our conclusion. Sorry for that.
To avoid it, next time, you could wait until the issue is tagged as feature, bug, or other. In case nobody do it, just ping us.

Those conclusions are that, instead of hiding the text of the buttons, it would be better to add a menu for the Ledger options, since they are probably be used scarcely, so they should not be targets as visible as the other 2.
Other option that we want to introduce in the future is create a new tab on the profile section for the Ledger, so those 2 button would be moved there. IMHO that would be a bigger feature than just an aesthetic change, so it should have a bigger bounty, maybe 1 or 0.

dated:
I can take care of the extra tab on the profile edition tab, but would keep these changes around as well. Personally i prefer the slimmer design, some other people said it'd be nice. Having the option doesn't, does it? :wink:

alexbarnsley:
@j-a-m-l I've said in the past that loading X ledger wallets shouldn't live in a profile tab because it's less of a setting and more of an action that can reset on wallet reload.

I do like the look of the slimmer design, and Oleg has also given the thumbs up when I showed him

dated:
Well, let me know how to proceed when there has been an agreement.

j-a-m-l:
@alexbarnsley immediately after that I explained the target problem and he reacted with a :+1: to move those options to a menu. So, @olejegcord, should we keep those Ledger buttons or should we move them to a menu?

alexbarnsley:
@j-a-m-l yes I am happy with a menu. I was referring to your second option about moving it to the profile screen:

>Other option that we want to introduce in the future is create a new tab on the profile section for the Ledger, so those 2 button would be moved there.

dated:
Just to confirm what the consensus is:

 - keep the option to toggle the text
 - reduce the ledger buttons to one button which opens a Menü

That about correct? If there are some designs for the menu already i'd appreciate it if you could pass them on to me. Otherwise I could either wait or come up with something.

zillionn:
IMO this shouldn't be a setting since it's not that important. The text could be shorter with longer tooltip:

button text: `Create`
tooltip: `create new wallet`

button text: `Import`
tooltip: `import wallet using address or passphrase`

j-a-m-l:
@dated I'd move the Ledger buttons to a menu, and, since doing that would free the same amount of space, I wouldn't remove the text from the buttons or add the configuration option.

What do you think @alexbarnsley?

alexbarnsley:
I actually prefer the buttons without text. It looks cleaner. Perhaps the setting could include buttons elsewhere in the app, not restricted to wallet buttons?

I realise this is digging up an old PR now, apologies for the delay!

olejegcord:
By default, we display buttons with icons http://prntscr.com/ndlcoe , in the settings when switching we show the button with the text http://prntscr.com/ndl0gg . Only the buttons should be square

alexbarnsley:
I think for now @olejegcord we keep the rounded buttons instead of the square buttons since you're showing a future design change. You okay with that?

olejegcord:
ok

dated:
Thanks for the updates on this! I'll try to get back to this ASAP.

zillionn:
Personally I don't like buttons without any text, because it's not clear what the button is for. Like the main buttons on the left panel, you have to click to find out what section they load. IMO icon buttons must have an overlay text.

dated:
@alexbarnsley, as for the Ledger Menu: is there a design already i can implement? I have two ideas otherwise for how it could be done:

1) toggable menu like on the wallet page with primary / secondary buttons
2) trigger a modal from where to change the settings

boldninja:
@olejegcord ^^ your opinion as well on this.

olejegcord:
What do you have in mind? Tell me more.

dated:
I'll make some mockups tonight.

dated:
image

How about this?

dated:
@alexbarnsley @boldninja @olejegcord are you happy with the above design?

boldninja:
I really like this Dated,  @olejegcord your opinion?

olejegcord:
we like it. In the future we will fix the buttons a little bit.  :)

dated:
Cool, yea, i saw your links above. That's a task for another day, and another PR ;)

dated:
Will refactor in the morning!

dated:
>If clicked on cancel, it could be kept open or not, as you prefer.

IMO it's better to keep the menu open in that case. I added the requested change and cleaned up a bit @j-a-m-l

PR #1148, opened 2019-03-22
feat: store sidebar sorting and filters

### LINKED ISSUE #1137: Filter Preference Save
No problem.

**Describe the solution you'd like**
I like the new filters that were implemented, however, I need to set it every time I open the desktop wallet. I would like it to remember my last used preference and save it. 

**Describe alternatives you've considered**
Set it every time I open the wallet?

**Additional context**
None.

dated:
## Proposed changes

Closes #1137.

## Types of changes

## Checklist


src/renderer/components/Wallet/WalletSidebar/WalletSidebarFilters.vue
```diff
@@ -155,7 +142,19 @@ export default {
       },
       sort: {
         order: this.sortOrder
-      }
+      },
+      sortOptions: [
```
j-a-m-l: @dated since this `Array` shouldn't change, it could be moved, frozen, to `$options`.

dated: done

dated:
Ah, finally found what you are seeing. There's a problem with existing profiles.

dated:
Should be fixed now @j-a-m-l.

j-a-m-l:
@dated sorting preferences are stored, but not the activated filters.

dated:
Right. I'll have a look at those later today.

dated:
I'd exclude the search query from the stored filters, do you agree @j-a-m-l?

j-a-m-l:
Yes, apart from that, it's necessary to highlight more that filters are active, so when a user backs after several days he could recognize that.

dated:
I'll add a little notice to the dashboard sidebar to indicate that filters are active, as currently there is no such warning. Other than that there is a noticeable delay currently when changing the sort order or applying filters, as the whole profile gets updated. If possible that should be refactored as to update only the changed values and thus increase performance.

j-a-m-l:
@dated do you consider this PR as WIP or finished?

dated:
I'll take a look at it tonight, i think there have been some design changes in the meantime, i haven't had a chance to check yet.

dated:
One thing that's missing is a notification on the dashboard, as there is no indicator of wether there are any active filters.

The best solution would be to display a respective notification on a per session basis imho, with the option to never display it again. Thoughts @alexbarnsley / @j-a-m-l?

dated:
Something like so:

image

@olejegcord what do you think?

olejegcord:
Maybe a good idea, but there are a questions. How will the user understand which filters we are talking about?  I don't know which filters we're talking about, because this block should display all the wallets 🤔

dated:
Currently the sidebar on the dashboard shows the filtered wallets, not the whole list. The filters can be applied on the wallets page. As this warning shows only when the user has activated one or more filters, he will likely understand what is meant by that.

dated:
The wording can be adjusted of course, this draft is more about the design and the concept behind it :)

olejegcord:
I think it's best to do a pop-up notification in the right corner. Since the sidebar will change in the future. 
In the notification, place your action buttons 
I draw, give me 15 min

olejegcord:
like this 
image

zillionn:
Btw it's kind of strange that there is such filter on the right panel when inside some wallet, but there is none on the actual `My Wallets` page.

And about the active filters. It's common to change the color/appearance of the filter icon when active. You could change its color, add small checkbox above it or the number of active filters:
newm

dated:
The filter icon becomes red if a filter has been activated actually. I don't like the popup too much as it is way to prominent imho and it enforces an interaction with the user - it can't be ignored. I'm trying some different ideas with the simplest being a floating icon with a tooltip on the pages where the filters cannot be modified.

zillionn:
You could just add a one line message above the wallets, something like this:
Screenshot 2019-05-17 at 10 54 04

olejegcord:
Stop. If we're talking about this filter http://prntscr.com/npqv3r (this filter only touches the inside of the wallet.)
it's not like he shouldn't be influencing this block http://prntscr.com/npqw4m

dated:
That settles it, I'll refactor it accordingly.

PR #1158, opened 2019-03-27
chore: add `prettier` to format code

j-a-m-l:
## Proposed changes
Any option or suggestion?

## Types of changes

## Checklist


j-a-m-l:
Some rules aren't configurable and they don't follow Vue recommendations, so, instead of ignoring those recommendations, it's probably better to keep or current toolset.

PR #1171, opened 2019-04-08
feat(plugins): implement ability to overwrite elements with slots

luciorubeens:
## Proposed changes

- Registers a global component to be used by plugins
- Updated the branch

## Types of changes


j-a-m-l:
@luciorubeens could you separate the other changes from the PR?

It would be better to follow a different approach to keep `v2.4.0` updated, specially since the current `develop` branch is broken (the problem with settings, etc.) and it would propagate the problems to the other branch.

PR #1172, opened 2019-04-09
fix: other network fees when higher than ARK static fees

alexbarnsley:
## Proposed changes

Resolves an issue with non-ARK chains, where the fee greater than ARK's static fees resulted in not being able to send transactions.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/services/client.js
+++ b/src/renderer/services/client.js
@@ -672,7 +672,6 @@ export default class ClientService {
    */
   async buildTransfer ({ amount, fee, recipientId, vendorField, passphrase, secondPassphrase, wif }, isAdvancedFee = false, returnObject = false) {
     const staticFee = store.getters['transaction/staticFee'](0) || V1.fees[0]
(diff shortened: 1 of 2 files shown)


src/renderer/services/client.js
```diff
@@ -669,9 +671,10 @@ export default class ClientService {
    * @returns {Object}
    */
   async buildTransfer ({ amount, fee, recipientId, vendorField, passphrase, secondPassphrase, wif }, isAdvancedFee = false, returnObject = false) {
+    const staticFee = store.getters['transaction/staticFee'](0) || V1.fees[0]
     // To ensure that transfers cannot be build with a bigger fee than V1
```
j-a-m-l:
```suggestion
    // To ensure that transfers cannot be built with bigger fees than the static fees allowed by the peer or V1
```

or remove the comment if you think that is unnecessary

alexbarnsley: I will remove as it's only in the transfer method, thanks 👌

PR #1174, opened 2019-04-09
fix: `InputSelect` dropdowns

j-a-m-l:
## Proposed changes
Some dropdowns were not working since the upgrade of dependencies. The component that was affected was used on:
 - Welcome screen: choose app language
 - New profile: all dropdowns
 - Send transfer: select address from wallets and contacts

## Types of changes

## Checklist


j-a-m-l:
@luciorubeens I've fixed that.

A similar problem was reported here: https://github.com/vuejs/vue/issues/9836

PR #1180, opened 2019-04-10
feat(plugins): implement ability to overwrite elements with slots

luciorubeens:
## Proposed changes

Register a global component to be used by plugins

## Types of changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ npm-debug.log
 npm-debug.log.*
 
 __tests__/*/.coverage
+
+plugins/*
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "electron-log": "^2.2.17",
     "electron-window-state": "^5.0.3",
     "emittery": "^0.4.1",
+    "fs-extra": "^7.0.1",
     "jsonschema": "^1.2.4",
     "localforage": "^1.7.3",
     "lodash": "^4.17.11",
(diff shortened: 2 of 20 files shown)


alexbarnsley:
@luciorubeens is there a particular plugin you have available that can be used to test this with? Thanks

luciorubeens:
@alexbarnsley Here you go: https://github.com/luciorubeens/ark-desktop-custom-chain. Currently have to access the route to load, but can be improved next.

alexbarnsley:
Looks great! How do you think it could work without having to be on a specific page? e.g. I enable this plugin and the footer is always "My custom chain" while on that profile?

Also, I'm getting a warning about portal-target not being PascalCase:

image

luciorubeens:
Yeh, I thought about loading at startup, does not it look good?

`PascalCase` problem fixed.

j-a-m-l:
@luciorubeens I just tried it, and it works, but, maybe it would be better to rename the `PortalTarget` to "plugin-footer", but continue exposing it as "footer", so the code of the wallet differentiates clearly which portals are intended to be used internally and which others are there for plugins.

luciorubeens:
Agreed. Done! @j-a-m-l

alexbarnsley:
@j-a-m-l yes, it needed PluginWrapper to point to the updated portal name. I've made the change if you want to test it.

PR #1191, opened 2019-04-16
feat(plugins): require permissions to include content on the app

j-a-m-l:
## Proposed changes
Currently permissions are required only to access some elements of the app inside the plugin, but users can't block plugins that introduce other elements on the app or modify them.

## Types of changes

## Checklist


src/renderer/services/plugin-manager.js
```diff
@@ -67,11 +68,23 @@ class PluginManager {
       profileId
     })
 
-    await this.loadComponents(pluginObject, plugin)
-    await this.loadRoutes(pluginObject, plugin)
-    await this.loadMenuItems(pluginObject, plugin, profileId)
-    await this.loadAvatars(pluginObject, plugin, profileId)
-    await this.loadWalletTabs(pluginObject, plugin, profileId)
+    const permissions = zipObject(plugin.config.permissions, fill(new Array(plugin.config.permissions.length), true))
```
alexbarnsley: This feels like a long-winded way of building one object when the object already exists. Why not just 

```
if (plugin.config.permissions.includes('COMPONENTS')) {

}
```

If you're insistent on doing the zip stuff, do it when first loading the JSON so the "zipping" isn't done every time the plugin is enabled

j-a-m-l: Using includes would iterate over the array on every check, that's why I'm building an object with the available permissions, to iterate over the array only once.

I though that the correct place to check permissions, was when enabling the plugin, @alexbarnsley, where would you do it?

alexbarnsley: I agree that the correct place to check for permissions is when enabling. But I think the `zipObject` response needs to be stored on the plugin object itself so it doesn't get called every time the plugin is enabled. Possibly here: https://github.com/ArkEcosystem/desktop-wallet/blob/9b0401831143f13c5ada0499eac63106162a5d90/src/renderer/services/plugin-manager.js#L459

```
    const permissions = zipObject(config.permissions, fill(new Array(plugin.config.permissions.length), true))

    config = {
      ...,
      permissions,
      ...
    }
```

j-a-m-l: I'm not sure about that, because it's more space-efficient using an Array to store that data, than storing an Object, which values are always `true`.

Using `zipObject` each time a user enables a plugin, something that it's not going to be very common, is not expensive enough to justify using that additional space.

The only use of that line is to avoid searching later, on each permission, if it can load data or not; as an alternative implementation, I could have used `_.difference` to obtain only the permissions that apply and then use the `loadX` method.

src/renderer/services/plugin-manager.js
```diff
@@ -67,11 +68,11 @@ class PluginManager {
       profileId
     })
 
-    await this.loadComponents(pluginObject, plugin)
-    await this.loadRoutes(pluginObject, plugin)
-    await this.loadMenuItems(pluginObject, plugin, profileId)
-    await this.loadAvatars(pluginObject, plugin, profileId)
-    await this.loadWalletTabs(pluginObject, plugin, profileId)
+    // COMPONENTS, ROUTES, MENU_ITEMS, AVATARS, WALLET_TABS
+    for (const permission of uniq(plugin.config.permissions)) {
```
luciorubeens: I think it needs a validation before calling the method:

```suggestion
    for (const permission of ['COMPONENTS, ROUTES, MENU_ITEMS, AVATARS, WALLET_TABS']) {
    	if (plugin.config.permission[permission]) {
    		// ...
    	}
    }
```

j-a-m-l: In this case, `permissions` is an Array, so it wouldn't be necessary to do that. The only possible problem that it could cause it's that some permissions should be loaded in order. I'll change it now.

luciorubeens: You are considering that all plugins will write the permissions correctly.

j-a-m-l: It's an array of permissions, if someone uses numbers, `null` or an object, it wouldn't work and it would throw an Error, so they could realize that it's a mistake.

alexbarnsley:
This is much more complicated than it needs to be. 4 packages from lodash is unnecessary.

No packages are needed, you can just `reduce` the array. I'm willing to settle for a variation of your first commit:

```
if (!plugin.config.permissions || !plugin.config.permissions.length) {
  return
}

const permissions = plugin.config.permissions.reduce((map, value) => {
  map[value] = true

  return map
}, {})

if (permissions.COMPONENTS) {
  await this.loadComponents(pluginObject, plugin)
}
if (permissions.ROUTES) {
  await this.loadRoutes(pluginObject, plugin)
}
if (permissions.MENU_ITEMS) {
  await this.loadMenuItems(pluginObject, plugin, profileId)
}
if (permissions.AVATARS) {
  await this.loadAvatars(pluginObject, plugin, profileId)
}
if (permissions.WALLET_TABS) {
  await this.loadWalletTabs(pluginObject, plugin, profileId)
}
```

In fact, the array could still be built in the `fetchPlugin` method since the same data is already being stored. Meaning it doesn't need doing every time a plugin is enabled.

j-a-m-l:
@alexbarnsley more permissions are going to be added, some of them depending on each other, so a way to avoid those duplicated check + load is loading first the essential ones.

j-a-m-l:
We're going to add potentially more permissions, @alexbarnsley. As an example:

 - `THEMES`: https://github.com/ArkEcosystem/desktop-wallet/pull/1194
 - `IFRAME_URL`: to allow inserting iframes that loads URLs (not all pages could be used on inframes: https://github.com/ArkEcosystem/desktop-wallet/pull/1192)
 - `BACKGROUNDS`: to add new backgrounds
 - `IDENTICONS`: to use other kind of identicons
 - `WALLET`: to read information about a specific wallet
 - `VOTES`: to read information about the votes of a wallet
 - `TRANSACTIONS`: to access the transactions of a wallet
 - `TRANSFER`: to send
 - `VOTE`: to vote
 - `I18N`: to use different messaages for the same language
 - `DASHBOARD`: to add content to the dashboard
 - `FOOTER`: to add content to the footer
 - `IMAGE`: to use other images (on the welcome screen, logo or other places)
 - `FEED`: to use a different source of announcements

Do you really think that the best solution is duplicating manually the block?
```js
if (permissions.X) {
  await this.loadX(pluginObject, plugin, profileId)
}
```
That approach also would check all the possible permissions on plugins that only require a subset of them or even none.

alexbarnsley:
@j-a-m-l I prefer verbose over the hard-to-read code and the use of 4 different lodash packages.

j-a-m-l:
We are already using `lodash`, so what's the problem on that?

What parts do you think that are hard to read? I could refactored them.

Apart from that, checking every permission on every plugin is unnecessary. In this case, we only check the permissions that are required by the plugin.

alexbarnsley:
The fact we load lodash as one package is a problem. it's a big package which is unnecessary to load everything since we don't use the majority of it. And by using 4 different ones seems unnecessary in itself for such a small task.

If you're insistent on a loop, just get it to loop through the permissions and run the load if it exists. Similar to how Lucio suggested to do it:

```js
import { camelCase } from 'lodash'

...

for (const permission of plugin.config.permissions) {
  const method = camelCase(`LOAD_PERMISSION_${permission}`)
  if (this.hasOwnProperty(method)) {
    this[method](pluginObject, plugin, profileId)
  }
}
```

and rename all load methods to `loadPermissionX`

j-a-m-l:
Webpack tree-shakes the modules (https://webpack.js.org/guides/tree-shaking/), so it should include only the code that is actually used. Those functions don't change anything since all of them are already being used on other parts. Anyway, we could use ```camelCase(`LOAD_PERMISSION_${permission}`)``` to avoid `upperFirst`.

As I explained on https://github.com/ArkEcosystem/desktop-wallet/pull/1191#discussion_r276609433 that approach would not fail if the user is trying to load permissions that don't exist, so it would make it harder to discover why the plugin don't work (a typo like `COMPONENT` instead of `COMPONENTS` would be harder to spot). An alternative is checking the existence and throw an `Error` in case the method doesn't exist.

It also wouldn't work with permissions that depends on loading others before (basically `COMPONENTS` and `ROUTE`).

I think that renaming those methods to `loadPermissionX` is a good idea to differentiate them from other methods, but it could be misleading because those methods load components, routes, etc. not their permissions. Instead of that, since, only `loadSandbox` uses the `loadX` form, I'd rename it to `createSandbox`, `prepareSandbox` or something like that.

alexbarnsley:
Fair enough about the tree shaking, I wasn't aware of that. And accept the point of loading certain permissions prior to others. I still think verbose would be perfectly fine here, instead of making it overly complicated.

Regardless, I've asked @luciorubeens to take a look since we clearly disagree with each other. I am happy with whichever way Lucio goes so we can get this PR merged.

PR #1194, opened 2019-04-18
feat(plugins): allow adding custom themes

j-a-m-l:
## Proposed changes
This PR includes a way to load and select new UI themes.

This PR depends on https://github.com/ArkEcosystem/desktop-wallet/pull/1191 to deal with permissions

## Types of changes

## Checklist


alexbarnsley:
@j-a-m-l do you have an example plugin I can test this with?

j-a-m-l:
@alexbarnsley https://github.com/j-a-m-l/ark-desktop-wallet-dark-contrast

alexbarnsley:
I like this 👍

src/renderer/services/plugin-manager.js
```diff
@@ -368,6 +369,38 @@ class PluginManager {
...
+      return
+    }
+
+    const pluginThemes = this.normalize(pluginObject.getThemes())
+    if (pluginThemes && isObject(pluginThemes)) {
+      // Validate the configuration of each theme and ensure that their CSS exist
+      const themes = Object.keys(pluginThemes).reduce((valid, themeName) => {
+        const config = pluginThemes[themeName]
+
+        if (isBoolean(config.darkMode) && isString(config.cssPath)) {
```
luciorubeens: It will only be imported if I set `darkMode` to `true`. I did not get it, this property is useful for what?

src/renderer/App.vue
```diff
@@ -345,6 +359,21 @@ export default {
...
+     * Webpack cannot require assets without knowing the path or, at least, part of it
+     * (https://webpack.js.org/guides/dependency-management/#require-context), so,
+     * instead of that, those assets are loaded manually and then injected directly on the DOM.
+     */
+    applyPluginTheme (themeName) {
+      if (themeName && this.pluginThemes) {
+        const theme = this.pluginThemes[themeName]
+        if (theme) {
+          const $style = document.querySelector('style[name=plugins]')
+          $style.innerHTML = fs.readFileSync(theme.cssPath)
```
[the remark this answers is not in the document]
j-a-m-l: Probably not, but we would need a way to pre-process the CSS first. I could try something.

src/renderer/services/plugin-manager.js
j-a-m-l: Nop, @luciorubeens, it only checks that `config.darkMode` is boolean. That property is used could be used to allow a light and a dark variant of each themes, but currently it's used to decide which style should use Changelly.

j-a-m-l:
@luciorubeens I've made several changes.

About overwriting user styles, it can happen everywhere, as the security testing plugin demonstrated. We don't have a good solution for that. At most we could try to remove rules before/after injecting the CSS, but it wouldn't matter, since users could include them on HTML or using new tricks.

PR #1204, opened 2019-04-24
fix: don't close custom peer modal when clicking inside

j-a-m-l:
## Proposed changes
After opening the modal to add a custom peer, it's impossible to use it, because when clicking inside it, it's automatically closed.

## Types of changes

## Checklist


alexbarnsley:
I think I tested this a few days ago but I've slept since. Just to confirm, this is when clicking the overlay to close the modal, it should keep the peer settings menu open in the sidebar? @j-a-m-l

j-a-m-l:
@alexbarnsley I've updated the PR description.

alexbarnsley:
🤔 i didn't experience that. I will re-test, probably after the next release

dated:
This is more of a problem with the click-outside directive imho, as the problem is present elsewhere as well (`settings -> reset data` for instance). The current behaviour also stops a menu from closing after re-clicking on the button which opens it.

dated:
Well, the easy fix for the reset data modal is an extra condition to emitting the close signal:

```
if (this.outsideClick && !this.isResetDataModalOpen) {
  this.$emit('close')
}
```

I added that in #1146; the second issue mentioned is still standing.

PR #1206, opened 2019-04-24
fix: fetch fees only if the network is available

luciorubeens:
## Proposed changes

When creating the initial profile the network has not yet been defined.

## Types of changes

### DIFF
--- a/src/renderer/store/modules/network.js
+++ b/src/renderer/store/modules/network.js
@@ -77,7 +77,7 @@ export default new BaseModule(NetworkModel, {
         network = rootGetters['session/network']
       }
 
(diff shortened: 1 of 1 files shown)


j-a-m-l:
@luciorubeens maybe it'd be better to check it on the initialization of the app or when that method is called before having created the first profile. That way, it would fix that problem, but it would not fail when calling that method without a network too, which could be useful for finding errors.

luciorubeens:
This method is called by the `synchronizer`: https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/renderer/services/synchronizer.js#L239-L241. Don't think this service should handle these specific cases.

PR #1210, opened 2019-04-26
fix: network model validation to accept numbers in slip44

luciorubeens:
## Proposed changes

When registering a custom network, an invalid type error is issued.

## Types of changes


alexbarnsley:
I've approved but i don't get this error on the `next` branch 🤔 how was you adding the network @luciorubeens?

PR #1215, opened 2019-04-30
fix: do not allow that the last profile is removed

j-a-m-l:
## Proposed changes
Allowing that users remove all profiles provokes lots of inconsistencies on different parts of the application; since users are forced to have 1 profile at least, detecting and solving those inconsistencies and side effects is more problematic that this approach.

What do you think @alexbarnsley and @luciorubeens?

## Types of changes

## Checklist


alexbarnsley:
Perhaps a future addition can be to say that a minimum of 1 profile is required. Or to force showing of the initial profile create screen that shows on first launch. However, this can wait and will be done on the next design update from Oleg.

PR #1218, opened 2019-05-02
fix: username validation when registering a new delegate

j-a-m-l:
## Proposed changes
The current `@arkecosystem/crypto` package (2.2.2) has removed the username validator, which was breaking the delegate registration process.
Following @faustbrian suggestion, instead of upgrading that package, since it'is  going to be modified until `core` 2.6, I've, instead, implemented the validation on the wallet.

Additionally, the confirmation screen has been fixed to display the sender label correctly.

## Types of changes

## Checklist

### DIFF
--- a/src/renderer/i18n/locales/en-US.js
+++ b/src/renderer/i18n/locales/en-US.js
@@ -923,7 +923,7 @@ export default {
     UNVOTE: 'Unvote',
     VOTES: 'Votes',
     VOTE: 'Vote',
-    USERNAME_EMPTY_ERROR: 'The username must hava at least 1 character',
+    USERNAME_EMPTY_ERROR: 'The username must have at least 1 character',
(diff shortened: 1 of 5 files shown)


src/renderer/i18n/locales/en-US.js
```diff
@@ -923,6 +923,7 @@ export default {
     UNVOTE: 'Unvote',
     VOTES: 'Votes',
     VOTE: 'Vote',
+    USERNAME_EMPTY_ERROR: 'The username must hava at least 1 character',
```
luciorubeens: Just a typo:

```suggestion
    USERNAME_EMPTY_ERROR: 'The username must have at least 1 character',
```

j-a-m-l: Thanks, @luciorubeens

PR #1219, opened 2019-05-02
fix: show the market chart when enabling it on the profile

j-a-m-l:
## Proposed changes
The market chart wasn't properly enabled or disabled when the related setting was changed on the profile edition.

## Types of changes
- [X] Bugfix (non-breaking change which fixes an issue)

## Checklist
- [X] I have read the [CONTRIBUTING](https://docs.ark.io/guidebook/contribution-guidelines/contributing.html) documentation
- [X] Lint and unit tests pass locally with my changes

### DIFF
--- a/src/renderer/components/MarketChart/MarketChart.vue
+++ b/src/renderer/components/MarketChart/MarketChart.vue
@@ -2,15 +2,16 @@
   <section class="MarketChart w-full flex-column">
     <slot />
     <LineChart
+      v-if="isActive"
       v-show="isReady"
       ref="chart"
       :chart-data="chartData"
       :options="options"
       :height="315"
-      @ready="show"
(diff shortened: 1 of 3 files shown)


alexbarnsley:
What exactly is the issue here? I can enable and disable the chart on the current `next` branch, and the setting saves between reloads for my mainnet profile. Can you explain the steps to recreate the issue this is fixing please?

j-a-m-l:
@alexbarnsley the menu to enable and disable the market chart was moved to the profile edition. On `next` it should not exist on the sidebar menu.

The problem occurs when you edit your profile to enable the chart, but then, the chart area is visible on the dashboard, but it's empty. It's related to how the `chart.js` library works: https://github.com/apertureless/vue-chartjs/issues/372

PR #1229, opened 2019-05-09
refactor: require user action when generating 2nd passphrase

### LINKED ISSUE #1221: Generate 2nd Passphrase New Passphrase Alert
1. If the user presses `second passphrase` button to generate a second passphrase, it shows a passphrase for them that they may write down or refresh and write down.

2. The user then proceeds to transaction completion page.

3. If the transaction fails and produces an error of some kind, the user must then begin the process again.

4. The user *may not realize* that the new second passphrase recommendation is different than their previous attempt, and so they may quickly go to the next screen and *not write down the new passphrase that was generated, discarding the old.*

5. Then the user will be locked out of their funds if that second transaction succeeds.

Solution: whenever user presses `second passphrase` to generate a second passphrase, show modal "A brand new second passphrase will now generate for this address. Please make a note of the new second passphrase and disregard any previous second passphrases you may have generated for this address." *OK*

Source: I generated a second passphrase for my arkstickers address a while back, and then recently came across this thought, and tested to show that multiple attempts to generate 2nd passphrase results in a different 2nd passphrase each time. As second passphrase users may be handling massive funds, we must warn the user to avoid a potential disaster

dated:
## Proposed changes

Implements and closes #1221.

## Types of changes

## Checklist


doubled1c3:
@dated something happened, it got merged but I don't see this feature as shown in desktop wallet 2.4.1? @j-a-m-l

alexbarnsley:
@doubled1c3 2.4.1 was only a minor patch to fix the sending of transactions due to a core issue at the time.

PR #1244, opened 2019-05-13
[WIP] feat: add option to export wallets

dated:
## Proposed changes

Adds the option to export ones wallets.

## Types of changes

## Checklist


dated:
any feedback is appreciated @alexbarnsley @j-a-m-l @ItsANameToo

dated:
I'll add a notice to the modal regarding the passphrase, for the import a 'Batch Import' button could be added to the import page?

j-a-m-l:
I'm not sure about that option; we could rename the buttons to "Import Wallets", but that page would need a modal to import the wallets and it would break the step style and behaviour of the process, so it wouldn't fit there smoothly.

I'd rather add other button. Maybe "Batch Import", near the "Import Wallet" button is clear enough.

zillionn:
I guess this will be a nice feature when you're switching computers or reinstalling the OS, right? If yes, how about to export the whole `Profile` with its settings and everything? Then when you install the desktop wallet on the new comp, you'll have an option to import the profile.

j-a-m-l:
Yes, exporting the whole profile could be a feature. In that case it would include data to import the network in case it doesn't exist yet.

dated:
I added the following warning to the modal:

>Your exported wallets will not contain your passphrases, only the addresses and respective names will be saved!

j-a-m-l:
It's ignoring Ledger wallets, even when they have a custom name.

dated:
Does this work for you @j-a-m-l? I added an additional option for the ledger wallets.

j-a-m-l:
@dated it works OK, but the Ledger switch should be visible or enabled only when there is a Ledger connected.

Apart from that, I'd change the format to use an object as the value instead of the custom name of the wallet, so the format is more flexible:
```
  "address": {
    "name": "my wallet"
  }
```

Related to that, I'd export other data per wallet that could be useful for other purposes (trading, taxes, etc.), such as:
 - public key
 - delegate username
 - voted delegate
 - current balance
 - current balance in the configured currency (BTC, EUR, etc.).
 - number of transactions

@alexbarnsley @luciorubeens @boldninja what do you think?

zillionn:
IMO there should be an option to select/deselect every object before exporting. Maybe something like this, with checkboxes:

dated:
The ledger option is now disabled if no ledger is connected and I've added some more fields for now:

```
  "wallets": [
    {
      "ANoNeeFDFo61XcQBYad6rEbPvfATwgjTNT": {
        "name": "test",
        "publicKey": "02ead78d2ecd00a966f45c1ea3fb8d2e032dcb704a451a1af0426789261d37459a",
        "balance": {
          "ARK": "1780.56728118",
          "USD": "1105.27"
        }
      }
    }
  ]
```

I'll try to add the remaining fields and some tests tomorrow.

dated:
I would refrain from the transaction counts for now as the API does return wrong values, and I don't see much use for that information anyway.

The other fields have been added. I removed the address as key and added it as a property. Are you ok with the current format @j-a-m-l?

src/renderer/components/Modal/ModalExportWallets.vue
```diff
@@ -0,0 +1,265 @@
...
+          active: false,
+          filter: el => el.name && el.name.length
+        },
+        excludeEmpty: {
+          active: false,
+          filter: el => el.balance
+        },
+        excludeLedger: {
+          active: false,
+          isDisabled: !this.isLedgerConnected,
```
luciorubeens: The `data` can not access the `computed` properties, you need to use the `watch`:

```js
watch: {
  isLedgerConnected: {
    handler(val) {
      this.options.excludeLedger.isDisabled = !val
    },
    immediate: true
  }
}
```

dated:
Thanks for the review @luciorubeens, I'll get the changes done asap.

dated:
All fixed, thanks again @luciorubeens.

PR #1255, opened 2019-05-20
fix: plugin table sorting

dated:
## Proposed changes

Fixes the sorting on the plugin table which was overwriting the sorting of the contacts.

## Types of changes

## Checklist


dated:
It should work now as expected @j-a-m-l

j-a-m-l:
It looks like the status column doesn't work.

dated:
It looks like when the plugins are first loaded, the `isEnabled` property is not set, which caused the sorting to fail. I fixed this on the sorting-side only now.

dated:
I'm getting some strange errors, will wrap this up tonight.

PR #1257, opened 2019-05-20
feat(plugins): warn about using the plugin system

j-a-m-l:
## Proposed changes
When the user tries to navigate to the plugins section, a modal would surge with a warning. If the user decides to enable them, that modal doesn't appear again.

## Types of changes

## Checklist


src/renderer/models/profile.js
```diff
@@ -1,3 +1,4 @@
+import isBoolean from 'lodash'
```
luciorubeens:
```suggestion
import { isBoolean } from 'lodash'
```

Issue #894, opened 2018-12-30
Introduce Local Caching of Transaction Lists


doubled1c3:
**Problem**

It seems like whenever I access a wallet to see it inside the wallet, the triple pulsing blue dots happen. Then when I go to a different screen and go back to the wallet in question, the dots immediately happen again. Are we calling transactions from peers every time we look at a transaction list? This is best case an annoyance as sometimes I have to wait 3-30 seconds every time I click a wallet, and sometimes it doesn't even load... worst case, we are congesting the peers who keep giving us the same transaction lists over and over. This is taking away from the fluidity of the user experience.

**Solution**

As it is a desktop wallet, we should be able to achieve instant viewing of transaction lists as we move between different addresses, through a local cache. Then new transactions should be pushed to the wallet via webhooks or manual refresh button. The wallet usage experience can become lightning fast and peers will be destrained.

Furthermore, when you are importing a wallet address, should have the option to download all transactions, which will result in a progress bar or log showing activity as the transactions are downloaded into the cache. After that point the peer is not needed again, unless a webhook event occurs or manual refresh button is pressed

Are there any factors regarding local caching that I'm missing?

zillionn:
I fully agree with you. What's the point of downloading all txs over and over again, they can't change, it's a BLOCKCHAIN! 🤣 

IMO you should pull only the newest txs, using the timestamp of the latest cached:
http://95.179.134.209:4003/api/v2/wallets/DFWTC6qyGgWPFp3NvWtmtvZV1mKF9ASVMk/transactions?timestamp.from=123123123

Webhooks are turned off on most peers.

doubled1c3:
Copy that. Yeah, let's get it implemented to achieve a big leap in user experience!

JeremiGendron:
@zillionn what's the correct API call for timestamp from? The above doesn't work.

zillionn:
I'm afraid it's not possible yet. Right @faustbrian ?

faustbrian:
https://github.com/ArkEcosystem/core/blob/master/packages/core-api/lib/versions/2/schema/transactions.js#L122-L129

zillionn:
I saw the schema, but does it really work? I don't see anywhere the code where it actually searches by timestamp.

dated:
Timestamp is supported by the `search` endpoint only.

faustbrian:
[...]

PR #1269, opened 2019-05-30
feat: cache wallet page transactions

j-a-m-l:
This PR introduces a simplistic, but effective, way to cache transactions (resolves https://github.com/ArkEcosystem/desktop-wallet/issues/894).

- [ ] Bugfix
- [x] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing


j-a-m-l:
uhmm, @luciorubeens could you try to capture under which circumstances do you see that banner? I've never see it (it works well when I sen a transfer).

About storing only the last transactions and only store wallets / contacts, my idea was making navigation smoother, so users don't have to wait to navigate to / from pages that have been already visited. That specially useful under bad network conditions due user connection or due using a slow peer.

About the expiration date, what do you propose? flushing those old transactions at the beginning? checking the date every time that they are accessed?

zillionn:
IMO you should cache all txs under one `key` (wallet address or public key) and use that cached object to show/order/paginate the txs. So if I go on the next page, you should check the cached object if it contains 50 txs older than oldest on the current page, otherwise query the API. You can use `/api/transactions/search` to retrieve 50 txs from timestamp < oldest tx on the current page.

And btw ordering of txs just doesn't work.

Screenshot 2019-06-02 at 18 52 15

Ordering by `Sender`, `Recipient` and `Smartbridge`:
Screenshot 2019-06-02 at 18 53 48

By `Date`:
page 1: 
Screenshot 2019-06-02 at 18 56 41
page 2:
Screenshot 2019-06-02 at 18 56 48

IMO you should only order the txs on the current page on the fly, same way the explorer is ordering them, without calling the API.

j-a-m-l:
Thanks, @zillionn. I'll fix the ordering on this PR.

About sorting the transactions of the current page only, we can't do it because we'd lose a lot of functionality: for instance, users would need to navigate between lots of pages to see the first one, instead of just ordering them ascendantely. In the explorer that use cases are not as important.

And about aggregating transactions by address (instead of using the address, but the pagination too) the problems with that approach are:
 - when you visit an address with thousands of transactions, you've to download all of them to sort them
 - depending on the order parameter, you'd retrieve different sets of transactions, so you wouldn't be able to know how to sort them with other order parameters or which other not-yet-received transactions should be included

Keep in mind that this PR focus on making the navigation faster, not on having a copy of the entire data of an address. That could be a waste of space in several scenarios.

zillionn:
I guess you don't understand me, do it as you decide.

j-a-m-l:
I've re-read it and, @zillionn, your approach looks like a good one until you start thinking about edge cases, but if you think that the current implementation could be improved, I'd love to hear your feedback.

dated:
i'd go so far as to say that remote sorting is the way to go, even on the explorer. i'll speak to @itsanametoo about it when i'm back.

j-a-m-l:
Caching the last 50 would cover only 5 pages, so it could be useful, but it's limited.

About the refresh button, using the search endpoint to fetch only the latest could be a good optimization in some cases. Probably we could keep this PR approach and, later, dig into that other kind of enhancements.

j-a-m-l:
The sorting problem is produced by core: https://github.com/ArkEcosystem/core/issues/2649

ItsANameToo:
I get the same banner as @luciorubeens had when I:

1) Open a wallet
2) Go to `next` transaction screen
3) Go `back` to previous screen
4) Screen will now have that "10 new transactions" banner

Regarding the sorting, I think that remote sorting would be nice in the wallet as it is always limited to a specific address (unlike the explorer where you also have all transactions on the network as an option). Right now sorting is  all over the place as amounts work, but sender / receiver don't seem to work or at least not spanned over multiple pages of transactions. However, I think that's out of scope for this PR

faustbrian:
@alexbarnsley what's the plan for this PR? Has been open for quite some time without any activity from the author.

alexbarnsley:
I'll leave open for now and try to fix the issues raised when I get some free time

PR #1386, opened 2019-08-16
refactor: improve passphrase detection in vendorfield

dated:
This PR improves the detection of passphrases in the vendorfield of the transaction form. The current implementation fails to detect a passphrase if it starts or ends with a space for example.

- [ ] Bugfix
- [ ] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)


alexbarnsley:
I think it should be the case that it shows as invalid if there is a space at the start/end. E.g. someone accidentally adds a space. With this change, it would highlight the passphrase as correct, but when importing it would import a different wallet because that space will be used to determine the address. it could also be the case that I want spaces or capitals in my passphrase so it's not a bip39 passphrase. or did I misunderstand the PR? 🤔

dated:
Hmm i will re-check the implications of these changes. It was intended only for the warning that shows up when a the vendorfield contains a bip39 passphrase.

alexbarnsley:
Ah vendorfield. My mistake, I thought it was importing a wallet 👌

dated:
Yea, the passphrase input is unaffected.

alexbarnsley:
2 things:

1 - I think the warning on the bip39 transfer modal has stopped working. I created a paper wallet, imported it (recognised as bip39), went to send a tx with it in the vendorfield and the field just shows as orange. I suspect the "x/255 remaining" is overriding it.

2 - would you mind creating a quick unit test for this, to test for both with and without spaces please

alexbarnsley:
image

dated:
will do

zillionn:
Have you thought about hiding the `Smartbridge` and `Transaction fee` fields and add a switch `Advanced settings` which will show them? 99% of the times users will not touch those 2 fields.

PR #1437, opened 2019-09-04
feat(aip-11): v2 transactions with nonces

alexbarnsley:
Waiting on the crypto package to be updated to 2.6.

- [ ] Bugfix
- [x] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [ ] It's submitted to the `develop` branch, _not_ the `master` branch
- [ ] All tests are passing


src/renderer/store/modules/network.js
```diff
@@ -69,6 +69,27 @@ export default new BaseModule(NetworkModel, {
...
+        network = rootGetters['session/network']
+      }
+
+      try {
+        const crypto = await Client.fetchNetworkCrypto(network.server)
+        const { constants } = await Client.fetchNetworkConfig(network.server)
+        commit('UPDATE', {
+          ...network,
+          crypto,
+          milestone: constants
```
luciorubeens: The network already has the `constants` field, it can only be updated instead of duplicating the data.

src/renderer/services/synchronizer/network.js
```diff
@@ -0,0 +1,3 @@
+export default async synchronizer => {
+  await synchronizer.$store.dispatch('network/updateData')
```
luciorubeens: Do we need to update this data often?

alexbarnsley: I think this is a yes and no answer. No, because of there being very little milestones. Yes, because when we are at a milestone we could do with knowing about it fairly promptly. The alternative is to advise that users need to reload the app for the wallet to acknowledge a milestone has been reached. Which do you think?

src/renderer/store/modules/network.js
alexbarnsley: Updated this @luciorubeens 	👌

src/renderer/services/synchronizer/network.js
luciorubeens: Ah, got it. Yeh, since the only use will be this milestone, I think we should advise users to reload.

alexbarnsley: I've removed the auto-sync @luciorubeens

PR #1438, opened 2019-09-04
refactor: hide delegate tab if wallet is not on profile

dated:
1. explicitly set `cursor-pointer` on anchors in the wallet address component (the pointer is missing on the transaction show modal for instance)
2. hide the delegate tab if the current wallet is not owned

- [x] Bugfix
- [ ] Feature
- [x] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing

- [ ] A convincing reason for adding this feature (to avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it)


alexbarnsley:
Perhaps make use of the vote status bar, hide the vote/unvote & search buttons

alexbarnsley:
Actually, that only shows on the delegate tab so I'm not sure where you would show it 🤔

dated:
Actually the status bar doesn't show for unowned wallets, so that information isn't necessarily obtainable from the first page of the delegate table either.

One could repurpose the dot that shows up in the wallet header when a wallet is not voting to show that a wallet is in fact voting, while displaying the delegate name on hover. But that might also require some modification to the red dot in the wallet table, which currently holds a conflicting type of information (as it shows up only if a voted delegate is out of the top X).

dated:
Opinions anyone? :D

alexbarnsley:
🤔 not sure why I added the label but didn't respond. maybe as you say, use the dot. Perhaps we show an orange dot for no vote, a green dot for a vote and show it for all wallets

dated:
:ok_hand: i'll try something in the next couple of days

alexbarnsley:
@dated did you get around to taking a look at this?

dated:
Nah, been busy with the plugin manager. I'll have a look this afternoon.

dated:
Instead of using colors (orange / green) I would propose to use an empty and filled circle instead.

alexbarnsley:
No worries! Yeh that sounds good 👌

src/renderer/components/Wallet/WalletHeading/WalletHeadingPrimaryActions.vue
```diff
@@ -114,11 +117,11 @@ export default {
   watch: {
     // Never show the not-voting icon until knowing if the wallet is voting or not
     'currentWallet.address' () {
-      this.showNotVoting = false
+      this.isVoting = false
```
luciorubeens: In all wallets the `isVoting` field remains false, perhaps we should change it to a computed field.

src/renderer/components/Wallet/WalletHeading/WalletHeadingPrimaryActions.vue
```diff
@@ -111,17 +110,10 @@ export default {
...
-    // Never show the not-voting icon until knowing if the wallet is voting or not
-    'currentWallet.address' () {
-      this.isVoting = false
     },
-    // To react to changes on the injected `walletVote` and changed not-voting icon immediately
-    'walletVote.username' () {
-      this.isVoting = !!this.walletVote.username
+
+    isVoting () {
+      return !!this.walletVote.username
```
luciorubeens: Hm, `walletVote` is not being updated when navigating between wallets, need to click the refresh button. It must be cleaned and updated when the `currentWallet` changes.

dated: I'll check/fix in a bit, thanks for pointing it out!

dated: I'm unable to reproduce. Switching between a voting and a non-voting wallet also changes the dot as intended. Am I missing something?

luciorubeens: Try browsing through non-imported wallets. Delegates tab > Open any delegate > Click on the address.

dated: Gotcha, i'll have a look ASAP.

dated: It worked when switching through the sidebar since that caused `fetchWalletVote` to be called. I changed it to call that method when `currentWallet` changes instead.

PR #1452, opened 2019-09-12
feat: timers in plugins

### LINKED ISSUE #1446: plugins: setTimeout and setInterval
It would be useful to allow plugins to access `setTimeout` and `setInterval`, for example, to fetch data from a 3rd party source either on delay or every x seconds. I can get around this with my own sources by using websockets that push data to connected sockets at regular intervals but that is not viable when accessing 3rd party sources out of the control of the plugin developer.

I think @alexbarnsley's concern is that this could lead to runaway code execution even after the plugin page has been closed as the execution could continue. I would suggest this could be alleviated by sandboxing the functions so their returned ID value is added to an internal array that acts like a stack so when the route is changed, `clearTimeout` and `clearInterval` are executed for all of the IDs in the array to clear them in order to prevent their execution once the plugin page has been closed.

alessiodf:
Resolves #1446 by implementing a `TIMERS` permission that exposes `clearInterval`, `clearTimeout`, `setInterval` and `setTimeout`.

It uses internal arrays to store the IDs of any timeouts and intervals to make sure they are all cleared once the route navigation changes and it also works with nested timers. A watchdog object is used to keep track of pending timeouts so they are immediately removed from the array once executed in order to prevent excessive memory usage that could potentially arise by repeatedly executing timeouts. The IDs are also removed from their respective arrays when `clearInterval` or `clearTimeout` are called. It also includes appropriate tests.

The reason for adding the feature was outlined in #1446.

- [ ] Bugfix
- [x] Feature
- [ ] Refactor
- [ ] Performance
- [ ] Tests
- [ ] Build
- [ ] Documentation
- [ ] Code style update
- [ ] Continuous Integration
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] All tests are passing
- [x] New/updated tests are included

### DIFF
--- a/src/renderer/services/plugin-manager.js
+++ b/src/renderer/services/plugin-manager.js
@@ -742,22 +742,20 @@ class PluginManager {
           return timerArrays.timeouts
         },
 
-        setInterval (...args) {
-          const functionArgs = args.splice(2)
+        setInterval (method, interval, ...args) {
           const id = setInterval(function () {
-            args[0](...functionArgs)
-          }, args[1])
+            method(...args)
+          }, interval)
           timerArrays.intervals.push(id)
           return id
         },
 
-        setTimeout (...args) {
-          const functionArgs = args.splice(2)
+        setTimeout (method, interval, ...args) {
           const id = setTimeout(function () {
-            args[0](...functionArgs)
-          }, args[1])
+            method(...args)
+          }, interval)
           timerArrays.timeouts.push(id)
-          timerArrays.timeoutWatchdog[id] = setTimeout(() => timers.clearTimeout(id), args[1])
+          timerArrays.timeoutWatchdog[id] = setTimeout(() => timers.clearTimeout(id), interval)
           return id
         }
       }
--- a/src/renderer/services/plugin-manager.js
+++ b/src/renderer/services/plugin-manager.js
@@ -743,13 +743,19 @@ class PluginManager {
         },
 
         setInterval (...args) {
-          const id = setInterval(...args)
+          const functionArgs = args.splice(2)
+          const id = setInterval(function () {
+            args[0](...functionArgs)
+          }, args[1])
           timerArrays.intervals.push(id)
           return id
         },
 
         setTimeout (...args) {
-          const id = setTimeout(...args)
+          const functionArgs = args.splice(2)
+          const id = setTimeout(function () {
+            args[0](...functionArgs)
+          }, args[1])
           timerArrays.timeouts.push(id)
           timerArrays.timeoutWatchdog[id] = setTimeout(() => timers.clearTimeout(id), args[1])
           return id
--- a/src/renderer/services/plugin-manager.js
+++ b/src/renderer/services/plugin-manager.js
@@ -714,6 +714,67 @@ class PluginManager {
       }
     }
 
+    if (config.permissions.includes('TIMERS')) {
+      const timerArrays = {
+        intervals: [],
+        timeouts: [],
+        timeoutWatchdog: {}
+      }
+
+      const timers = {
+        clearInterval (id) {
+          clearInterval(id)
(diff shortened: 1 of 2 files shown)


src/renderer/services/plugin-manager.js
```diff
@@ -714,6 +714,67 @@ class PluginManager {
...
+        },
+
+        setInterval (...args) {
+          const id = setInterval(...args)
+          timerArrays.intervals.push(id)
+          return id
+        },
+
+        setTimeout (...args) {
+          const id = setTimeout(...args)
```
alexbarnsley: This needs modifying because the below exposes `require`:

```
    walletApi.timers.setTimeout(function () {
      console.log('setTimeout', this.require)
    }, 100)
```

Probably work better this way:

```
        setTimeout (callback, timer) {
          const id = setTimeout(function () {
            callback()
          }, timer)
          timerArrays.timeouts.push(id)
          timerArrays.timeoutWatchdog[id] = setTimeout(() => timers.clearTimeout(id), timer)
          return id
        }
```

src/renderer/services/plugin-manager.js
```diff
@@ -714,6 +714,67 @@ class PluginManager {
...
+        get intervals () {
+          return timerArrays.intervals
+        },
+
+        get timeouts () {
+          return timerArrays.timeouts
+        },
+
+        setInterval (...args) {
+          const id = setInterval(...args)
```
alexbarnsley: Same regarding scope

src/renderer/services/plugin-manager.js
alessiodf: Thanks, good catch. Though your proposed approach would break the modern implementation for `setTimeout` and `setInterval` which allows you to use a function as the first argument, timer in ms as the second argument, and further arguments to pass to the called function in subsequent arguments. For example: `setTimeout(console.log, 100, "hello", "world")` should print `hello world` but that functionality would break with the proposed changes.

So I've refactored it like so:

```
setTimeout (...args) {
  const functionArgs = args.splice(2)
  const id = setTimeout(function () {
    args[0](...functionArgs)
  }, args[1])
  timerArrays.timeouts.push(id)
  timerArrays.timeoutWatchdog[id] = setTimeout(() => timers.clearTimeout(id), args[1])
  return id
}
```

Now using `this` doesn't expose anything it shouldn't.

src/renderer/services/plugin-manager.js
alessiodf: Same as above

src/renderer/services/plugin-manager.js
```diff
@@ -743,13 +743,19 @@ class PluginManager {
         },
 
         setInterval (...args) {
```
alexbarnsley: I think it's worth explicitly specifying the arguments here (`callback` and `interval`) since 2 arguments are always required anyway

src/renderer/services/plugin-manager.js
```diff
@@ -743,13 +743,19 @@ class PluginManager {
...
-          const id = setInterval(...args)
+          const functionArgs = args.splice(2)
+          const id = setInterval(function () {
+            args[0](...functionArgs)
+          }, args[1])
           timerArrays.intervals.push(id)
           return id
         },
 
         setTimeout (...args) {
```
alexbarnsley: Same here (e.g. `callback` and `timeout`)

src/renderer/services/plugin-manager.js
alessiodf: Done, using `method` instead of `callback` to prevent the `no-callback-literal` error in eslint.

src/renderer/services/plugin-manager.js
alessiodf: Done, using `method` instead of `callback` to prevent the `no-callback-literal` error in eslint.

PR #1469, opened 2019-09-24
fix: network modal missing version

alexbarnsley:
## Summary

The network `version` value wasn't being saved in previous wallet versions. This is a patch to re-populate the `version` from the network.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/components/Network/NetworkModal.vue
```diff
@@ -346,7 +347,29 @@ export default {
...
+
+        if (networkConfig) {
+          this.form.version = networkConfig.version.toString()
+        } else {
+          try {
+            ClientService.fetchNetworkConfig(this.form.server)
+              .then(network => {
+                console.log('fetchNetworkConfig')
+                this.form.version = network.version.toString()
+                this.$error('Address Version was missing - please check your wallets')
```
luciorubeens: Maybe move this to the locale file and remove debugging.

alexbarnsley: Yes you're right 🤔 will do it a bit later on. Thanks 👌

alexbarnsley:
oops

src/renderer/components/Network/NetworkModal.vue
alexbarnsley: Done @luciorubeens

PR #1472, opened 2019-09-26
chore: adjust workflows to match all branches on push

dated:
## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,11 +1,12 @@
 name: Lint
 
 on:
-    push:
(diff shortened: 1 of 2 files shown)


faustbrian:
Leaving this here for reference https://github.community/t5/GitHub-Actions/glob-in-path-filter-is-still-unsupported-despite-the-newly/td-p/32425.

I'll leave the PR open until a GitHub developer on that or some other topic confirms that they settled on a format for their pattern matching.

PR #1505, opened 2019-10-08
fix: get version of custom peer

### LINKED ISSUE #1504: Dashboard with custom peer
When I use a custom peer (no matter which IP), the dashboard is not loading anything, it's not querying the `/api/v2/wallets/search` and `/api/v2/transactions/search` endpoints.

**Expected Behavior**
To query those endpoints and show my latest txs.

**Current Behavior**
If I click `Connect to best` it works, but if I set the same IP as a custom peer, it doesn't load anything.

1. Click the cloud icon to change the peer
2. Connect to best and copy the IP
3. Set the same IP as a custom peer
3. Reload the app
4. Dashboard is showing the loading animation but nothing is loading...

**Your Environment**

- Version used: 2.6.2
- Operating System and version: macOS Mojave

dated:
## Summary

Custom peers are missing the `version` property, causing the capability check to error. Fixes #1504.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


luciorubeens:
I think Alex is solving this in https://github.com/ArkEcosystem/desktop-wallet/pull/1469. @alexbarnsley What do you think?

alexbarnsley:
@luciorubeens my PR fixes an issue with the NetworkModal, whereas it looks like this one fixes issues fetching the version when validating custom peers so they're different 👌

PR #1519, opened 2019-10-23
misc: it-IT update

danielstc:
## Summary
Italian language update

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


danielstc:
> Thanks for the translations! Apologies for the delay, we have been trying to decide how to go about dealing with languages in the future. Our plan is to use plugins to provide languages externally. I will merge this but will not be merging anymore after it. Thank you!

Official plugins or community plugins?

alexbarnsley:
These will most likely be community plugins. The problem we have up to now is we cannot verify the quality of translations that are submitted. This isn’t aimed at anyone in particular, but we don’t have people internally who can verify translations are accurate. We will keep you updated on how to integrate translations as a plugin in due course. Thank you!

danielstc:
@alexbarnsley 
Okay, thanks. I hope I can keep doing that. Yes, Sam had already explained this to me. I may be repetitive, but I think it's important for the end user to have their own language as an interface, just as it's important for developers to have their own reference language, especially if you have to manage value.  I just find it a bit dangerous to leave everything in the hands of the community without filters, as here on GitHub you can still do a review, while with the plugins, as I understand it, there is no kind of verification. Someone could misrepresent the text with malicious intentions.
Maybe now it's premature, but in the future it would be interesting to have someone on your team, or someone from outside qualified to check. As an alternative you should have the community that reviews the translations, but it is clear that you need a control by a multitude of users to be reliable.

PR #1554, opened 2019-11-22
fix: sandbox on initial plugin load

alexbarnsley:
## Summary

Fixes issue with additional sandbox data not being available in the initial plugin file (`src/index.js`).

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/renderer/services/plugin-manager/plugin-sandbox.js
+++ b/src/renderer/services/plugin-manager/plugin-sandbox.js
@@ -52,7 +52,9 @@ export class PluginSandbox {
     }
 
     return {
-      document
+      document,
+      ...this.sandbox,
+      walletApi: this.walletApi
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/renderer/services/plugin-manager/plugin-sandbox.js
```diff
@@ -52,7 +52,9 @@ export class PluginSandbox {
     }
 
     return {
-      document
+      document,
```
luciorubeens: Hm, just tested it and we use the `document` object just for compiling templates. Then we can use `getComponentVM()` for the initial file instead of adding the same sandbox for both.

alexbarnsley: 🤔 I'm not sure what you mean. If you've already done the changes to test it, do you want to open a PR with the changes and I take a look?

alexbarnsley:
Dealt with in #1560

PR #1611, opened 2020-01-20
fix: reset page when perpage update

### LINKED ISSUE #1606: bug: transaction pagination rows
**Expected Behavior**

When I change the amount of rows while being on a specific page of transactions, it should not show 0 transactions when I switch to a larger rows per page amount.

**Current Behavior**

When you switch from 10 rows per page to 50 rows per page while having paginated, you can run into an empty page.

**Possible Solution**

Don't show an empty page, either default to page 1 when switching or show the transactions you were at but with the new row count (e.g. if you were a tx 40 (4x 10) and switch to 50 rows, it can show tx 40-90)

1. Go to a wallet with multiple pages of transactions 
2. Set rows per page to 10 if it isn't already
3. Navigate to the next page a couple times
4. Switch rows per page to 50
5. Press "prev" to go back a page
6. End up with an empty page if you now surpass the amount of transactions the wallet has (for example there are 100 transactions, you were on page 8 first, then switching to 50 rows would try to fetch transactions 350-400 which don't exist). It also shows an incorrect transaction count in the pagination

10 rows:
image

switched to 50 rows:
image

davimello28:
## Summary

Fix #1606 

Reset the `currentPage` to `1` after updating the rows. 

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/renderer/components/Wallet/WalletTransactions/WalletTransactions.vue
+++ b/src/renderer/components/Wallet/WalletTransactions/WalletTransactions.vue
@@ -272,6 +272,7 @@ export default {
 
     onPerPageChange ({ currentPerPage }) {
       this.transactionTableRowCount = currentPerPage
+      this.currentPage = 1 // Reset page to 1 when change display rows
       this.__updateParams({ limit: currentPerPage, page: 1 })
       this.loadTransactions()
     },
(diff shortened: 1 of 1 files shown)


alexbarnsley:
@davimello28 can you fix conflicts please

alexbarnsley:
@davimello28 can you fix conflicts and add a quick test please

dated:
FYI i added tests for this method in #1612, I can add this change there and adapt the tests accordingly.

alexbarnsley:
I'm happy for you to do that @dated 👍

PR #1612, opened 2020-01-20
refactor: set currentPage / currentPerPage only if different from current value

dated:
## Summary

Follow up to #1611.

When the per page count is changed vue-good-table emits the `on-per-page-change` event twice, which in turn result in two ~api calls to be made~ calls to the `loadTransactions` method. By checking if the current value is equal to the emitted value first ~only one request will be done~ the above method will be called only once.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


davimello28:
This works better.

dated:
Relevant issue in the `vue-good-table` repo: https://github.com/xaksis/vue-good-table/issues/654

PR #1633, opened 2020-01-29
fix: avatar error on new profile during onboarding

### LINKED ISSUE #1613: Not selecting an avatar returns a TypeError
The smallest user flow into the app (Next > Next > Next)  returns a TypeError. 

**Expected Behavior**

No errors.

**Current Behavior**

Return errors. Restarting the wallet seems to make everything work as expected.

image

**Possible Solution**

Either select a default avatar, prevent users from starting the wallet without an avatar or accept starting without an avatar. 

1. Reset the configurations of your wallet (Config > Reset Data)
2. Main Screen from Onboarding > Click Next.
3. Skip
4. Provide a Profile Name.
5. Next
6. Next
7. Done
8. See the errors.  

**Your Environment**

- Version used: `develop` branch.
- Environment name and version : Firefor 72.0.1 
- Operating System and version: Debian 10

brenopolanski:
## Summary

Fixes #1613.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/App.vue
```diff
@@ -155,7 +155,7 @@ export default {
       return this.$store.getters['session/background'] || `wallpapers/${this.hasSeenIntroduction ? 1 : 2}Default.png`
     },
     hasAnyProfile () {
-      return !!this.$store.getters['profile/all'].length
```
alexbarnsley: I feel like the previous line was a better check, so maybe look into the `profile/all` store module to see what can be changed there instead. Could you then also add tests for that scenario please (if the file is missing, just add tests for the method you change)

brenopolanski: Well... anyway the code  `return this.$store.getters['session/profile']` will return an active profile. 
@alexbarnsley maybe change the computed property from `hasAnyProfile` to `hasProfile`?

alexbarnsley: Yeh can do. It's not a major change. In theory `session/profile` should always be set so it should be okay. Could you add some tests for this please. There isn't currently a test file so just add the mounting stuff and test this particular computed method. Thanks!

alexbarnsley: @brenopolanski 👀

__tests__/unit/App.spec.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  it('should have the right name', () => {
+    const wrapper = mountPage()
+    expect(wrapper.name()).toEqual('DesktopWallet')
+  })
+
+  describe('Computed properties', () => {
+    describe('hasProfile', () => {
+      it('should has a profile', () => {
```
alexbarnsley:
```suggestion
      it('should have a profile', () => {
```

src/renderer/App.vue
```diff
@@ -160,8 +160,8 @@ export default {
     background () {
       return this.$store.getters['session/background'] || `wallpapers/${this.hasSeenIntroduction ? 1 : 2}Default.png`
     },
-    hasAnyProfile () {
-      return !!this.$store.getters['profile/all'].length
+    hasProfile () {
```
alexbarnsley: `is` and `has` should really return a boolean

brenopolanski: @alexbarnsley do you prefer to use `isProfile()` instead of `hasProfile()` ?

alexbarnsley: I think `hasProfile` is better in this context but it just needs to be a boolean

PR #1634, opened 2020-01-29
fix: add truncate filter to the wallet balance in the profile

### LINKED ISSUE #1630: The wallet balance leaves the grid div margins
**Expected Behavior**

Screenshot from 2020-01-27 22-18-42

**Current Behavior**

Screenshot from 2020-01-27 22-08-48

**Possible Solution**

Just use a [truncate filter](https://github.com/ArkEcosystem/desktop-wallet/tree/develop/src/renderer/filters) in https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/renderer/pages/Profile/ProfileAll.vue#L41 

1. Go to My wallets;
2. Click on Import Wallet;
3. Import the address `AUDud8tvyVZa67p3QY7XPRUTjRGnWQQ9Xv`
4. Go to profile.

**Your Environment**

- Version used: ark-desktop-wallet 2.7.0
- Environment name and version: Chrome v76.0.3809.146 - Node.js v12.4.0

brenopolanski:
## Summary

Fixes #1630.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/pages/Profile/ProfileAll.vue
```diff
@@ -38,7 +38,7 @@
             </div>
 
             <span class="font-bold my-2 text-lg">
-              {{ profileBalance(profile) }}
+              {{ profileBalance(profile) | truncate(15) }}
```
alexbarnsley: Could you add tests for this, please. An alternative place to truncate could be in the `profileBalance` method as that already has 1 test, just add additional tests for longer variations.

dated: Wouldn't the `truncate` css class a better alternative here? Truncating like above will yield unexpected results for locales where the currency symbol follows the amount.

brenopolanski: @alexbarnsley tests added.

alexbarnsley: @brenopolanski I actually like @dated's suggestion to use the `truncate` css class 😅 Would you mind trying that approach instead please. Sorry, I only just saw dated's reply

brenopolanski: @alexbarnsley @dated i tested using `truncate` css class and the result is better:

Screenshot from 2020-02-12 11-43-32

For the `truncate` class works with `<span>` tag it's necessary to use the `display: inline-block` property, then the `.my-2` class works well now.

brenopolanski:
@alexbarnsley can you review again this PR?

PR #1638, opened 2020-01-30
refactor: simplify blurred modal backdrop & ui tweaks

### LINKED ISSUE #1632: Side menu in horizontal mode has menu items aligned to the top
**Expected Behavior**

Screenshot from 2020-01-28 21-38-57

**Current Behavior**

Screenshot from 2020-01-28 21-39-15

**Possible Solution**

**(1)** For the menu items **Settings** and **Network Status** set `height: 100%` or use tailwindcss `h-full`.

**(2)** The alignment of the app in horizontal mode is squeezed with the right scroll. Removing the `width: calc(100vw + 1rem);` in https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/renderer/App.vue#L435 it works well.

1. Just resize the app to 750px

Peek 2020-01-28 22-01

**Your Environment**

- Version used: ark-desktop-wallet 2.7.0
- Environment name and version: Chrome v76.0.3809.146 - Node.js v12.4.0

### LINKED ISSUE #1636: bug: bleeding background on blurred elements
As it turns out there's the tiniest bleed at the edges even when using a negative margin on the blurred element. I will refactor this to use `backdrop-filter: blur(4px);` instead, which appears to get properly rid of the bleeding edges and doesn't require the nasty workaround with negative margins.

image

left: refactored / no bleed

dated:
## Summary

Closes #1632, closes #1636.

Also includes several other tweaks to the menu and padding, and fixes the rounded borders on the wallet details.

There's are some more problems with the wallet details component but that will require some more investigation.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


brenopolanski:
I tested here.. and this PR solved the problems in the issue #1632 :+1:

src/renderer/App.vue
```diff
@@ -154,9 +129,6 @@ export default {
   }),
 
   computed: {
-    hasBlurFilter () {
```
brenopolanski: I'm getting the message on my devtools console: `Property or method "hasBlurFilter" is not defined on the instance but referenced during render.` Because of `hasBlurFilter` here: https://github.com/ArkEcosystem/desktop-wallet/blob/6eb4b85a3fa35b9e60b7a45f9e6f56d5ff87e1ab/src/renderer/App.vue#L83

dated: Looks like the `button-dropdown` got added in one of the merges, I'll need to take a look at how it uses the blur class.

dated: Ok, looks like the `blur` on the button dropdown can be safely removed. Strangely the button dropdown wasn't positioned correctly for me during development (`this.$refs.buttonDropdown` always being `undefined`) while it displays correctly on 2.8.0, therefore I added a resize handler to also reposition the dropdown while it is open. @alexbarnsley could you have a look at this also? In particular 85700dc which adds the resize handler.

davimello28:
Closes #1639

PR #1645, opened 2020-02-02
fix: handle invalid address from qr code

dated:
## Summary

The `InputDelegate` component assumed that every address scanned from a qr code actually belonged to a delegate. This PR adds an additional check to make sure it does and removes the call to truncate the delegates username, since usernames are capped at a length of 20 by default.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/renderer/components/Input/InputDelegate.vue
+++ b/src/renderer/components/Input/InputDelegate.vue
@@ -65,7 +65,6 @@ import ModalQrCodeScanner from '@/components/Modal/ModalQrCodeScanner'
 import { MenuDropdown } from '@/components/Menu'
 import Cycled from 'cycled'
 import InputField from './InputField'
-import truncate from '@/filters/truncate'
 import { includes, isEmpty, map, orderBy } from 'lodash'
 
 export default {
@@ -176,7 +175,7 @@ export default {
           publicKey: object.publicKey
         }
 
-        delegate.name = `${truncate(object.username, 25)} (${this.wallet_truncate(object.address)})`
+        delegate.name = `${object.username} (${this.wallet_truncate(object.address)})`
 
         return delegate
       })
--- a/src/renderer/components/Input/InputDelegate.vue
+++ b/src/renderer/components/Input/InputDelegate.vue
@@ -306,7 +306,9 @@ export default {
         this.$error(this.$t('MODAL_QR_SCANNER.DECODE_FAILED', { data: value }))
       }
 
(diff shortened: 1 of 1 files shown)


brenopolanski:
LGTM!

This PR solves the following problem, for instance:

If the wallet is using the `mainnet` and scanning a QR Code from a `devnet` delegate, the UI returns:

Screenshot from 2020-02-13 22-37-39

Using the code:

```javascript
this.model = this.$store.getters['delegate/byAddress'](address).username
```

It will return `undefined`.

Screenshot from 2020-02-13 22-40-25

brenopolanski:
@dated 

There is a bug with `classList.includes` (https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/renderer/components/Input/InputDelegate.vue#L246) and you can solve here on this PR.

> The `Element.classList` is a read-only property that returns a live `DOMTokenList` collection of the class attributes of the element. 

**Source:** https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

So the methods available in the `classList` are:

- DOMTokenList.item(index)
- DOMTokenList.contains(token)
- DOMTokenList.add(token1[, token2[, ...tokenN]])
- DOMTokenList.remove(token1[, token2[, ...tokenN]])
- DOMTokenList.replace(oldToken, newToken)
- DOMTokenList.supports(token)
- DOMTokenList.toggle(token [, force])
- DOMTokenList.entries()
- DOMTokenList.forEach(callback [, thisArg])
- DOMTokenList.keys()
- DOMTokenList.values()

`includes` is not available in `classList`, so you can change to `contains`.

dated:
I'll take a look!

dated:
`InputAddress`, `InputDelegate` and `InputSelect` all have references to `classList`. I adjusted all three of them. The `onBlur` method could possibly be moved to it's own mixin, at least in `InputAddress` and `InputDelegate` it is exactly the same.

brenopolanski:
> `InputAddress`, `InputDelegate` and `InputSelect` all have references to `classList`. I adjusted all three of them. The `onBlur` method could possibly be moved to it's own mixin, at least in `InputAddress` and `InputDelegate` it is exactly the same.

Works fine @dated :+1:

brenopolanski:
@dated I agree with you about creating a mixin for the `onBlur` method. Let's see what @alexbarnsley  thinks about it...

Issue #1651, opened 2020-02-06
refactor: make modals reusable and stackable.


davimello28:
This can be considered a bug, but nothing is affected yet since we don't have any modal stacking each other.

I am creating a `ModalCloseConfirmation` component that asks to confirm when exiting the transaction modal (`Wallet > Send`), similar to the one when deleting a contact. I was expecting to reuse `ModalConfirmation`, but it is not possible since the `TransactionModal` already uses `ModalWindow`.

It is impossible to create a modal that descends from another modal. We have multiple problems like broken circular reference, broken UI and event leak. 

Other dialog screens like `Loading` and `QR-scan` are using their own portals. 

![image](https://user-images.githubusercontent.com/3607502/73970456-f903fa00-48fb-11ea-8e8a-d000529006ef.png)

**Possible Solution**

- Enable Modals to use circular reference correctly. We should be able to call `ModalConfirmation` from within a `ModalWindow`, or `ModalWindow` from `ModalWindow`. This can make easier to implement new flows, especially considering the developer roadmap.
- Create the concept of `dialog` or `alert` for windows that cannot stack and let modals stack.  

**Ambient**
- Version used: `develop`. 
- Environment name and version (e.g. node.js v10.15.3): 12.12.0
- Operating System and version: Debian 10

PR #1654, opened 2020-02-09
feat: close confirmation modal

davimello28:
Displays a close confirmation when try closing with backdrop the click or button if `props.confirmClose` is specified on `ModalWindow`.

## Summary

Check #1651. 

## Checklist

- [X] Tests _(if necessary)_
- [X] Ready to be merged


src/renderer/components/Modal/ModalCloseConfirmation.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <div
```
alexbarnsley: Out of curiosity, how come you didn't use the ModalWindow component as the other modals do? I think it may be worth making them consistent

alexbarnsley: I think I know. Is it because the confirm modal is shown inside ModalWindow?

src/renderer/components/Modal/ModalWindow.vue
```diff
@@ -136,11 +150,18 @@ export default {
...
 
   data: () => ({
-    isMaximized: true
+    isMaximized: true,
+    showConfirmationModal: false,
+    hasChanged: false
   }),
 
+  beforeCreate () {
+    this.$options.components.ModalConfirmation = require('./ModalConfirmation').default
```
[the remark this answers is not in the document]
davimello28: This was due to a previous circular reference from `ConfirmationModal`.

src/renderer/components/Modal/ModalCloseConfirmation.vue
davimello28: Yep. We cannot use `ModalWindow` inside `ModalWindow`. I was expecting to use `ModalConfirmation`, but it also uses `ModalWindow`. Modals cannot stack (see #1651), so I have to create an entire component isolated from the rest.

alexbarnsley:
@davimello28 there seems to be a failing test

faustbrian:
https://github.com/ArkEcosystem/desktop-wallet/pull/1654/files#diff-6dc9bdfca16f596aee2b06566ee40a74R10 mousedown is no longer used for the closing so the test is outdated and failing.

PR #1656, opened 2020-02-10
feat: add coincap price api

brenopolanski:
## Summary

Add the CoinCap pricing API. This PR implements CoinCap as a new option at the price provider.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/renderer/services/price-api/coin-cap.js
```diff
@@ -0,0 +1,213 @@
...
+   * @param {String} token
+   * @param {String} currency
+   * @param {Number} limit
+   * @param {String} type
+   * @param {String} dateFormat
+   * @return {(Object|null)} Return API response data or null on failure
+   */
+  static async fetchHistoricalData (token, currency, days, _, dateFormat = 'DD.MM') {
+    const tokenId = await this.getTokenId(token)
+    const currencyData = await this.getCurrencyData()
```
alexbarnsley:
```js
    const { rates } = await this.getCurrencyData()
```

src/renderer/services/price-api/coin-cap.js
```diff
@@ -0,0 +1,213 @@
...
+  /**
+   * Fetch market data from API
+   * @param {String} token
+   * @return {(Object|null)} Return normalized market data or null on failure
+   */
+  static async fetchMarketData (token) {
+    const tokenId = await this.getTokenId(token)
+    const currencyData = await this.getCurrencyData()
+
+    return this.__transformMarketResponse(currencyData, tokenId)
```
alexbarnsley: As each variable is only used once, maybe do something like:

```js
  return this.__transformMarketResponse(
    await this.getCurrencyData(),
    await this.getTokenId(token)
  )
```

src/renderer/services/price-api/coin-cap.js
```diff
@@ -0,0 +1,213 @@
...
+    if (this.tokenLookup) {
+      return this.tokenLookup[token.toUpperCase()]
+    }
+
+    try {
+      const uri = `${MARKET.source.coinCap}/assets?limit=${limit}`
+      const { body } = await got(uri, {
+        json: true
+      })
+      const { data } = body
```
alexbarnsley: This is unnecessary since `data` is only used once. I'd just use `body.data.reduce...`

src/renderer/services/price-api/coin-cap.js
```diff
@@ -0,0 +1,213 @@
...
+  static async getCurrencyData () {
+    try {
+      const uri = `${MARKET.source.coinCap}/rates`
+      const { body } = await got(uri, {
+        json: true
+      })
+      const { data, timestamp } = body
+      const arkData = await this.fetchArkData()
+
+      const assets = data.reduce((map, value, index) => {
```
alexbarnsley: Instead of 2 reduce loops, do 1 (removing the hardcoded ark references):

```
  const response = {
    assets: { [arkData.symbol.toUpperCase()]: arkData },
    rates: { [arkData.symbol.toUpperCase()]: arkData.rateUsd },
    timestamp
  }

  foreach (const value of data) {
    response.assets[value.symbol.toUpperCase()] = value
    response.rates[value.symbol.toUpperCase()] = value.rateUsd
  }

  return response
```

src/renderer/services/price-api/coin-cap.js
```diff
@@ -210,4 +196,33 @@ export default class CoinCapAdapter {
...
+  /**
+   * Convert an amount of money to another currency
+   * @param {Number} amount
+   * @param {Object} options
+   * @return {Number} Returns conversion result
+   */
+  static __convert (amount, { from, to, base, rates }) {
+    if (from && to) {
+      // If `from` equals `base`, return the basic exchange rate for the `to` currency
+      if (from === base && has(rates, to)) {
```
alexbarnsley: Does `has` only check if the key exists? If so it could be swapped out for `Object.prototype.hasOwnProperty.call(rates, to)`. It might be longer but saves on relying on external packages. In the future, we'd like to remove lodash completely (if possible)

brenopolanski: Good tip :+1:

PR #1667, opened 2020-02-11
fix: business, bridgechain & delegate tweaks

alexbarnsley:
## Summary

- Fixes typos for Business & Bridgechain name validation
- Fixes typos for successful Business & Bridgechain transaction creation messages
- Fixes bug where Business Resign button shows for non-business wallets
- Fixes bug where Business Resign button shows for resigned business
- Fixes bug where Delegate Resign button shows for resigned delegates

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/services/wallet.js
```diff
@@ -127,7 +127,6 @@ export default class WalletService {
    * @returns {Boolean}
    */
   static canResignDelegate (wallet) {
-    console.log('canResignDelegate', wallet)
```
davimello28: Why we don't have a debug mode?

alexbarnsley: That was debug code I added during development. Leaving stuff like that in would make the code messy. We use `logger.$error` for some areas where it might be useful (e.g. a try/catch) but that's all.

brenopolanski: It has already been removed!

PR #1668, opened 2020-02-12
fix: transaction list multi-payment amount

alexbarnsley:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/i18n/locales/en-US.js
```diff
@@ -991,6 +991,7 @@ export default {
       HTLC_CLAIM: 'Timelock Claim could not be created',
       HTLC_REFUND: 'Timelock Refund could not be created',
       MULTI_PAYMENT: 'Multi payment could not be created',
+      MULTI_PAYMENT_TO_SELF: 'Excluding {amount} the sender sent to themselves',
```
dated: FYI we eventually settled on `themself` instead of `themselves` on the explorer (https://dictionary.cambridge.org/dictionary/english/themself).

alexbarnsley: I'm going with `Excluding {amount} sent to itself` since I'm not happy with either to be honest.

alessiodf: Agree that neither `themself` nor `themselves` sounds natural in this context, ditto with the explorer.

brenopolanski:
@alexbarnsley a fix for this PR.

A wallet renamed using 120 characters exceeds the limit of the modal horizontally when the UI is **<1100px**.

Screenshot from 2020-02-17 14-01-45

Screenshot from 2020-02-17 14-01-55

As a suggestion, truncate the `recipient`.

PR #1583, opened 2019-12-11
test: plugin manager unit tests

dated:
## Summary

Adds unit tests for plugin manager related parts, mainly the plugin store module.

image

## Checklist

- [ ] Documentation _(if necessary)_
- [x] Tests _(if necessary)_
- [x] Ready to be merged

PR #1679, opened 2020-02-18
refactor: omit themes from list of all plugins

alexbarnsley:
## Summary

Hides Themes when listing all plugins. Instead, themes will only be visible from their own category

image

Excluded tests while they are being written in #1583 

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/renderer/store/modules/plugin.js
+++ b/src/renderer/store/modules/plugin.js
@@ -44,7 +44,9 @@ export default {
 
         let match = true
 
-        if (category && category !== 'all') {
+        if (category === 'all') {
+          match = match && !plugin.config.categories.includes('theme')
+        } else if (category && category !== 'all') {
           match = match && plugin.config.categories.includes(category)
         }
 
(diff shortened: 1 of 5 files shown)


src/renderer/store/modules/plugin.js
```diff
@@ -44,7 +44,9 @@ export default {
 
         let match = true
 
-        if (category && category !== 'all') {
+        if (category === 'all') {
+          match = match && !plugin.config.categories.includes('theme')
+        } else if (category && category !== 'all') {
```
dated:
```diff
- } else if (category && category !== 'all') {
+ } else if (category) {
```
Too late but here but the check against `'all'` can be removed.

ItsANameToo: yes

alexbarnsley: where were you? we needed you

PR #1691, opened 2020-02-20
fix: splashscreen fullscreen & ui tweaks

alexbarnsley:
## Summary

- Disable saved fullscreen state & apply once main window shows
- UI tweaks - spacing & disable text selection

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -197,9 +197,7 @@ app.on('ready', () => {
 })
 
 app.on('window-all-closed', () => {
-  if (process.platform !== 'darwin') {
-    app.quit()
-  }
+  app.quit()
 })
 
 app.on('activate', () => {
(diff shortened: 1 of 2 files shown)


src/main/index.js
```diff
@@ -192,9 +195,7 @@ app.on('ready', () => {
 })
 
 app.on('window-all-closed', () => {
-  if (process.platform !== 'darwin') {
-    app.quit()
-  }
+  app.quit()
```
ItsANameToo: :thinking: might want to discuss this to be sure if this is what we want to have happen on mac

alexbarnsley: Me and Sam discussed it :colbert: it fixes a bug, plugins can't be installed after: closing all windows (not quitting the app) then opening the window again and trying to install a plugin

dated: Has anyone looked into that yet? :hmhm:

Issue #1687, opened 2020-02-20
Custom peer


zillionn:
1. When click `Connect custom peer` it shows the port in the host field:
<img width="409" alt="Screenshot 2020-02-20 at 14 45 52" src="https://user-images.githubusercontent.com/3462869/74934829-b5c38400-53ef-11ea-8049-7c56bf9539a0.png">

2. If I add a forward slash `/` after the host, i.e. `http://1.2.3.4/`, the peer will not be accepted:
<img width="844" alt="Screenshot 2020-02-20 at 14 44 11" src="https://user-images.githubusercontent.com/3462869/74934714-7bf27d80-53ef-11ea-9f7d-6f7970a25dd0.png">

- Version used: 2.8.0
- Operating System and version: macOS Catalina 10.15.3

ArkEcosystemBot:
**Thanks for opening this issue!** A maintainer will review this in the next few days and explicitly select labels so you know what's going on.

_If no reviewer appears after a week, a reminder will be sent out._

ArkEcosystemBot:
This issue has been closed. If you wish to re-open it please provide additional information.

PR #1621, opened 2020-01-23
fix: peer validation

davimello28:
## Summary

- Fixes #1618 and other possible https -> http on fetching new peers. 
- Accepts an instance of `URL()` on `validatePeer`.

## Checklist

- [X] Documentation _(if necessary)_
- [X] Tests _(if necessary)_
- [ ] Ready to be merged

PR #1692, opened 2020-02-21
fix: custom peer host validation

### LINKED ISSUE #1298: Custom peer window labelling is unclear
**Current Behavior**

The current labelling on the ‘connect to custom peer’ window is unclear. The Host/IP field requires a protocol prefix but the field is labelled ‘IP/Host’ not URL, not is an example given (until you get it wrong)

The port field should be pre-filled with the default network port

**Possible Solution**

Give an example underneath the ‘IP/Host’ field, and rename it only to ‘host’ or ‘Server’

Pre-fill the port field with the default port.

[Issue reported in Slack]

davimello28:
## Summary

- Closes both issues on #1687
- Closes #1298 

## Checklist

*This is a temporary fix.*

- [X] Documentation _(if necessary)_
- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


src/renderer/components/Modal/ModalPeer.vue
```diff
@@ -133,7 +133,9 @@ export default {
   mounted () {
     if (this.currentPeer) {
       if (this.currentPeer.host) {
-        this.form.host = this.currentPeer.host
+        const scheme = this.currentPeer.isHttps ? 'https://' : 'http://'
```
brenopolanski: You can move this line to the top, because `line 140` has the same code.

Example:

```
  mounted () {
    if (this.currentPeer) {
      const scheme = this.currentPeer.isHttps ? 'https://' : 'http://'
      
      if (this.currentPeer.host) {
        const hostname = new URL(this.currentPeer.host).hostname
        this.form.host = scheme + hostname
      } else if (this.currentPeer.ip) {
        this.form.host = `${scheme}${this.currentPeer.ip}`
      }

      if (this.currentPeer.port) {
        this.form.port = this.currentPeer.port
      }
    }
  }
```

src/renderer/components/Modal/ModalPeer.vue
```diff
@@ -147,6 +149,10 @@ export default {
 
   methods: {
     emitConnect () {
+      const host = this.form.host
+      if (host.endsWith('/')) {
+        this.form.host = host.slice(0, -1)
+      }
```
brenopolanski: I tested using `////` at the end:

Screenshot from 2020-02-21 10-19-02

You may be able to create a recursive function to remove 1 or more slashes at the end. Like this:

```
    emitConnect () {
      const { host } = this.form
      const path = url => url.endsWith('/') ? path(url.slice(0, url.length - 1)) : url
      this.form.host = path(host)
      this.$emit('connect', {
        peer: this.form,
        closeTrigger: this.closeTrigger
      })
    }
```

davimello28: As I said before, I don't think that we should try to cover all the possibilities of input now. I have a task to refactor the whole peer system and this will probably be changed in a near feature.

I can make it recursive but the system should support paths for the peers. The application is not responsible for resolving the URLs.

src/renderer/components/Modal/ModalPeer.vue
davimello28: Done!

src/renderer/components/Modal/ModalPeer.vue
```diff
@@ -147,6 +149,10 @@ export default {
 
   methods: {
     emitConnect () {
+      const host = this.form.host
```
brenopolanski: Here you can use (destructuring): `const { host } = this.form`

src/renderer/components/Modal/ModalPeer.vue
brenopolanski: Okay :+1:

alexbarnsley:
@davimello28 are we keeping #1621 and this?

PR #1693, opened 2020-02-21
ci: draft releases when merging to master

faustbrian:
Drafts a new release based on the `version` property in the `package.json` file. The changelog is currently not automatically created but could be added if needed.

It's important to use the versioning commands of `npm` or `yarn` to ensure that the workflows and artifacts are properly versioned. Manually editing files should be avoided as those commands exist for a reason.

**This intentionally only runs on a push to master as what is merged should be 100% working based on tests and builds on `develop` passing.**


dated:
unfortunately this breaks builds for win and macos on forks since the build steps depend on github secrets which, well... are secret.

faustbrian:
You'll still be busy another 8 weeks with tests so don't act like it's urgent @dated  :trollface:

dated:
:cripes:

PR #1736, opened 2020-02-28
refactor: adjust business transaction form validation

dated:
## Summary

Similar to #1730, but for business transactions.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/components/Transaction/TransactionForm/TransactionFormBusiness/mixin.vue
```diff
@@ -161,10 +161,53 @@ export default {
...
       return `${this.$t('WALLET_BUSINESS.NAME')} - ${this.$t('VALIDATION.MAX_LENGTH', [maxNameLength])}`
     },
 
+    isFormValid () {
+      if (this.isUpdate) {
+        return !this.$v.form.$invalid && !(
+          this.hasSameName &&
+          this.hasSameWebsite &&
+          this.hasSameVat &&
+          this.hasSameRepository
```
brenopolanski: As `VAT` and `Repository` are not **required fields**, It's not necessary to validate these fields using the functions `hasSameVat()` and `hasSameRepository()`.

dated: Those are needed to remove the properties when the value didn't change, as to not be included in the update, when in fact the value hasn't been changed.

brenopolanski: If I change the values to `VAT` and `Repository` I don't see it in `businessAsset` variable. Now if the values are empty fields it returns me as: `{name: "test", vat: "", repository: ""}`.

dated: I'll recheck this later. It will need some adjustments anyway for the empty values on update, as "unsetting" these optional fields won't be implemented in core, https://github.com/ArkEcosystem/core/issues/3552

dated: All cases should be handled correctly now, let me know if you find anything else @brenopolanski

brenopolanski: @dated works fine... but on "Bussines Update" i have `VAT` and `Repository` as required:

Screenshot from 2020-03-04 15-39-33

These fields should not be required in edit mode.

dated: Does your Business have these fields already defined? Then they are required by the form. I could change it to just discard empty values :thinking: whatever makes more sense

dated: Maybe add a helper-text saying that the values will stay the same instead of invalidating the inputs.

dated: The goal of having them required on update is to not let the user think that they can remove those values, which is not possible.

brenopolanski: yeah @dated these fields were defined.
@alexbarnsley what do you think?

alexbarnsley: I had a chat with @dated about it. Since core requires those fields if they've already been provided, I think they should show as "required" if the data is removed. I've tested and seems to work okay

PR #1744, opened 2020-02-29
[WIP] feat: animate interactions with input editable list

### LINKED ISSUE #1731: ux: insufficient user feedback when adding recipients to multipayment and other lists
When adding recipients to a multipayment transaction there is insufficient feedback to the user that the recipient was added correctly.

The first few recipients are obvious, as the user gets to see them added to the list, when the list has reached his max height however the only indication of success is the changed count of items above the list.

As a workaround the list of items could be displayed in reverse, so that a newly added items is added at the top of the list.

To aid this a short animation on the background color or the likes could be added.

The above applies also to the other uses of `InputEditableList` (seed nodes, multisignature registrations, ...)

dated:
## Summary

Closes #1731.

This PR adds a transition group around the items of the `InputEditableList` component in order to animate the addition or removal of entries. For these animations a unique, primitive identifier has to be provided, the array index won't work as key. Since multiple components use this list component, each with a unique item structure, a common property is added to each item of the arrays used which can then be used as key.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
@brenopolanski could you check if you still get the scrollbars on linux?

brenopolanski:
@dated scrollbars still appear.

Sam (@avfcpieface) can you check if the scrollbars show on the Windows? See the issue too: https://github.com/ArkEcosystem/desktop-wallet/issues/1731

dated:
Thanks for checking @brenopolanski , I'll try some more.

dated:
I'm unable to fix this unfortunately as I can't test on linux. @brenopolanski do you want to take a look? Otherwise I'll close this for now and revisit it at a later time.

alexbarnsley:
@avfcpieface did you see Breno's comment here?

brenopolanski:
> I'm unable to fix this unfortunately as I can't test on linux. @brenopolanski do you want to take a look? Otherwise I'll close this for now and revisit it at a later time.

@dated i'll look at it

PR #1747, opened 2020-03-01
fix: don't manipulate raw transaction when it is stored

### LINKED ISSUE #1746: bug: resending transactions fails without error
Resending an expired transaction fails while still showing a success message, the reason being that the raw transaction is extended by a `timestamp` property when it is stored after being sent the first time. This will trigger the core validation to fail and return an `ERR_FROM_FUTURE`:

```
{"$id":[{"type":"ERR_FROM_FUTURE","message":"Transaction $id is 1490085705 seconds in the future"}]}
```

dated:
## Summary

Fixes #1746 by not manipulating the raw transaction.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


alexbarnsley:
Could you add tests for this please @dated

dated:
Are you satisfied, @alexbarnsley?

alexbarnsley:
It's a good start :trollface:

dated:
> I think that we should also use Date.now() from now on. [Just in case anyone is wondering why](https://stackoverflow.com/questions/12517359/performance-date-now-vs-date-gettime).

Yep that's exactly why I changed it here. `getTime()` only makes sense if you need to reference the same `Date` instance multiple times.

PR #1753, opened 2020-03-02
refactor: replace moment with dayjs

faustbrian:
Replaces all occurrences of `moment` with `dayjs`.


src/renderer/mixins/formatter.js
```diff
@@ -34,7 +34,20 @@ export default {
...
+
+      if (userLanguage === 'en-us') {
+        userLanguage = 'en'
+      }
+
+      try {
+        require(`dayjs/locale/${userLanguage}`)
+      } catch {
+        userLanguage = 'en'
+      }
```
dated: More generic:

```suggestion
      try {
        require(`dayjs/locale/${userLanguage}`)
      } catch {
        try {
          userLanguage = userLanguage.split('-')[0]
          require(`dayjs/locale/${userLanguage}`)
        } catch {
          userLanguage = 'en'
        }
      }
```

dated: This will also handle locales like `de-DE`, `it-IT` ...

faustbrian: Also more expensive due to how require works inside of electron.

dated: Gimme a minute, I'll boil it down to one `require`

dated: Update above

alexbarnsley:
`moment` is used in `src/renderer/services/crypto/transaction-signer.js` since a different PR was merged

PR #1755, opened 2020-03-02
perf: remove summary when marking announcement as read

dated:
## Summary

Since an announcement can't be marked as `unread` and the summary isn't shown once it has been marked as `read`, there is no reason to keep the summary around.

Includes a migration to remove the summary from already read announcements.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


dated:
Thanks @brenopolanski! Should be fixed by 9be6492.

PR #1766, opened 2020-03-03
feat: fetch known wallets from repository

dated:
## Summary

The known wallets are now available in the https://github.com/ArkEcosystem/common repository.

NB: When updating a profile vuex shows a warning, which is unrelated to these changes.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


davimello28:
@dated How are you planning to do this?

dated:
The known wallets will be fetched from https://github.com/ArkEcosystem/common

src/renderer/store/modules/network.js
```diff
@@ -79,12 +80,34 @@ export default new BaseModule(NetworkModel, {
...
+        }
+
+        if (network.knownWalletsUrl) {
+          try {
+            const knownWallets = await got(network.knownWalletsUrl, {
+              json: true
+            })
+            network.knownWallets = knownWallets.body
+          } catch (error) {
+            console.error('Could not retrieve known wallets: ', error)
```
brenopolanski:
```suggestion
            this._vm.$logger.error('Could not retrieve known wallets: ', error)
```

src/renderer/store/modules/network.js
```diff
@@ -93,7 +116,7 @@ export default new BaseModule(NetworkModel, {
         Managers.configManager.setConfig(cloneDeep(crypto))
         Managers.configManager.setHeight(constants.height)
       } catch (error) {
-        console.error('Could not update network data:', error)
+        console.error('Could not update network data: ', error)
```
brenopolanski:
```suggestion
        this._vm.$logger.error('Could not update network data: ', error)
```

src/renderer/components/Network/NetworkModal.vue
```diff
@@ -651,6 +667,14 @@ export default {
           return !this.showFull || /^https?:\/\//.test(value)
         }
       },
+      knownWalletsUrl: {
+        required () {
+          return true
```
alexbarnsley: I don't think this field should be required, otherwise the wallet won't be able to add bridgechains without it 🤔

dated: It always validates to true, it's basically a "Hack" so i could use the required url field error method.

alexbarnsley: so it does :trollface:

PR #1798, opened 2020-03-09
fix: unset innerHTML of plugin style when setting default theme

dated:
## Summary

When changing back from a plugin theme to a default theme the inner html of the dedicated `<style>` element isn't unset, which causes some custom styles to still stay in effect.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


dated:
:bleh:

faustbrian:
@dated conflicts

PR #1810, opened 2020-03-10
feat: multipayment display modals details

dated:
## Summary

**TransactionShow**

 - aligns the smartbridge to the right and the label to top

**TransactionConfirmMultiPayment**

 - adds the recipient count
 - adds the total amount
 - adds the vendorfield and fee

**TransactionModal**

- limits the max height to `80vh`

image
image
image

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/__tests__/unit/components/Transaction/TransactionConfirm/TransactionConfirmMultiPayment.spec.js
+++ b/__tests__/unit/components/Transaction/TransactionConfirm/TransactionConfirmMultiPayment.spec.js
@@ -1,5 +1,6 @@
 import { createLocalVue, mount } from '@vue/test-utils'
 import installI18n from '../../../__utils__/i18n'
+import CurrencyMixin from '@/mixins/currency'
 import TransactionConfirmMultiPayment from '@/components/Transaction/TransactionConfirm/TransactionConfirmMultiPayment'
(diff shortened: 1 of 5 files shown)


davimello28:
Everything seems ok to me, but I have a scroll bar on the recipients under Debian 10 at 1366x768.  

image

On the third screen that you showed above, there is a very small part of the third transaction displayed. I think this can lead the user to ignore the third transaction if the scrollbar isn't displayed. I think that it would be better if you either aways display it or remove the clip at all.

alexbarnsley:
All looks good, but we think the right-aligned vendorfield text looks a bit odd (I can see why you made it right-aligned, to match the other fields). I think one of the possible alternatives might be better:

- Revert to left-aligned
- Justify
- Move to the bottom (above payment list) and show it how it's shown on the confirmation screen:

image

dated:
I'll see how it looks with justified alignment. Alignment to the left is a big no-go here imho, it breaks the design too much. Showing it underneath would be the better alternative, compared to left-alignment that is.

dated:
I've went with `text-justify` for now @alexbarnsley.

PR #1816, opened 2020-03-11
fix: also delete global options when removing a plugin

dated:
## Summary

When removing a plugin only the options stored for a proper plugin id are deleted, the global options are not regarded. With these PR the "remove options" toggle shows up also for globally stored plugin options, and deletes those accordingly.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


davimello28:
I think it is not directly related to this, but I've used the ChangeNOW Plugin and set to delete everything. It seems that the ID is kept on `pluginOptions` even after uninstalling it. 

image

dated:
Unrelated but fixed in c85da28, thanks!

PR #1818, opened 2020-03-11
fix: periodically update bridgechain list

### LINKED ISSUE #1817: bug: the bridgechain list is not updated automatically after registration
After registering a bridgechain, the bridgechain list doesn't auto update. User is forced to either reload or change peer for it to show.

1. Go to wallet
2. Go to the secondary menu in the header
3. Click in Registration menu
4. Register  Bussines (if you don't have one)
5. Go to the Business tab
6. Register  Bridgechain (if you don't have one)
7. The bridgechain list is not updated automatically after registration

**Your Environment**

- Version used: ark-desktop-wallet 2.8.1
- Environment name and version: Chrome v80.0.3987.86 - Node.js v12.13.0

brenopolanski:
## Summary

Fixes #1817.

## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


alexbarnsley:
I'm not sure this will work as expected because it only reloads the bridgechain list after submitting the transaction (unless I misunderstood). The problem is that the transaction isn't forged instantly. I think it's best to have a timer on that bridgechain list to keep refreshing the list every minute or so to check for changes

brenopolanski:
> I'm not sure this will work as expected because it only reloads the bridgechain list after submitting the transaction (unless I misunderstood). The problem is that the transaction isn't forged instantly. I think it's best to have a timer on that bridgechain list to keep refreshing the list every minute or so to check for changes

@alexbarnsley what do you think: https://github.com/ArkEcosystem/desktop-wallet/commit/21ec4f25c102141c6b738050093a46ba06eb1c76#diff-cebae3f54745803997fc366f1bb91e3b ?

src/renderer/components/Wallet/WalletBusiness/WalletBusinessBridgechains.vue
```diff
@@ -122,12 +127,18 @@ export default {
...
+        if (!this.wallet_fromRoute || this.isFetching) {
+          return
+        }
 
-      this.isLoading = true
-      this.fetchBridgechains()
+        this.isLoading = true
+        this.fetchBridgechains()
+      } catch (error) {
+        throw new Error('It is not possible load bridgechain list')
```
alexbarnsley: I would log instead of throwing an error. I've experienced it where throwing an error has halted components meaning they don't work properly

brenopolanski: make sense :+1: I'll add the log

alexbarnsley:
I partly think the fetchBridgechains method shouldn't clear the screen while loading. It looks weird because of the flicker where the data clears and re-appears 🤔

PR #1821, opened 2020-03-12
feat: add badge when waiting for signature on multisig transactions

davimello28:
Display a little badge inside the multisignature menu tab when waiting for that wallet signature.

image

## Summary

Since the logic inside fetching the waiting transactions resides inside the UI component, the store is changed according to its transactions. The correct way should be either refactoring the component by creating a service for the transactions or moving the whole logic to the store. 

## Checklist

- [ ] Tests
- [x] Ready to be merged


src/renderer/store/modules/transaction.js
```diff
@@ -287,6 +292,14 @@ export default {
         networkId: rootGetters['session/profile'].networkId,
         staticFees
       })
+    },
+
+    /**
+     * Sets if there is a multisignature ttransaction waiting for this wallet signature.
```
brenopolanski: @davimello28 change from `ttransaction` to `transaction`.

src/renderer/components/Wallet/WalletTransactions/WalletTransactionsMultiSignature.js
```diff
@@ -65,6 +65,11 @@ export default {
         if (publicKey === WalletService.getPublicKeyFromWallet(this.wallet_fromRoute)) {
           this.$set(this, 'fetchedTransactions', response.transactions)
           this.totalCount = response.totalCount
+          if (this.totalCount > 0) {
```
alexbarnsley: I think the check needs improving here to only include transactions that require the wallet's signature at that point in time. I would do something like this:

```
let hasPendingTransaction = false
for (const transaction of response.transactions) {
  if (TransactionService.needsWalletSignature(transaction, WalletService.getPublicKeyFromWallet(this.wallet_fromRoute))) {
    hasPendingTransaction = true
    break
  }
}

if (hasPendingTransaction) {
  this.$store.dispatch('transaction/setWaitingMultisignature', true)
} else {
  this.$store.dispatch('transaction/setWaitingMultisignature', false)
}
```

This is used in the TransactionShowMultiSignature.vue component

davimello28: Alex, I did some little improvements. This is a computed property. 

```js
isWaitingForSignature () {
  const transactions = this.fetchedTransactions
  const wallet = WalletService.getPublicKeyFromWallet(this.wallet_fromRoute)

  function checkTransactionForWaitingSignatures (transaction) {
    return TransactionService.needsWalletSignature(
      transaction,
      wallet
    )
  }

  return transactions.some(t => checkTransactionForWaitingSignatures(t))
}
```

faustbrian:
Closing due to other changes that are coming which nullify this.

PR #1822, opened 2020-03-12
feat: fetch remote fees from node api

davimello28:
Fetch the remote fees from https://explorer.ark.io/api/node/fees


davimello28:
I'm reverting the changes made previously to solve the issue, but it seems that a better hierarchy between the fees is needed.

There are three ways that the wallet can get a fee from the node:

https://explorer.ark.io/api/node/fees 
https://explorer.ark.io/api/transactions/fees 
and the hardcoded V1 at the configuration.

The main component that makes the request for the fees can use both a request to a local host and a remote host. This makes a bit hard to understand where the fee is coming from.

alexbarnsley:
@davimello28 could you add a test to make sure the `minFee` from `feeStatistics` gets used please

davimello28:
I've refactored the tests a bit to include all the rendered fees in the same test case (case `fee > maxV1fee` and `fee < maxV1fee`) and remove hardcoded fees on those tests.

alexbarnsley:
@davimello28 tests seem to be failing

I've tested and I think the min fee change has created an unexpected issue. It should be possible to send a transaction if the fee is below the minimum - if I change it in the text field it becomes invalid and can't proceed to the next step

PR #1823, opened 2020-03-12
feat: sort menu items alphabetically

dated:
## Summary

Since the menu items are added sequentially by looping over the loaded plugins they can be out of order. Finding a specific one among multiple menu items becomes a challenging tasks, more so with the scrolling behaviour introduced in #1820.

This PR also moves the `sortByProps` file out of the components folder, where it does not strictly belong imho.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
good job, windows :trollface:

PR #1825, opened 2020-03-12
feat: allow plugins to connect to best peer

dated:
## Summary

Plugins might need to find a peer that meets certain criteria. For example a peer that doesn't use estimates in their database lookups. Currently, the only way to achieve that is by prompting the user to refresh the peer himself.

This PR exposes the `connectToBest` action from the `peer` store module as `walletApi.peers.connectToBest()` (needs the `PEER_ALL` permission) so that plugins are able to find a peer that fits the plugins needs.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


davimello28:
Is it safe? It feels weird to me that a plugin can control what I'm connected to. It seems a break in the hierarchy of the application, much like a browser selecting which network I should connect to (despite the fact that a peer can be random).

dated:
I do agree actually. At the very least it should not be possible if the current peer is a custom peer.

alexbarnsley:
@dated maybe this can be reworked so that a modal is shown for the user to confirm changing to best peer 🤔 that way it can't be used in the background because the user will be notified. Not too sure how to go about where to put that modal without looking into it more closely though

dated:
Closing this as I agree with the modal-based approach.

PR #1830, opened 2020-03-12
refactor: activate the plugin menu item when navigating to plugin route

dated:
## Summary

With these changes the plugin page menu item gets activated upon navigating to a plugin page.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


davimello28:
I couldn't make it work. I triple check the code and the branch. I'm not sure if this is the desired behavior. I can see the `active-item` in `AppSideMenu` component, but the button dont get the active style class.

Menu-activation-style-fail

dated:
menu
works as intended here 🤔

PR #1897, opened 2020-04-10
refactor: replace some lodash methods with native alternatives

brenopolanski:
## Summary

**Lodash imports removed:**

- isEmpty
- includes
- concat
- some
- keys
- flatten

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


alexbarnsley:
@brenopolanski I'm wondering if it's worth adapting this PR to use https://github.com/ArkEcosystem/utils as they're sort of the same thing 🤔

brenopolanski:
@alexbarnsley The initial idea is to remove lodash imports as much as possible and use native functions. I think there is no need for now to add this `ArkEcosystem/utils` package.

src/renderer/utils/index.js
```diff
@@ -58,6 +65,7 @@ const min = nums => {
 const upperFirst = str => `${str.charAt(0).toUpperCase()}${str.slice(1)}`
 
 export {
+  cammelToUpperSnake,
```
faustbrian, line 68:
```suggestion
  camelToUpperSnake,
```

brenopolanski: Thanks @faustbrian

PR #1936, opened 2020-05-19
feat: auto-add the recipient if loading the modal using a URI

brenopolanski:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


dated:
This problem is partially handled in #1946. The issue here is that the utility methods used all rely on the session network while we generally need to be looking at the network of the selected wallet.

PR #2012, opened 2020-06-01
feat: add `SvgIcon` component

brenopolanski:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/renderer/app/components/SvgIcon/index.ts
```diff
@@ -0,0 +1,4 @@
+import SvgIcon from "./SvgIcon.vue";
+
+export default SvgIcon;
```
faustbrian, line 3: Avoid default exports due to them being bad practice because they could be imported differently. Always require an explicit export and import unless you are forced to provide a default export by some tooling like Vue plugins.

src/renderer/app/components/SvgIcon/SvgIcon.vue
```diff
@@ -0,0 +1,29 @@
+<template>
```
faustbrian, line 1: Needs a test that ensures that this component renders accordingly to all possible property combinations.

PR #2074, opened 2020-06-09
feat: update wallet name modal

alexbarnsley:
## Summary

Also updates:
- swaps form input label Y padding from bottom to top
- flattens lang keys
- renames onClick to onClose on modal component

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


alexbarnsley:
Okay 👌 should I create a generic Modal folder under `domains/wallets/components` since there will likely be more?

faustbrian:
@alexbarnsley Try to name components more in terms of what they are doing for a human. In technical terms `ModalNameWallet` might seem fine but you will have to look at the code to figure out what it does.

Something like `ChangeWalletName` or `UpdateWalletName` makes the intent obvious and has no noise by having the `Modal` suffix in there since that doesn't really change the fact that the component is responsible for changing the name of a wallet. Could still keep the `Modal` suffix it seems unnecessary when the component name itself makes it clear what is being done.

faustbrian:
> Okay 👌 should I create a generic Modal folder under `domains/wallets/components` since there will likely be more?

Just make one folder for each component to keep things separated. They all do one specific thing and have unique styling. We should avoid those hyper generic multi-purpose components that the old wallet had where there would be 30 `if` clauses to renderer different fields and instead simply create components that have fixed styles and then we feed the data via hooks.

alexbarnsley:
All makes sense @faustbrian, I think I've covered your points

PR #2121, opened 2020-06-12
refactor: improve `CreateWallet` design

luciorubeens:
## Summary

- Fix `Alert` padding and icons
- Fix `MnemonicVerification` checkmark icon
- Fix `Circle` background when checked
- Minor tweaks.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/components/Alert/Alert.tsx
```diff
@@ -9,31 +9,60 @@ type AlertProps = {
...
+		}
+	}}
+`;
+
+const AlertIconWrapper = styled.div<{ size: string }>`
+	${({ size }) => {
+		switch (size) {
+			case "small":
+				return tw`px-4 py-2`;
+			case "default":
```
faustbrian, line 30: Could simply use `default:` instead of `case "default":` to avoid that the default needs to be explicitly specified as a size.

src/app/components/Alert/Alert.tsx
```diff
@@ -9,31 +9,60 @@ type AlertProps = {
 	size?: "small" | "default" | "large";
 };
 
-const Wrapper = styled.div(getStyles);
+const AlertContent = styled.div<{ size: string }>`
+	${({ size }) => {
+		switch (size) {
+			case "small":
+				return tw`px-6 py-4`;
+			case "default":
```
faustbrian, line 17: Could simply use `default:` instead of `case "default":` to avoid that the default needs to be explicitly specified as a size.

faustbrian:
@luciorubeens conflicts

PR #2128, opened 2020-06-13
feat: add `<Checkbox>` component

brenopolanski:
## Summary

Screenshot from 2020-06-13 08-47-03

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/components/Checkbox/style.ts
```diff
@@ -0,0 +1,30 @@
+import tw, { css } from "twin.macro";
+
+const baseStyle = [tw`form-checkbox h-4 w-4 transition duration-150 ease-in-out`];
+
+const getColor = (color: string): any => {
+	const colorBase = `var(--theme-color-${color})`;
+
+	switch (color) {
+		case "primary":
```
faustbrian, line 9: The `case "primary":` case could probably be treated as `default:` because primary and default are pretty much the same thing in this context. One thing less to define repeatedly when you use the component as it will be the most common choice.

src/app/components/Checkbox/style.ts
```diff
@@ -0,0 +1,30 @@
+import tw, { css } from "twin.macro";
+
+const baseStyle = [tw`form-checkbox h-4 w-4 transition duration-150 ease-in-out`];
+
+const getColor = (color: string): any => {
```
faustbrian, line 5: This whole function could probably be simplified to something like the below to avoid the repetitive `switch` statement.

```ts
const getColor = (color: string): any => {
	if (!["primary", "success", "danger", "warning"].includes(color)) {
		throw new Error(`Failed to find a color for "${color}"`);
	}

	const colorBase = `var(--theme-color-${color})`;

	return css`
		color: ${colorBase};
	`;
};
```

brenopolanski: Thanks for the tip!

I made the code clearer doing:

```
const getColor = (color: string = "primary"): any => {
	const colorBase = `var(--theme-color-${color})`;

	return css`
		color: ${colorBase};
	`;
};
```

brenopolanski: @faustbrian  If we are going to add this code snippet:

```
if (!["primary", "success", "danger", "warning"].includes(color)) {
	throw new Error(`Failed to find a color for: "${color}"`);
}
```

We should create a `<ErrorBoundary>` component.

- https://reactjs.org/docs/error-boundaries.html
- https://github.com/testing-library/react-testing-library/issues/400#issuecomment-509698450
- https://github.com/JoschuaSchneider/use-error-boundary

brenopolanski: @faustbrian I added this package https://github.com/bvaughn/react-error-boundary, we will use it more throughout development to create fallback errors.

I didn't add the code:

```
if (!color) {
	color = "primary";
}
```

Because there is a `defaultProps` on the component:

```
Checkbox.defaultProps = {
	type: "checkbox",
	color: "success",
};
```

src/app/components/Checkbox/__tests__/Checkbox.test.tsx
```diff
@@ -1,15 +1,40 @@
 import React from "react";
 import { render } from "@testing-library/react";
-
+import { ErrorBoundary } from "react-error-boundary";
 import { Checkbox } from "../Checkbox";
 
 describe("Checkbox", () => {
+	beforeAll(() => {
+		jest.spyOn(console, "error").mockImplementation(() => null);
```
faustbrian, line 8: Why was this added back?

brenopolanski: As I'm forcing an error in the test, it will show in the terminal. So `jest.spyOn(console, "error").mockImplementation(() => null)` hide it.

faustbrian: Wouldn't that also end up swallowing legitimate errors that might be logged by some dependencies or yarn and npm itself since they use `console.*`?

brenopolanski: Yeah, true. I'll add:

```
afterAll(() => {
  console.error.mockRestore()
})
```

and  it works again

faustbrian: That would still swallow legitimate errors while the tests run. I'll merge it for now and look at it later but these kinds of things should be avoided because they make debugging more difficult because everything in the CLI with JavaScript makes use of `console.*` to display messages and if something acts out or starts to break after a dependency update you would have to remove the mock for the logging, which isn't the first thought for most people.

PR #2144, opened 2020-06-16
feat: add size props to modal component

brenopolanski:
## Summary

Added size `"sm" | "md" | "lg" | "xl" | "2xl" | "3xl"`

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/components/Modal/Modal.tsx
```diff
@@ -18,14 +19,33 @@ type ModalContentProps = {
...
 	description?: string;
 	image?: React.ReactNode;
+	size?: "small" | "medium" | "large" | "xlarge" | "default";
 	onClose?: any;
 };
 
+const ModalContainer = styled.div<{ size: string }>`
+	${({ size }) => {
+		switch (size) {
+			case "small":
```
faustbrian, line 29: Maybe we should stick to the same naming as the classes itself. Meaning `sm/md/lg/xl` because that is what people already know from Tailwind.

brenopolanski: Yeah, I agree

faustbrian: There seem to be a few other components where the same naming conventions could be applied.

src/app/components/Modal/Modal.tsx
```diff
@@ -18,14 +19,33 @@ type ModalContentProps = {
 	title: string;
 	description?: string;
 	image?: React.ReactNode;
+	size?: "sm" | "md" | "lg" | "xl" | "default";
```
faustbrian, line 22: Could we completely omit the `"default"` value and just allow null? Since we have the `default` case it seems redundant to keep this value.

src/app/components/Modal/Modal.tsx
```diff
@@ -18,14 +19,33 @@ type ModalContentProps = {
...
 	onClose?: any;
 };
 
+const ModalContainer = styled.div<{ size: string }>`
+	${({ size }) => {
+		switch (size) {
```
faustbrian, line 28: Now that you follow the TailwindCSS naming conventions you can simplify this method to something like the below which will make it shorter and more readable.

```ts
const ModalContainer = styled.div<{ size: string }>`
    ${({ size }) => {
        const allowedSizes: string[] = ["sm", "md", "lg", "xl"];

        if (allowedSizes.includes(size)) {
            return tw`max-w-${size}`;
        }
        
        return tw`max-w-2xl`;
    }
`;
```

faustbrian: You could also consider creating a method to handle those size behaviours across multiple components, maybe something like the below.

```ts
const handleComponentSize = (size?: string | undefined, allowedSizes: string[], defaultSize: string, callback: Callable) => {
    if (!size) {
        return defaultSize;
    }

    if (allowedSizes.includes(size)) {
        return callback(size);
    }
    
    throw new Error(`Failed to find ${size} in ${JSON.stringify(allowedSizes)}`)
}
```

```ts
const ModalContainer = styled.div<{ size: string }>`
    ${({ size }) => handleComponentSize({
        size,
        allowedSizes: ["sm", "md", "lg", "xl"],
        defaultSize: tw`max-w-2xl`,
        callback: size => tw`max-w-${size}`
    });
`;

const ModalContainer = styled.div<{ size: string }>`
    ${({ size }) => handleComponentSize({
        size,
        allowedSizes: ["sm", "md", "lg", "xl"],
        defaultSize: tw`max-w-2xl`,
        callback: size => tw`max-w-${size}`
    });
`;
```

This would remove a lot of the boilerplate that is duplicated across components that make decisions about sizes. Another approach would be to use objects like in the example below.

**Before**
```ts
const getSize = (size: string): any => {
	switch (size) {
		case "small":
			return tw`w-8 h-8 px-2 py-1 text-sm`;
		case "large":
			return tw`w-11 h-11 px-2 py-1 text-sm`;
		default:
			return tw`w-10 h-10 px-4 py-2 text-base`;
	}
};
```

**After**
```ts
const getSize = (size: string): any => {
    if (!size) {
        return tw`w-10 h-10 px-4 py-2 text-base`;
    }

	return {
		"sm": tw`w-8 h-8 px-2 py-1 text-sm`,
		"lg": tw`w-11 h-11 px-2 py-1 text-sm`,
	}[size];
};
```

This approach makes the code more readable by using an early exit if the size is empty, which means we want to use the default, and then uses an object to simply map the sizes to their styles. This example obviously misses some error handling but you get the idea.

faustbrian:
@brenopolanski conflicts

src/app/components/Modal/Modal.tsx
brenopolanski: Perhaps `handleComponentSize` can be more generic? I mean, just change the name to something like `handleComponentVariations`? So it can work with:

- `color?: "primary" | "success" | "danger" | "warning" | "neutral";`
- `variant?: "solid" | "plain" | "outline";`
- `size?: "small" | "default" | "large" | "icon";`

faustbrian: Yeah it could be more generic to handle other scenarios too but if it gets too abstract it's not worth the effort because it will add more complexity than it removes.

brenopolanski: > Now that you follow the TailwindCSS naming conventions you can simplify this method to something like the below which will make it shorter and more readable.
> 
> 
> const ModalContainer = styled.div<{ size: string }>`
>     ${({ size }) => {
>         const allowedSizes: string[] = ["sm", "md", "lg", "xl"];
> 
>         if (allowedSizes.includes(size)) {
>             return tw`max-w-${size}`;
>         }
>         
>         return tw`max-w-2xl`;
>     }
> `;

Okay. i'll make the changes using your suggestion above and then finish the "Install Plugin Modal". After that, i'll send a separate PR to `handleComponentSize | handleComponentVariations`.

brenopolanski: I'm having problems with string interpolation inside `tw`. There is an open issue about this: https://github.com/ben-rogerson/twin.macro/issues/17

brenopolanski: @faustbrian i removed the `default` option and added `3xl` size. For now we can use switch statement. I did a test case for `If-else vs. switch vs. object`: https://jsperf.com/if-else-vs-switch-vs-object

faustbrian:
@brenopolanski conflicts

src/app/components/Modal/Modal.tsx
faustbrian: > @faustbrian i removed the `default` option and added `3xl` size. For now we can use switch statement. I did a test case for `If-else vs. switch vs. object`: [jsperf.com/if-else-vs-switch-vs-object](https://jsperf.com/if-else-vs-switch-vs-object)

It's not about performance because that hardly matters in a case like this but it's about readability and duplication.

brenopolanski: should i change to:

```
return {
  "sm": tw`max-w-sm`,
  ...
  "lg": tw`max-w-lg`,
}[size];
```

?

PR #2148, opened 2020-06-16
feat: add compact variant to `TransactionListItem`

goga-m:
## Summary
Add `compact` variant to `TransactionListItem` component. Uses same data fields as `default` variant.

Seen in Notifications transactions lists.

Preview:

2020-06-16-151729_705x161_scrot

Usage: 
```jsx
<TransactionListItem variant="compact" {...data}></TransactionListItem>
```

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/app/components/TransactionListItem/TransactionListItem.tsx
+++ b/src/app/components/TransactionListItem/TransactionListItem.tsx
@@ -40,7 +40,7 @@ export const TransactionListItem = ({
 		return (
 			<tr
 				onClick={onTxClick}
-				className="border-b cursor-pointer border-theme-neutral-200"
+				className="border-b border-theme-neutral-200 cursor-pointer"
 				data-testid="transaction__row"
 			>
 				<td className="w-20 py-4 mt-1">
@@ -64,7 +64,7 @@ export const TransactionListItem = ({
 	return (
 		<tr
 			onClick={onTxClick}
-			className="border-b cursor-pointer border-theme-neutral-200"
+			className="border-b border-theme-neutral-200 cursor-pointer"
 			data-testid="transaction__row"
 		>
 			<td className="w-48 py-1 text-sm text-theme-neutral-600"> {date} </td>
@@ -92,5 +92,4 @@ export const TransactionListItem = ({
 TransactionListItem.defaultProps = {
 	walletTypeIcons: [],
 	actions: [],
-	variant: "default",
 };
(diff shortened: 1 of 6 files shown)


src/app/components/TransactionListItem/TransactionListItem.tsx
```diff
@@ -4,14 +4,16 @@ import { Icon } from "app/components/Icon";
...
-type TransactionListItemProps = {
+export type TransactionListItemProps = {
 	date: string;
 	avatarId: string;
 	type: string;
 	address?: string;
 	walletName?: string;
 	amount: string;
 	fiat: string;
+	variant?: "default" | "compact";
```
faustbrian, line 15: Will there be more variants or could the `compact` style simply be a boolean property?

src/app/components/TransactionListItem/TransactionListItem.tsx
```diff
@@ -4,14 +4,16 @@ import { Icon } from "app/components/Icon";
 import { Label } from "app/components/Label";
 import React from "react";
 
-type TransactionListItemProps = {
+export type TransactionListItemProps = {
```
faustbrian, line 7: Any reason for this being exported?

src/app/components/TransactionListItem/TransactionListItem.tsx
goga-m: There might be designs that need some variation of a transaction list item. I didn't make it boolean so that it would be easier to extend in such case.

src/app/components/TransactionListItem/TransactionListItem.tsx
goga-m: This is going to be required in `Notifications` component, [here](https://github.com/ArkEcosystem/desktop-wallet/commit/e1270220354f88e0c359bcfa26d927a5610afaf9#diff-d07d23621008e0910c251c32bca4d76cR22)

Is it better to add it in models.ts?

faustbrian: Yep, that should go into the models file in that case.

src/app/components/TransactionListItem/TransactionListItem.tsx
faustbrian: Then the `"default"` option should be removed because no the argument is optional and no value should be treated as default.

goga-m: You are absolutely right. Not necessary at all. Fixing it.

goga-m:
My bad. Undoing the style guide violation fixes :)

PR #2153, opened 2020-06-16
feat: expose environment

### LINKED ISSUE #2065: Expose `Environment` from `platform-sdk-profiles` through Context API
The `Environment` class from the Platform SDK is responsible for managing all data and profiles. This means we need access to it basically everywhere in the components to access the current profile, wallets, app settings and so on.

Components could access it via https://reactjs.org/docs/context.html but for classes and functions outside of components the instance might need to be shared differently. Though ideally we would avoid having to pass around instances that are persisted in some cache and rather pass the instances or specific data to methods if they need it which prevents unwanted mutations an side-effects.

## Related

```ts
import { Environment } from "@arkecosystem/platform-sdk-profiles";

import { httpClient } from "./http-client";

const env: Environment = new Environment({ httpClient, storage: "indexeddb" });

export default {
	install(Vue) {
		Vue.prototype.$env = env;
		Vue.prototype.$profiles = env.profiles();
	},
};
```

```ts
import { HttpClient } from "@/support/services/http";

export const httpClient = new HttpClient();

export default {
	install(Vue) {
		Vue.prototype.$http = httpClient;
	},
};
```

```ts
import { Contracts } from "@arkecosystem/platform-sdk";
import { Reqwest } from "@kodekeep/reqwest";

export class HttpClient implements Contracts.HttpClient {
	readonly #client: Reqwest;

	public constructor() {
		this.#client = new Reqwest();

		this.#client.retry(0, 0);
	}

	public timeout(seconds: number): HttpClient {
		this.#client.timeout(seconds);

		return this;
	}

	public withoutEncoding(): HttpClient {
		this.#client.withOptions({ responseType: "buffer" });

		return this;
	}

	public async get(path: string, query: object = {}): Promise<Record<string, any>> {
		return (await this.#client.get(path, query)).throw().json();
	}

	public async post(path: string, body: object, headers: object = {}) {
		return (await this.#client.withHeaders(headers).post(path, body)).throw().json();
	}
}
```

clucasalcantara:
## Summary
Resolves #2065 

Adds `profiles` from `platform-sdk`. 

A few changes are made to make this integration work since `platform-sdk` relies on the ledger lib some workarounds were needed. The major impact will be on the speed to install any library since the `post-install` script will run and before a `yarn dev` when the `electron-rebuild` will be needed to match versions and references.

If you have any problems running tests from a `yarn test` cmd after a `yarn dev, just run `npm rebuild` or a `yarn` to refresh modules refs.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
@clucasalcantara conflicts and e2e is red

faustbrian:
Closing in favour of https://github.com/ArkEcosystem/desktop-wallet/pull/2159. we will, later on, need some things from here to get ledger integrated but we will work on that once the time comes.

PR #2234, opened 2020-06-23
feat: add `RecipientList` component

goga-m:
## Summary
Extract Recipient list as a separate components under domains/transaction.
Appears in the transaction flow, in 1st step (transaction-send form, editable with buttons) and read-only on transaction overview step.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


clucasalcantara:
@goga-m 🌮

PR #2243, opened 2020-06-23
feat: add My Contacts page

dated:
## Summary

Adds the My Contact page.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
@dated failing tests

src/app/components/Address/Address.tsx
```diff
@@ -38,7 +38,7 @@ export const Address = ({ address, addressClass, walletName, maxChars, size }: P
 					addressClass || (walletName ? "text-theme-neutral-400" : "text-theme-neutral-800")
 				} font-semibold ${size && fontSizes[size]}`}
 			>
-				{truncateStringMiddle(address, maxChars)}
+				{maxChars && maxChars > 0 ? truncateStringMiddle(address, maxChars) : address}
```
faustbrian, line 41: Should the `maxChars && maxChars > 0` part be handled inside `truncateStringMiddle` so that it would return the `value` without any truncating if it is false? Then it would apply to all places that use `truncateStringMiddle` and not just this place.

dated: https://github.com/ArkEcosystem/desktop-wallet/pull/2243/commits/16e6fadd84197039458c7cf51a1b1d38f14a69e0 that ok?

PR #2297, opened 2020-06-30
refactor: add dynamic tooltip to Clipboard component

dated:
## Summary

Adds a dynamic tooltip to the `Clipboard` component.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


src/app/components/Clipboard/Clipboard.test.tsx
```diff
@@ -33,6 +35,30 @@ describe("Clipboard", () => {
...
+				</Clipboard>
+			</I18nextProvider>,
+		);
+
+		act(() => {
+			fireEvent.mouseEnter(getByTestId("clipboard__wrapper"));
+		});
+
+		expect(baseElement).toHaveTextContent("Copy to clipboard");
+		expect(baseElement).not.toHaveTextContent("Copied!");
```
faustbrian, line 52: Should probably use the `CLIPBOARD.COPIED` key here instead of hardcoding text, same for the `Copy to clipboard`.

dated: I think there's multiple occurrences of such checks across the test suite, probably worthy of their own PR. I'll fix this once i'm back at the desk.

PR #2306, opened 2020-06-30
feat: add i18n wrapper for unit tests

dated:
## Summary

Overrides the `render` method of the testing-library to wrap the tested component in a `I18nextProvider`.

Instead of importing from `@testing-library/react` you should now import from `test-utils` instead, e.g.

```js
// don't
import { fireEvent, render } from "@testing-library/react";

// do
import { fireEvent, render } from "test-utils";
```

Also adds `tsx` and removes `vue` from the prettier config.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


faustbrian:
@dated conflicts

PR #2314, opened 2020-07-01
fix: ark uri scheme

### LINKED ISSUE #2286: ark: uri scheme specifying address but no amount no longer works in 2.9.2
In 2.9.1 and earlier you could specify an `ark:` URI like `ark:AUexKjGtgsSpVzPLs6jNMM6vJ6znEVTQWK` which would pre-fill the modal with the address to send to, while allowing the user to specify the amount to send.

This is broken in 2.9.2 as it now brings up a default modal with nothing pre-filled, no address, nada. If you specify the amount then it works e.g. `ark:AUexKjGtgsSpVzPLs6jNMM6vJ6znEVTQWK?amount=1` but that is no good for situations where you don't know how much the user wants to send.

I've received complaints about this new unexpected behaviour from people playing the PVP wallet games since they can no longer start a new game easily, since it uses the former `ark:` URI structure i.e. an address with no amount, as it is supposed to pre-fill the game's wallet address in the modal but allow the user to enter their own desired amount so they can pick their own wager. This new unanticipated behaviour now causes a terrible user experience.

brenopolanski:
## Summary

Fixes #2286 

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
+++ b/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
@@ -573,6 +573,8 @@ export default {
               this.wallet_truncate(this.schema.address)
             ]))
           }
+        } else {
+          this.$set(this, 'recipientId', this.schema.address || '')
         }
 
         this.$set(this.form, 'vendorField', this.schema.vendorField || '')
--- a/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
+++ b/src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
@@ -560,8 +560,8 @@ export default {
       }
 
       try {
-        if (this.schema.address && this.schema.amount) {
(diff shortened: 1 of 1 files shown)


src/renderer/components/Transaction/TransactionForm/TransactionFormTransfer.vue
```diff
@@ -573,6 +573,8 @@ export default {
               this.wallet_truncate(this.schema.address)
             ]))
           }
+        } else {
```
alexbarnsley, line 576: This bypasses the `validateAddress` check. You'll be better just to move the amount check I think

brenopolanski: It check first if there is a valid address. Then, if there was an `address` and `amount` it must be added to `recipients`, otherwise fill in the `recipientId` input.

brenopolanski:
@samharperpittam can you test please?

samharperpittam:
> @samharperpittam can you test please?

on it

PR #2331, opened 2020-07-01
feat: add app wrapper

clucasalcantara:
## Summary
- Add app wrapper to share navbar between pages routes

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+				{location.pathname !== "/" && (
+					<NavigationBar currencyIcon="Ark" balance="34,253.75" userInitials="IO" />
+				)}
+				{renderRoutes(routes)}
+			</EnvironmentProvider>
+		</main>
+	</I18nextProvider>
+);
+
+export default withRouter(App);
```
faustbrian, line 28: As mentioned before, avoid `export default` unless necessary for some specific reason.

clucasalcantara: Yep, already updated, it was just for testing and I've used the component template from vs code :P

PR #2352, opened 2020-07-02
feat: add navbar routing

clucasalcantara:
## Summary
Depends on #2345 and #2347 merged to be synchronized since they all are originated from the same branch and could contain duplicated implementations

- Adds nav bar routing
- Adds routing mapping for settings, plugins, help, exchange, and contacts
- Add e2e to navbar routing
- Adjust settings page to don't depend on props
- Adjust navbar routes

## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


src/domains/profile/pages/Welcome/Welcome.tsx
```diff
@@ -52,7 +53,11 @@ export const Welcome = () => {
 							<div className="mt-6 mb-8 space-y-3">
 								{profiles.map((profile: any, index: number) => (
 									<ProfileCard
-										handleClick={() => history.push(`/portfolio/${profile.id()}`)}
+										handleClick={() => {
+											updateAppState("activeProfile", profile);
+
+											return history.push(`/profiles/${profile.id()}`);
```
luciorubeens, line 59: I think we should avoid this pattern, as it can result in duplicate/out-of-sync values by our experience with Mobile Wallet, making it a router-based application by design and as a single point of truth may be better to maintain and debug as we do in the Explorer and Desktop Wallet v2, also useful for paginations, deep linking and such.

This is probably all we need when accessing the endpoint `/profile/1/wallet/AXzxJ8Ts3dQ2bvBR1tPE7GUee9iSEJb8HX`:

```ts
import React from "react";
import { userParams } from "react-router-dom";
import { EnvironmentContext } from "app/contexts/Environment";

export function useActiveProfile() {
  const env = React.useContext(EnvironmentContext);
  const { profileId } = useParams();

  return env.profiles().get(profileId);
}

export function useActiveWallet() {
  const profile = useActiveProfile();
  const { address } = useParams();
  return profile.wallets().findByAddress(address);
}

export const ProfilePage() {
  const activeProfile = useActiveProfile();
  
  return <h1>{activeProfile.name()}</h1>
}

export const WalletPage() {
  const activeWallet = useActiveWallet();
  
  return <h1>{activeWallet.balance()}</h1>
}
```

clucasalcantara: Agreed @luciorubeens! Nice point, but, this don’t solve the problem with the NavBar since she is outside the route (injected with wrapper) we can’t access the profile Id param there (arrives undefined, tried with useParams, match, location) 🤔

This was mostly to give NavBar knowledge about the active profile since we need to build their nav links with the active profile id 🤔🙃

Gonna think in another solution to have it there tho, ideas are welcome 😬

PR #2377, opened 2020-07-06
feat: add `RouterView` component to render default layout

luciorubeens:
## Summary

Proposal to add router parameters to the Navbar.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged

Missing tests that should be fixed in the base branch.


src/app/hooks/env.ts
```diff
@@ -1,26 +1,34 @@
...
 
-	try {
-		return profile.wallets().findByAddress(address);
-	} catch {
-		return undefined;
-	}
+	return React.useMemo(() => {
+		if (profile) {
+			try {
+				return profile.wallets().findByAddress(address);
```
faustbrian, line 28:
```suggestion
				return profile.wallets().findById(id);
```

PR #2381, opened 2020-07-06
refactor: reduce `Toggle` switch size

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/app/components/Toggle/Toggle.tsx
+++ b/src/app/components/Toggle/Toggle.tsx
@@ -6,11 +6,15 @@ const Input = styled.input`
 `;
 
 const Handle = styled.div`
-	${tw`inline-flex rounded-full relative h-2 w-10 cursor-pointer bg-theme-neutral-light`}
+	height: 7px;
+	${tw`inline-flex rounded-full relative w-9 cursor-pointer bg-theme-neutral-light`}
 `;
 
 const HandleInner = styled.span`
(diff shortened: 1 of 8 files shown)


goga-m:
@samharperpittam could you review and confirm that the new size looks ok?
Thanks!

src/app/components/Toggle/Toggle.tsx
```diff
@@ -6,11 +6,15 @@ const Input = styled.input`
 `;
 
 const Handle = styled.div`
-	${tw`inline-flex rounded-full relative h-2 w-10 cursor-pointer bg-theme-neutral-light`}
+	height: 7px;
```
brenopolanski, line 9: Maybe use `rem` instead of `px` :hmhm:? Tailwind uses `rem` for all types of sizes.

alexbarnsley: I would say using px here is fine since tailwind doesn't have a size for 7px, and using it once doesn't warrant it being added to the tailwind config 🤔

brenopolanski: You're right! :+1:

goga-m: Indeed, 7px seemed just the right size to make the placement of the handle align with the reduced  gray line and as Alex mentioned it wasn't possible with tw classes.

Thanks for reviewing!

PR #2460, opened 2020-07-14
fix: update `InputRange` to check if the value is greater than the max

brenopolanski:
## Summary

This PR fix this behavior below:

Screenshot from 2020-07-14 07-46-39

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/components/Input/InputRange.tsx
```diff
@@ -22,7 +22,12 @@ export const InputRange = React.forwardRef<HTMLInputElement, Props>(
 		const handleInput = (value: string) => {
 			const fraction = Math.pow(10, magnitude! * -1);
 			const amount = BigNumber.make(value).times(fraction);
-			setValues([amount.toNumber()]);
+
+			if (amount.isGreaterThan(max)) {
+				setValues([max]);
+			} else {
+				setValues([amount.toNumber()]);
```
clucasalcantara, line 29: Hi, @brenopolanski !

Nice catch 🌮  I just think that we don't need this else block over here, you could simplify this complexity by returning the setValues with max inside the if and return the set values with the amount if it isn't greater than like:

```
if (amount.isGreaterThan(max)) return setValues([max]);

return setValues([amount.toNumber()];
```

brenopolanski: @clucasalcantara Don't you think it's a little strange to return 'setValues`? I mean, because `setValues` updates only the state `values`. What do you think @luciorubeens ?

clucasalcantara: Why strange? The return is only to jump out the function when the branch decided to be executed is found (is greater than or not) but I can be wrong ¯\_(ツ)_/¯. I just think that we should add cyclomatic complexity in functions when we can use a direct execution, which is the same approach when you use a return inside a `switch` statement to avoid the usage of `break`.

brenopolanski: ahh @clucasalcantara  i understand now what do you mean :+1:

src/app/components/Input/InputRange.tsx
```diff
@@ -22,7 +22,12 @@ export const InputRange = React.forwardRef<HTMLInputElement, Props>(
 		const handleInput = (value: string) => {
 			const fraction = Math.pow(10, magnitude! * -1);
 			const amount = BigNumber.make(value).times(fraction);
-			setValues([amount.toNumber()]);
+
+			if (amount.isGreaterThan(max)) {
+				return setValues([max]);
```
luciorubeens, line 27: It works as expected, however, automatically changing the value may be unexpected behavior for the user, it is best handled with validation. We can try something like:

```
const rangeValues = [Math.min(values[0], 100)];
<Range values={rangeValues} />
```

brenopolanski: Good trick and the code is cleaner @luciorubeens :ok_hand:

I removed the code with `isGreaterThan()` and added:

```
const handleInput = (value: string) => {
	const fraction = Math.pow(10, magnitude! * -1);
	const amount = BigNumber.make(value).times(fraction);
	setValues([amount.toNumber()]);
};

const rangeValues = [Math.min(values[0], max)];

<InputCurrency value={rangeValues[0]} />

<Range values={rangeValues} />
```

src/app/components/Input/InputRange.tsx
```diff
@@ -45,7 +41,7 @@ export const InputRange = React.forwardRef<HTMLInputElement, Props>(
 					}}
 					magnitude={magnitude}
 					type="text"
-					value={values[0]}
+					value={rangeValues[0]}
```
luciorubeens, line 44: Just leave `values[0]` here, if the user sets above the `max` a validation message should appear (handled by `useForm`) and not change the value he entered. This is how the `input[type=number]` works.

brenopolanski: done @luciorubeens :+1:

PR #2461, opened 2020-07-14
feat: implement profile creating

alexbarnsley:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


alexbarnsley:
@faustbrian can you check and make sure I've understood the SDK side of it please. I also wasn't sure how to access the env/profiles within the second test to make sure it was still created

alexbarnsley:
the merge has broken shit. i'll need to fix it tonight

src/domains/profile/pages/CreateProfile/CreateProfile.tsx
```diff
@@ -61,13 +64,26 @@ export const CreateProfile = ({ onSubmit }: CreateProfileProps) => {
...
 
+	const submitForm = async ({ name, currency, isDarkMode, marketProvider }: any) => {
+		const profile = env.profiles().create(name);
+		profile.settings().set(ProfileSetting.MarketProvider, marketProvider);
+		profile.settings().set(ProfileSetting.ChartCurrency, currency);
+		profile.settings().set(ProfileSetting.Theme, isDarkMode ? "dark" : "light");
+
+		await env.persist();
+
+		if (typeof onSubmit === "function") {
```
clucasalcantara, line 82: Hi, @alexbarnsley! Just a small doubt :3

Do we really need this `typeof` checking over here? Since we are using TS isn't better to add the right type for the submit (function) and expect to have it otherwise it'll break in compilation time? 🤔

alexbarnsley: Added this more because onSubmit is nullable. I'll change it.

src/domains/profile/pages/CreateProfile/CreateProfile.tsx
```diff
@@ -61,13 +64,26 @@ export const CreateProfile = ({ onSubmit }: CreateProfileProps) => {
...
 
+	const submitForm = async ({ name, currency, isDarkMode, marketProvider }: any) => {
+		const profile = env.profiles().create(name);
+		profile.settings().set(ProfileSetting.MarketProvider, marketProvider);
+		profile.settings().set(ProfileSetting.ChartCurrency, currency);
+		profile.settings().set(ProfileSetting.Theme, isDarkMode ? "dark" : "light");
+
+		await env.persist();
+
+		if (onSubmit) {
```
faustbrian, line 82: Why is that check needed, shouldn't it always be defined because something would go wrong if it isn't?

clucasalcantara: Agreed, that what I was talking when I questioned about the `typeof`, ideally the submit prop shouldn't be required?

faustbrian:
> @faustbrian can you check and make sure I've understood the SDK side of it please. I also wasn't sure how to access the env/profiles within the second test to make sure it was still created

What test do you mean, I assume https://github.com/ArkEcosystem/desktop-wallet/pull/2461/files#diff-b0f51b717f96d11aff34382b99e06356R80?

alexbarnsley:
@faustbrian that's redundant now if I make the callback required

alexbarnsley:
@faustbrian that callback is now required

src/domains/profile/pages/CreateProfile/CreateProfile.tsx
alexbarnsley: It's removed now

Issue #1605, opened 2020-01-17
Desktop Wallet doesn't work on Linux (Debian/Ubuntu)


davimello28:
This can be caused by different problems and some of them are hard to debug. You can use [Debugtron](https://github.com/bytedance/debugtron) to inspect your application after building it. 

## 1) Dependencies for USB are not installed. 

As specified on the `README.md`, you should install those packages. This can be identified as soon as you try to `npm install` the package.

```sh
sudo apt-get install libudev-dev libusb-1.0-0-dev
```

## 2) BCrypto doesn't include the correct symbols 

**This fails silently when building, even with all the env variables and verbose modes enabled.** You can identify this problem when running `npm run dev`.

When running a Linux environment, `bcrypto` can mistakenly use the OpenSSL symbols included on the OS instead of the included `/src` folder. This removes EC capabilities of the compiled version. 

To solve this issue, you need to access `node_modules/bcrypto` manually and force an installation by using `npm install`. 

```sh
cd node_modules/bcrypto && npm install
```

## 3) chrome-sandbox isn't owned by root or doesn't have 4755 permission.

This happens on systems that run with kernels without `kernel.unprivileged_userns_clone = 1`, like Debian. (Read more about [it here](https://security.stackexchange.com/questions/209529/what-does-enabling-kernel-unprivileged-userns-clone-do)).

Since it is not desirable that [npm asks for root permissions](https://github.com/electron/electron/issues/17972)

You have three ways of solving this problem: 

1. Change the ownership and the permissions of `chrome-sandbox`. This is easier to do if you're using the `linux-unpacked` binaries directly. 

```sh
cd build/target/linux-unpacked
sudo chown root chrome-sandbox 
sudo chmod 4755 chrome-sandbox
```

2. **Not recommended**. Change the kernel to enable creating user namespaces (with [some security downsides](https://lwn.net/Articles/543442/)). 

```sh
echo kernel.unprivileged_userns_clone=1 >> /etc/sysctl.conf
```

3. **Not recommended**. You can disable `chrome-sandbox` entirely passing the `--no-sandbox` tag to the application. This exposes both the application and the host to malicious code.

```sh
cd build/target/linux-unpacked && ark-desktop-wallet --no-sandbox
```

## 4) There is more than one instance running. 

The app won't run if there is more than one instance running. If you tried running the application without checking the previous topics, chances are that you still have the application running in the background (without the GUI). 

[...]

PR #2527, opened 2020-07-25
Update package.json

NayiemWillems:
Issue #1605 

Wallet doesn't compile in linux environment and returns error with the "yarn dev" script.
bcrypto works flawlessly on Node console, but when executing inside Electron some problems arise.

Installing bcrypto directly inside the node_modules directly solves this issue for now.

## Summary

Added Yarn script in package.json

Makes life easier for those that work in a linux environment.

Single line added. Installs bcrypto from node_modules directory.

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


brenopolanski:
Hey @NayiemWillems

Instead of adding a new script, you can add a note in the [README](https://github.com/ArkEcosystem/desktop-wallet/blob/develop/README.md) with information about the problem and the command to install it: `cd node_modules/bcrypto && npm install`.

NayiemWillems:
Sure, that's fine.

PR #2532, opened 2020-07-27
refactor: sdk wallet integration into components

alexbarnsley:
## Summary

Changes components that branch from the Dashboard Wallet list component. Took a while with a lot of changes so I think it may be messy in areas

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


alexbarnsley:
This PR is pretty much ready functionality-wise (waiting on #2570 to properly test though).

What isn't ready is the memory limit where tests are now failing, I think because of the extra tests/additional functionality. I'm kind of stuck now, probably because of how large I made this PR 

I was debugging and noticed that if I invalidate the `jest.setup.js` file (e.g. import a package that doesn't exist), all tests fail to run as expected, but I still get the Out of Memory error at the end (see image below). I didn't expect this as none of the tests are running, only the imports. 

image

jest.setup.js:
```
import { env } from "./src/utils/testing-library";
import invalidPackage from "invalid-package";
```

I think it could be [this](https://github.com/ArkEcosystem/desktop-wallet/blob/3.0-react/src/utils/testing-library.tsx#L15) line in `utils/testing-library` since removing it stops the out of memory, but it could just be a coincidence.

@clucasalcantara @luciorubeens @goga-m @brenopolanski if anyone has any ideas, that would be great! I think I'm just clutching at straws now.

TL;DR I'm stuck

alexbarnsley:
@faustbrian I'm aware of the conflicts - sorting them along with fixing tests based on changes that were made before pushing.

goga-m:
@alexbarnsley  tried the case with invalid package in `3.0-react` as well (that has the memory fix) and reproduced Out of Memory error. I assume that this would be caused as a side-effect from --runInBand and gc not running properly on thrown errors. Running it without --runInBand seems to go through the errors without hitting memory limit.

PR #2540, opened 2020-07-28
chore: fix table classes

goga-m:
## Summary
Use empty string when column className is not provided.
Removes 'undefined` class names.

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/app/components/Table/Table.tsx
+++ b/src/app/components/Table/Table.tsx
@@ -60,7 +60,11 @@ export const Table = ({ children, data, columns, hideHeader }: TableProps) => {
 										data-testid={`table__th--${thIndex}`}
 										{...column.getHeaderProps(column.getSortByToggleProps())}
 									>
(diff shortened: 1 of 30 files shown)


src/app/components/Table/Table.tsx
```diff
@@ -60,7 +60,11 @@ export const Table = ({ children, data, columns, hideHeader }: TableProps) => {
 										data-testid={`table__th--${thIndex}`}
 										{...column.getHeaderProps(column.getSortByToggleProps())}
 									>
-										<div className={`flex flex-inline align-top ${column.className}`}>
+										<div
+											className={`flex flex-inline align-top ${
+												column.className ? column.className : ""
```
faustbrian, line 65:
```suggestion
												column.className || ''
```

goga-m: Me rushing :-)

PR #2555, opened 2020-07-29
fix: trigger advanced mode confirmation and use defaults

brenopolanski:
## Summary

- Update title to `Profile Settings`
- Trigger advanced mode modal
- Populate profile settings with data

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


brenopolanski:
@faustbrian i will add it

src/domains/setting/pages/Settings/available-settings/General.tsx
```diff
@@ -266,7 +266,9 @@ export const General = ({ env, formConfig, onSubmit }: GeneralProps) => {
 										}).toString(),
 									})}
 									options={PlatformSdkChoices.passphraseLanguages}
-									defaultValue={activeProfile?.settings().get(ProfileSetting.Bip39Locale)}
+									defaultValue={
+										activeProfile?.settings().get(ProfileSetting.Bip39Locale) || "english"
```
faustbrian, line 270: Those defaults shouldn't be handled here, they should be set when a profile is created.

The issue I reported with the defaults/values not being filled happens when you save your settings and reload the wallet. The values exist but the `Select` doesn't seem to be using them. Testing this requires to change the storage driver to `indexeddb` because the `yarn dev` command will use an in-memory storage without persistence.

brenopolanski: ahhh make sense now.. i saw that you added these default values here: https://github.com/ArkEcosystem/desktop-wallet/commit/110939c0cfdca5d19f75e7cd97a130c00d83502b

I will remove my default values

PR #2656, opened 2020-08-12
chore: upgrade storybook

clucasalcantara:
## Summary
- Upgrade storybook and plugins to the latest stable version (6.0.4)

## Checklist

- [ ] Tests _(if necessary)_
- [x] Ready to be merged


clucasalcantara:
it's a minor, but it would be good to use the latest storybook version since they are adding a bunch of cool and useful features, alongside perf improvements. I won't waste too much time on this upgrade since it looks like that we have a mismatch from the webpack used by the storybook and the CRA requirements, gonna focus on our prioritized tasks to the new version and later (when we have time) get back to this but if anyone finds a fast solution then 👯 

Leaving this for now

faustbrian:
@clucasalcantara made a card internally to complete the storybook 6 migration guide because things like knobs were replaced.

PR #2668, opened 2020-08-14
feat: load wallet from route for transaction send

alexbarnsley:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/hooks/env.ts
```diff
@@ -13,5 +13,5 @@ export const useActiveWallet = () => {
 	const profile = useActiveProfile();
 	const { walletId } = useParams();
 
-	return useMemo(() => profile.wallets().findById(walletId), [profile, walletId]);
+	return useMemo(() => (walletId ? profile.wallets().findById(walletId) : undefined), [profile, walletId]);
```
faustbrian, line 16: This should never return anything but a wallet or an exception. If `useActiveWallet` is called and no wallet is found then `useActiveWallet` shouldn't have been called anyways.

alexbarnsley: Then how do I cater for the Send Transaction form allowing for a wallet from route, or no wallet from route @faustbrian? There will be the scenario (e.g. ARK URI) where the send page can be opened without an activeWallet. I added the walletId check because if it was undefined findById returned all wallets.

faustbrian: I would first have the user select the wallet from which he wants to interact with the URI data, redirect internally to a route that contains the wallet ID and then open the modal based on that instead of allow `useActiveWallet` to break out of its expected behaviour and increasing the likelihood of bugs.

The user has to select a wallet anyways and all of the internal routing after that we can do since the transaction signing process is now a separate page and not a modal that floats around on any random page.

alexbarnsley:
@faustbrian I'll fix the build issues once I figure out what to do about the wallet

src/app/hooks/env.ts
alexbarnsley: That's doable. We'll probably need some sort of modal with the network and sender fields, which can be shown prior to the send screen when activating a URI. That's the only scenario I can think of right now.

I think I will make the network field readonly to prevent the sender from becoming undefined/null

PR #2700, opened 2020-08-20
feat: sync delegates in background

clucasalcantara:
## Summary
- Trigger a background job for sync delegates when a profile is selected using electron ipc as worker

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


faustbrian:
@clucasalcantara After https://github.com/ArkEcosystem/desktop-wallet/pull/2715 is merged you can use `usedCoinsWithNetworks` to get a list of all coins and wallets that are used. Then you can use those to create promises of `env.coins().syncDelegates(coin, network)` and resolve them with `Promise.allSettled`.

This should ideally be done before the `env.boot()` method is called so that some requests are already cached and afterwards this should be done in the background.

src/app/App.tsx
```diff
@@ -34,11 +34,38 @@ const Main = () => {
...
+		coinsInUse.forEach((coin) => {
+			const coinNetworks = coinsData[coin];
+			coinNetworks.forEach((network) => {
+				delegatesPromises.push(Promise.resolve(env.coins().syncDelegates(coin, network)));
+			});
+		});
+
+		Promise.allSettled(delegatesPromises).then((results: any) => {
+			results.forEach(({ status }: { status: string; value: any }) => {
+				if (status !== "fulfilled") throw new Error("Error synchronizing delegates");
```
faustbrian, line 52: We shouldn't throw an error and abort evaluating the results just because an error occurred. I don't even see why we need to evaluate them at all. The only reason to use `allSettled` instead of `all` is so that we don't encounter an exception if 1 out of 20 coins fails to sync their data.

src/domains/transaction/components/VoteDetail/VoteDetail.tsx
```diff
@@ -57,19 +57,13 @@ export const VoteDetail = ({ transaction, walletAlias, ticker, isOpen, onClose }
 	const [delegates, setDelegates] = useState<ReadOnlyWallet[]>([]);
 
 	useEffect(() => {
-		const syncDelegates = async () => {
+		const syncDelegates = () => {
 			setIsLoadingDelegates(true);
 
-			// TODO: make senderWallet non-nullable
```
faustbrian, line 63: `senderWallet` is still nullable.

src/app/App.tsx
```diff
@@ -34,11 +34,38 @@ const Main = () => {
...
+
+		coinsInUse.forEach((coin) => {
+			const coinNetworks = coinsData[coin];
+			coinNetworks.forEach((network) => {
+				delegatesPromises.push(Promise.resolve(env.coins().syncDelegates(coin, network)));
+			});
+		});
+
+		Promise.allSettled(delegatesPromises).then((results: any) => {
+			results.forEach(({ status }: { status: string; value: any }) => {
```
faustbrian, line 51: Always use `for (const of )` instead `forEach` for consistency with all other code.

src/app/App.tsx
```diff
@@ -34,11 +34,38 @@ const Main = () => {
...
 
+	const syncDelegates = (env: Environment) => {
+		console.log("Running delegates sync...");
+		const coinsData = env.usedCoinsWithNetworks();
+		const coinsInUse = Object.keys(coinsData);
+		const delegatesPromises: any = [];
+
+		coinsInUse.forEach((coin) => {
+			const coinNetworks = coinsData[coin];
+			coinNetworks.forEach((network) => {
```
faustbrian, line 45: Always use `for (const of )` instead `forEach` for consistency with all other code.

src/app/App.tsx
```diff
@@ -34,11 +34,38 @@ const Main = () => {
 
 	const isOnline = useNetworkStatus();
 
+	const syncDelegates = (env: Environment) => {
+		console.log("Running delegates sync...");
+		const coinsData = env.usedCoinsWithNetworks();
+		const coinsInUse = Object.keys(coinsData);
+		const delegatesPromises: any = [];
+
+		coinsInUse.forEach((coin) => {
```
faustbrian, line 43: Always use `for (const of )` instead `forEach` for consistency with all other code.

src/domains/transaction/components/VoteDetail/VoteDetail.tsx
clucasalcantara: Yeah, didn't touch vote details to fix this part yet, getting the background sync working first, gonna update this before the final PR 🚀

src/domains/transaction/components/VoteDetail/VoteDetail.tsx
```diff
@@ -60,7 +60,11 @@ export const VoteDetail = ({ transaction, walletAlias, ticker, isOpen, onClose }
 		const syncDelegates = () => {
 			setIsLoadingDelegates(true);
 
-			setDelegates(DelegateMapper.execute(senderWallet!, (transaction as Contracts.VoteData).votes()));
+			if (senderWallet) {
```
faustbrian, line 63: Shouldn't there instead be a check that right after the `senderWallet` is retrieved that it isn't empty instead of setting defaults if it is empty? If the sender can't be found in the current profile then you are dealing either with corrupted local data or transactions that belong to a wallet outside of your profile, which would indicate that the SDK is fetching wrong information from the network.

luciorubeens: @faustbrian Hm ok so when that doesn't find the sender, should it be redirected off the page? or just get it from transaction `transaction.wallet()`?

faustbrian: If the sender can't be found within the profile then there is some data corruption since a vote transaction is always sent from one of your own wallets that are imported into your wallet so I can't even think of a scenario where the wallet for the vote transaction wouldn't exist since you need to have the wallet imported to see the transaction. Is there even a real scenario where the wallet wouldn't be found?

luciorubeens: There is not, however `findByAddress` is nullable, so should we just assume that it will always be defined with `senderWallet!`?

faustbrian: It should be checked if `senderWallet` is undefined and if it is throw an exception because then something is wrong with either the SDK or the data that is persisted.

faustbrian (approved):

PR #2713, opened 2020-08-24
fix: adjust transaction row link props

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/domains/transaction/components/TransactionTable/TransactionRow/TransactionRow.tsx
```diff
@@ -53,8 +53,7 @@ export const TransactionRow = ({
 				<div className="inline-block align-middle">
 					<Link
 						data-testid="TransactionRow__ID"
-						to=""
-						onClick={onClick}
+						to={{ pathname: transaction.explorerLink() }}
```
dated:
```suggestion
						to={transaction.explorerLink()}
```

should be enough ☝️

goga-m: Right :+1: Needs also to stop propagate the onClick event for the row as it still throws the error

faustbrian (approved):

PR #2748, opened 2020-08-28
feat: sync wallet data

clucasalcantara:
## Summary
- Replace sync votes and sync identity per initialize run
- Add syncExchangeRate to the wallet data sync
- Add scheduler service to run syncs periodically and persist when timers run

## Checklist

- [x] Tests _(if necessary)_
- [ ] Ready to be merged


src/app/services/Scheduler.ts
```diff
@@ -0,0 +1,12 @@
+export const Scheduler = (interval = 300000) => ({
```
faustbrian, line 1: Functions should always be `camelCase` and class `PascalClass`.

src/domains/vote/pages/Votes/Votes.tsx
```diff
@@ -94,10 +94,7 @@ export const Votes = () => {
 	}, [activeProfile, network]);
 
 	const loadDelegates = useCallback(
-		async (wallet) => {
-			// TODO: move this to profile initialising and run it every X period
-			// eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
-			await env.coins().syncDelegates(wallet?.coinId()!, wallet?.networkId()!);
+		(wallet) => {
 			// eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
 			const delegates = env.coins().delegates(wallet?.coinId()!, wallet?.networkId()!);
```
faustbrian, line 99: This should also be possible to be replaced if delegates are synced in the background.

faustbrian:
After this PR there should be another PR to include https://github.com/ArkEcosystem/platform-sdk/releases/tag/0.9.318 in the same way as delegate syncing to avoid constantly fetching fees when a user moves to a transaction send page.

src/app/services/Scheduler.ts
clucasalcantara: Do you think that we should create a class instead of just a function in this case to explicit that the Scheduler should be `service/module` ? 🤔

src/app/App.tsx
```diff
@@ -52,15 +52,35 @@ const Main = ({ syncInterval }: Props) => {
 			setShowSplash(false);
 		};
 
+		const syncWallets = async () => {
+			console.log("Running wallet data sync...");
+			const profiles = env.profiles().values();
+
+			for (const profile of profiles) {
+				const [wallet] = profile.wallets().values();
```
faustbrian, line 60: What's the intention of only syncing the first wallet of each profile?

clucasalcantara: Just debug since I was having g issues to successfully sync the wallet votes, I'm gonna replace this to an iteration over profile wallets.

src/app/App.tsx
```diff
@@ -54,23 +54,23 @@ const Main = ({ syncInterval }: Props) => {
...
-
-					// TODO: sync this through `env.exchangeRates().syncAll()` instead of per wallet.
-					// await wallet?.syncExchangeRate();
-
-					setShowSplash(false);
-				} catch (error) {
-					console.error(`Error synchronizing wallet data ${error}`);
-					setShowSplash(false);
+				const wallets = profile.wallets().values();
+				for (const wallet of wallets) {
```
faustbrian, line 62: The syncing should be extracted into a function that is then pushed to an array of promises and resolved with `Promise.allSettled`. This implementation would be a lot slower compared to that if someone has dozens of wallets.

clucasalcantara: gonna do it just like delegates than, updating!

PR #2753, opened 2020-08-30
refactor: add delegate resignation loading state

goga-m:
## Summary
When loading fees and wallet delegate on `Resign Resignation` first step, show skeletons instead of page loader.

2020-08-30-191143_897x755_scrot

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged

### DIFF
--- a/src/app/__snapshots__/App.test.tsx.snap
+++ b/src/app/__snapshots__/App.test.tsx.snap
@@ -6,7 +6,7 @@ exports[`App should close splash screen if not demo 1`] = `
     class="debug-screens"
   >
     <div
-      class="sc-fzoJus cXGpQF"
+      class="sc-fzqMdD fRMNNm"
       data-testid="RouterView__wrapper"
     >
       <div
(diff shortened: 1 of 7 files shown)


faustbrian (changes requested):
The fees shouldn't have a loading indicator because they would be removed after https://github.com/ArkEcosystem/desktop-wallet/pull/2748#issuecomment-683221907

faustbrian (changes requested):
The delegate should also not require a loading animation because the only data that is used from it is the username and the implementation has 2 issues.


```
	useEffect(() => {
		const fetchDelegateInfo = async () => {
			const delegate = await activeWallet.client().delegate(activeWallet.address());
			setDelegate(delegate);
		};
		fetchDelegateInfo();
	}, [activeWallet]);
```

Issue 1 is that we are loading the delegate info even though we have all of them synced in-memory. Issue 2 is that we are loading this info at all because the only data that is used from it is the username, which already exists on the wallet itself since you have to have the wallet imported in order to resign it.

faustbrian:
Closing this because these loading states are only necessary due to flawed implementations.

PR #2772, opened 2020-09-01
test: fix e2e errors

goga-m:
## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/domains/profile/pages/MyRegistrations/MyRegistrations.tsx
```diff
@@ -74,8 +76,14 @@ export const MyRegistrations = ({ blockchainRegistrations, businessRegistrations
 	};
 
 	useEffect(() => {
-		const delegateRegistrations = activeProfile.registrationAggregate().delegates();
-		setDelegates(delegateRegistrations);
+		const fetchDelegates = async () => {
+			await env.delegates().syncAll();
```
goga-m, line 80: @faustbrian I had to sync for delegates here otherwise it throws:
```
Error: The delegates for [ARK.devnet] have not been synchronized yet. Please call [syncDelegates] before using this method.
```
and redirects the user to "Something went wrong" screen, which is why the e2e tests for "My Registrations" were failing.

Tested by syncing delegates upon boot (with persist) but strangely enough the error was still thrown.
Based on the syncing PR from @clucasalcantara, where would you recommend to call the delegate sync?

faustbrian: The delegates should be synced before the profile is accessed. Though not sure if it is worth fixing it here because @clucasalcantara PR will move it.

faustbrian (approved):

PR #2791, opened 2020-09-03
feat: list developer registrations

faustbrian:


src/domains/profile/pages/MyRegistrations/MyRegistrations.tsx
```diff
@@ -79,6 +79,11 @@ export const MyRegistrations = () => {
 				.registrations(Enums.EntityType.Business);
 			setBusinesses(businessRegistrations.items());
 
+			const developerRegistrations = await activeProfile
+				.entityAggregate()
+				.registrations(Enums.EntityType.Developer);
+			setDevelopers(developerRegistrations.items());
```
goga-m, line 85: @faustbrian tests fail because when calling `entityAggregate().registrations()` multiple times, only the first call returns results, the next call  returns empty results.

Working Example: 
```js
const developerRegistrations = await activeProfile.entityAggregate().registrations(Enums.EntityType.Developer);
// Returns developer entities
```
Erroring example:
```js
const businessRegistrations = await activeProfile.entityAggregate().registrations(Enums.EntityType.Business);
// Returns business entities

const developerRegistrations = await activeProfile.entityAggregate().registrations(Enums.EntityType.Developer);
// Empty
```

goga-m:
@faustbrian after notice from @samharperpittam that the Developer section will not be used (until further notice), I will remove the table from  "My Registrations" page for now and we will add it later if needed.

src/domains/profile/pages/MyRegistrations/MyRegistrations.tsx
goga-m: This case is satisfied when calling for developer entities on the `Erroring example` 
```
if (lastResponse && !lastResponse.hasMorePages()) {
	return reject(
		`Fetched all transactions for ${syncedWallet.id()}. Call [#flush] if you want to reset the history.`,
	);
}

```
https://github.com/ArkEcosystem/platform-sdk/blob/196a9b7d84f927b01b4f00ed40a54008fae4a225/packages/platform-sdk-profiles/src/profiles/aggregates/transaction-aggregate.ts#L64

But flushing works only when calling same type twice. Doesn't seem to work with the case of `Erroring Example`

PR #2793, opened 2020-09-04
fix: avoid random default fees

faustbrian:
Defaults should be simply 0 and not some random numbers.


faustbrian:
Merged into https://github.com/ArkEcosystem/desktop-wallet/pull/2795

PR #2799, opened 2020-09-04
feat: sync wallet votes when import a wallet

brenopolanski:
## Summary

Task: https://app.clickup.com/t/236zv3

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/App.tsx
```diff
@@ -61,10 +61,26 @@ const Main = ({ syncInterval }: Props) => {
 
 			await env.verify(shouldUseFixture ? fixtureData : undefined);
 			await env.boot();
-			await Promise.allSettled([syncDelegates(), syncWallets(), syncFees(), syncExchangeRates()]);
+
+			const profiles = env.profiles().values();
+			const wallets = profiles.map((profile) => profile.wallets().values()).flat();
+
+			const syncWalletsVotes = async () => {
+				for (const wallet of wallets) {
```
faustbrian, line 69: This will resolve the promises in sequence. This will be taking forever for people with a lot of wallets.

brenopolanski: Actually, I won't need this code anymore, because `WalletService.syncAll()` (https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk-profiles/src/environment/services/wallet-service.ts#L13) already has a call to `syncVotes()`. However, I'll need `syncVotes()` when importing a wallet.

PR #2892, opened 2020-09-29
fix: avoid demo in production

clucasalcantara:
## Summary
- Fix demo mode usage between production/development.

When switching from dev to prod mode if you get any `ValidationError` related to profiles please run:

`indexedDB.deleteDatabase("localforage")`

Once in production, the storage should work fine, but between demo/prod exists this inconsistency due to electron instances

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


luciorubeens:
Fixed in #2890

PR #2926, opened 2020-10-05
feat: scan for ledger wallets

luciorubeens:
## Summary

- [x] Waiting device modal
- [x] Waiting connection
- [x] Synchronize wallets
- [x] Summary and change names

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/contexts/Ledger/use-ledger.ts
```diff
@@ -32,6 +34,20 @@ export const useLedger = (transport: typeof Transport) => {
...
 	);
 
+	const importLedgerWallets = useCallback(
+		async (wallets: { address: string; index: number }[], coin: Coins.Coin, profile: Profile) => {
+			for (const { address, index } of wallets) {
+				const wallet = await profile
+					.wallets()
+					.importByAddress(address, coin.network().coin(), coin.network().id());
+				wallet.data().set(WalletFlag.Ledger, true);
+				wallet.data().set("LEDGER_INDEX", index);
```
faustbrian, line 44: Add a `WalletFlag` to the SDK for this or this could get removed because the data is unknown.

luciorubeens: Just used the current flag to set the index, looks ok or better to add a new one?

faustbrian: Should work though an explicit `LedgerIndex` attribute makes it more clear what it is and what it could be used for.

faustbrian: https://github.com/ArkEcosystem/platform-sdk/releases/tag/0.9.393

src/app/contexts/Ledger/use-ledger.ts
```diff
@@ -0,0 +1,182 @@
...
+						};
+					} catch {
+						// New Cold Wallet
+						wallet = {
+							address,
+							balance: BigNumber.ZERO,
+							index: cursor,
+							isNew: true,
+						};
+						hasMore = false;
```
faustbrian, line 139: Doesn't this completely abort importing of any wallets if a single wallet can't be retrieved via API? This doesn't necessarily mean that the wallet doesn't exist, could be just a timeout or rate limiting.

src/app/contexts/Ledger/use-ledger.ts
```diff
@@ -0,0 +1,182 @@
...
+
+				const instance = await env.coin(coin, network);
+				const slip44 = instance.config().get<number>("network.crypto.slip44");
+
+				dispatch({ type: "busy" });
+
+				let hasMore = true;
+				let cursor = 0;
+
+				while (hasMore) {
```
faustbrian, line 112: Since this `while` loop performs an API call it would be better to push these tasks into an array of promises and then use `Promise.allSettled` to resolve them faster instead of running 1 request at a time.

src/app/contexts/Ledger/use-ledger.ts
```diff
@@ -0,0 +1,182 @@
...
+					const path = formatDerivationPath(slip44, cursor);
```
faustbrian, line 113: This whole process of preparing the path, publicKey, and address should ideally be in the SDK as an `importFromLedger` method. The wallet has way too much knowledge about the specifics of how the ledger data is used to create a wallet from it.

src/app/contexts/Ledger/use-ledger.ts
luciorubeens: Should I check that the error is not 404 and retry?

faustbrian: I would simply continue and store the wallets that failed in an array. Then when all wallets are imported we can tell the user that wallets X, Y and Z failed to be imported and let them retry because something like a rate limit timeout can persist for up to 60 seconds. Probably best to ask Oleg for a design.

src/app/contexts/Ledger/use-ledger.ts
luciorubeens: The scan stops until it finds an empty wallet, in this case it will be searched until the first one with `status: rejected` & `reason:404`?

src/app/contexts/Ledger/use-ledger.ts
luciorubeens: Agreed

src/app/contexts/Ledger/use-ledger.ts
faustbrian: This issue goes back to https://github.com/ArkEcosystem/desktop-wallet/pull/2926#discussion_r503025589. We shouldn't be aborting this process just because a few wallets threw a 404 or had a failing HTTP request for whatever reason. All the wallets that fail to be imported should simply be gathered and shown to the user at the end with a reason for why it failed and then let them retry.

src/app/contexts/Ledger/use-ledger.ts
```diff
@@ -0,0 +1,182 @@
...
+import { Profile, WalletFlag } from "@arkecosystem/platform-sdk-profiles";
+import { BigNumber } from "@arkecosystem/platform-sdk-support";
+import Transport from "@ledgerhq/hw-transport";
+import retry from "async-retry";
+import { useCallback, useEffect, useMemo, useReducer, useRef } from "react";
+
+import { useEnvironmentContext } from "../Environment";
+import { defaultLedgerState, ledgerStateReducer } from "./Ledger.state";
+
+const formatDerivationPath = (coinType: number, index: number) => `44'/${coinType}'/${index}'/0/0`;
```
DalderupMaurice, line 11: I don't know the requirements or anything regarding this PR (I just glanced through it), but shouldn't the index be at the last part as specified in the bip44 standard? The "index" currently changes the account, not the address_index, or was this the intention? cc @faustbrian

src/app/contexts/Ledger/use-ledger.ts
luciorubeens: Ok, how many wallets must be fetched? Doesn't look good to show 50 empty wallets for the user to choose if they really want to import it

src/app/contexts/Ledger/use-ledger.ts
luciorubeens: Hey @DalderupMaurice thanks for letting me know, @sleepdefic1t had already contacted me about this and we will fix it.

PR #2972, opened 2020-10-15
fix: node empty response

goga-m:
## Summary
Fixes issue with empty response returned from server (with 200 status).
As this seems to happen on the first GET request when explicitly providing get headers.

This PR removes request headers to test HTTP response data.

## Checklist

- [ ] Tests _(if necessary)_
- [ ] Ready to be merged


goga-m:
Closing it as the fix is already in progress in core API.

PR #2990, opened 2020-10-19
fix: register entity should be available only for ark wallet

brenopolanski:
## Summary

Task: https://app.clickup.com/t/92xgxc

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


luciorubeens (commented):
Better not to show it instead of disabling. Can also be extended to others registrations types like 2nd Signature.

https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/domains/transaction/pages/SendEntityRegistration/Step1.tsx#L59

luciorubeens (approved):

PR #3072, opened 2020-11-07
refactor: catch boot error

dated:
## Summary

Shows the `ApplicationError` when an error occurs during the initial boot.

https://app.clickup.com/t/a4rrf4

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


src/app/App.test.tsx
```diff
@@ -174,7 +174,7 @@ describe("App", () => {
 	});
 
 	it("should render application error if the app fails to boot", async () => {
-		const envSpy = jest.spyOn(env, "boot").mockImplementation(() => {
+		const envSpy = jest.spyOn(Environment.prototype, "boot").mockImplementation(() => {
```
faustbrian, line 177: 🤨

dated: i don't have access to the specific env instance that needs to be mocked for this test, so changing all of them is the only option - unless there is a better approach i'm unaware of.

faustbrian:
Task linked: [CU-a4rrf4 Catch error during application boot](https://app.clickup.com/t/a4rrf4)

luciorubeens (approved):

PR #3112, opened 2020-11-20
fix: select multiple news

luciorubeens:
## Summary

https://app.clickup.com/t/8wqde7

## Checklist

- [x] Tests _(if necessary)_
- [x] Ready to be merged


samharperpittam (changes requested):
Currently displaying 15 articles per token, which can result in 60 articles displayed per page. We should change this to default to 15 articles per page regardless if 1 or 4 assets are selected via the filter.

samharperpittam (approved):

PR #3181, opened 2021-01-08
feat: disable screenshot protection in development builds

ItsANameToo:


ItsANameToo:
https://github.com/sindresorhus/electron-is-dev/issues/22 :sadup:

ItsANameToo:
if the tests pass after the last commit, I'll move the relevant checks the `is-dev` package does to a custom method as we seem to make use of the `ELECTRON_IS_DEV` flag for testing and then it should work

ItsANameToo:
why are you like this
image

ItsANameToo:
When importing the package it messes up `jest` as `electron` isn't defined, so I redefined the functionality inside the `electron-utils` file for the time being.

Preferably we would have a single file that we can use in `electron/index.js` and in `electron-utils`, so we can either ditch the package or the implementation. If we can resolve the `electron` issue that we also have in the plugin PR, we might solve it here at the same time and can readdress it

PR #3218, opened 2021-01-19
fix: scroll to top on route change

goga-m:
## Checklist

-   [ ] Ready to be merged

### DIFF
--- a/src/router/RouterView.tsx
+++ b/src/router/RouterView.tsx
@@ -23,10 +23,10 @@ export const RouterView = ({ routes, wrapper, middlewares }: Props) => {
 	const previousPath = useRef("");
 	useEffect(() => {
 		history.listen((route) => {
-			if (route.location?.pathname === previousPath.current) return;
-
-			previousPath.current = route.location?.pathname;
-			window.scrollTo(0, 0);
+			if (!previousPath.current || route.location?.pathname !== previousPath.current) {
+				previousPath.current = route.location?.pathname;
+				window.scrollTo(0, 0);
+			}
 		});
 	}, [history]);
 


faustbrian:
Task linked: [CU-cgz1zt changing page on the Delegate Table does not trigger scroll to top](https://app.clickup.com/t/cgz1zt)

ItsANameToo (changes requested):
pagination in the delegate table still does not scroll to top for me

src/domains/vote/components/DelegateTable/DelegateTable.tsx
```diff
@@ -164,6 +164,7 @@ export const DelegateTable = ({
 
 	const handleSelectPage = (page: number) => {
 		setCurrentPage(page);
+		window.scrollTo({ top: 0, behavior: "smooth" });
```
ItsANameToo, line 167: in `News.tsx` we have a `useEffect(() => window.scrollTo({ top: 0, behavior: "smooth" }), [currentPage, coins]);`, would something like that work for delegates too? e.g. watching a change in `page` prop and then scroll up instead of adding it to this method

goga-m: Sure, we could put it inside useEffect. Are there other actions that require to scroll to the top?

ItsANameToo: only when the page changes I think as that forces new content in the table 🤔

ItsANameToo (approved):

samharperpittam (approved):

PR #3255, opened 2021-01-27
fix: focus on input when click on search icon

romanlikhovich:
## Summary

This changes fixed bug with input focus when users clicked on the search icon

With these changes, users don't need to set focus on input manually

I added a new prop isFocused for the Input component. In the Input component, I created a ref via React hook useRef.
On the component render via React hook useEffect I check if the prop isFocused is true. And then I call the focus method for current ref. 
## Checklist

-   [+] My changes look good in both light AND dark mode
-   [+] The change is not hardcoded to a single network, but has multi-asset in mind
-   [+] I checked my changes for obvious issues, debug statements and commented code
-   [+] Documentation _(if necessary)_
-   [ ] Ready to be merged

I think the new props isFocused should be added to the documentation


romanlikhovich:
Will do new branches directly in the repo

PR #3275, opened 2021-01-29
fix: buttons focus only in tabbing no onclick

romanlikhovich:
## Summary

Fix setting focus on button only on tabbing no click

These changes are important for UI

Added focus-visible polyfill and added some styles for focus only on tabbing

## Checklist

-   [ ] Ready to be merged


ItsANameToo:
we might be able to remove the polyfill at a later time when we upgrade electron, as it will be natively supported in chrome v86 and up

ItsANameToo (approved):

PR #3284, opened 2021-02-01
refactor: use smartbridge in supported networks in `ReceiveFundsForm`

goga-m:
## Summary
Show smartbridge field in Receive funds modal (and QR code) in ark networks.

## Checklist

-   [x] Ready to be merged


faustbrian:
Task linked: [CU-c52g5n QR code modal is made for ark and will not work on other networks](https://app.clickup.com/t/c52g5n)

faustbrian:
Marking this as draft as the PSDK should determine this, not the wallet based on some hardcoded strings.

src/domains/wallet/components/ReceiveFunds/ReceiveFundsForm.tsx
```diff
@@ -19,6 +19,11 @@ export const ReceiveFundsForm = () => {
 		register("amount");
 	}, [register]);
 
+	const isSmartbridgeUsedInNetwork = useMemo(() => {
+		const smartbridgeNetworks = ["ark.mainnet", "ark.devnet"];
```
ItsANameToo, line 23: this should actually depend on the network configuration, e.g. https://dwallets.ark.io/api/node/configuration will show a value for the smartbridge that determines if they are enabled or not. It shouldn't be hardcoded to just ark mainnet/devnet as there can be plenty of custom networks that also support smartbridge

ItsANameToo (changes requested):

src/domains/wallet/components/ReceiveFunds/ReceiveFundsForm.tsx
goga-m: Thanks @ItsANameToo. Is this network configuration available for the other supported networks with a similar structure?

ItsANameToo: PSDK will handle it as brian mentioned, not something that should concern the wallet. It will be similar for all ARK based networks, but other networks will need special handling as they might not have it at all or in a different fashion

goga-m: Great :+1:

goga-m: @faustbrian @ItsANameToo patched the fix using `network?.can(FeatureFlag.MiscellaneousMemo);`

PR #3301, opened 2021-02-02
feat: allow plugins manual installation from urls

luciorubeens:
## Summary

https://app.clickup.com/t/ck1vtt

## Checklist

-   [x] Ready to be merged


ItsANameToo (changes requested):
- modal seems to miss a close button icon
image
- currently you can input any repository and as long as it is available on npm it will try to show it as a plugin (e.g. try https://github.com/ArkEcosystem/explorer). Might be better if there are some checks inside the package.json file to ensure it's an actual plugin and report back to the user if it isn't. Otherwise they will only notice after trying to install it that it has nothing to do with it being a plugin

also some conflicting files

luciorubeens:
Can you verify @ItsANameToo? Now it checks for required keywords.

ItsANameToo:
much better, thanks @luciorubeens

ItsANameToo (approved):

PR #3313, opened 2021-02-03
feat: report plugin

luciorubeens:
## Summary

https://app.clickup.com/t/df2nyb

## Checklist

-   [x] Ready to be merged


ItsANameToo (changes requested):
- the review icon is hardcoded to blue, making it not work with hover / dark mode
image
- reporting only works for plugins that are installed. If a plugin is not installed, it will show a report button but clicking on it doesn't do anything. Either reporting should be clear that it only works after a user has installed a plugin (e.g. greyed out and a tooltip indicating why), or the report functionality should always be available (imo preferred option, but discuss on slack)

also some conflicts to resolve

luciorubeens:
@ItsANameToo fixed

ItsANameToo (approved):

PR #3411, opened 2021-02-24
feat: add back button to navigation bar

dated:
## Summary

https://app.clickup.com/t/egz6a0

## Checklist

-   [x] Ready to be merged


faustbrian:
Task linked: [CU-egz6a0 add `back` button](https://app.clickup.com/t/egz6a0)

ItsANameToo (commented):
Scenario:
- portfolio > import > go through the 3 steps > import wallet > get redirected to the portfolio page; at this point the back button does not reset, so clicking `back` will bring you to the import page again. Meanwhile, when you click the `portfolio` page from the navbar, it resets and you have no "back" state, so should we also apply this whenever we redirect the user to one of the top level pages?

ItsANameToo:
also conflict due to other merge @dated

dated:
I initially had another implementation which made use of a `backToUrl` prop passed to the component instead of simply going back to the previous page, i went with the latter since it wasn't exactly clear which pages should be pointed to by the back button but it looks like we could use a combination of both.

dated:
Re back from top level: i think that was a design requirement but it makes sense to have the button enabled as long as there is a page to go back to.

ItsANameToo:
yeah the back button might need some finetuning to be useful, I'll ask sam to take a look too to see what he thinks

samharperpittam:
Having used this, it does feel a bit strange that there's times where the back button is active on top level pages, and times where it's not. I think we need to decide one way or the other.

For now, I'd suggest that we reset the back button every time the user is redirected to a top level page. 

We can use that as a starting point and see what feedback we get as we enter a testing phase.

ItsANameToo (approved):

PR #3421, opened 2021-02-26
fix: do not clear values in the transfer form

luciorubeens:
## Summary

https://app.clickup.com/t/dx46qv
https://app.clickup.com/t/en509z

## Checklist

-   [x] Ready to be merged

### DIFF
--- a/src/domains/profile/components/SelectRecipient/SelectRecipient.tsx
+++ b/src/domains/profile/components/SelectRecipient/SelectRecipient.tsx
@@ -44,10 +44,8 @@ export const SelectRecipient = React.forwardRef<HTMLInputElement, SelectRecipien
 		const isInvalidValue = isInvalid || fieldContext?.isInvalid;
 
 		useEffect(() => {
-			if (address) {
-				setSelectedAddress(address);
-			}
-		}, [address, setSelectedAddress]);
+			setSelectedAddress(address);
+		}, [address]);
 
 		const { allAddresses } = useProfileAddresses({ profile, network });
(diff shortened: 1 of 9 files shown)


faustbrian:
Task linked: [CU-dx46qv [Multipay] Bug](https://app.clickup.com/t/dx46qv)

ItsANameToo (changes requested):
- [ ] switching from multiple to single when you filled in fields but haven't added the recipient yet works, but if you add a single recipient (so the fields are cleared and the recipient list below gets populated) and then switch back to single, it will not fill in the recipient anymore. Is this intended? When I fill in 1 recipient on the "multiple" tab and click "continue" and then go back to the page, it will have filled in the single recipient on the "single" tab so then it does work.
- [ ] when I add a recipient, the address value does not clear. This makes it appear as if you have a value filled in while you actually do not and it results in scenarios where you do not see an "add recipient" button while all the fields seem to be filled 
image

other than that it works fine, amount is kept between screens and multiple recipients are too

ItsANameToo:
I just noticed that https://github.com/ArkEcosystem/desktop-wallet/pull/3432 is also concerned about the recipient input not clearing, so keep an eye on that if it has not been merged yet by the time you work on your changes

ItsANameToo:
@luciorubeens also some conflicts to resolve now :hide:

ItsANameToo (approved):

ItsANameToo:
@luciorubeens some snapshot failures :(

PR #3427, opened 2021-03-01
feat: update all plugins functionality

luciorubeens:
## Summary

https://app.clickup.com/t/3h72ya

## Checklist

-   [ ] Ready to be merged


ItsANameToo:
feature wise it seems to work correctly, just some UI changes:

- [ ] can you make the "update all" button also appear on the card view of the "my plugins" tab?
- [ ] we'll need a singular version of the update text as it now talks about plural when it's just 1 plugin. Should read `There is 1 update available for your plugins. Would you like to update it?` when it's about a single plugin
image
- [ ] (possibly, not final yet) we might need a red dot on the "my plugins" tab too to indicate that users can go there to update all

ItsANameToo (changes requested):
see comment

ItsANameToo (approved):

PR #3453, opened 2021-03-04
refactor: add cryptoasset step to send transfer from navbar

luciorubeens:
## Summary

https://app.clickup.com/t/egzf73

## Checklist

-   [ ] Ready to be merged


samharperpittam (changes requested):
@luciorubeens - Quick note on this.

When selecting 'Send' from the navbar, I don't think it really makes sense to show all networks if no wallets have been added for them.

For example, if I add just 1 ark devnet wallet to my profile, when I select Send from the navbar there's little point in showing APN or other networks that I can't send from anyway as I haven't added a wallet for them. If that makes sense?

So in a scenario where my profile only has 1 devnet wallet, when I select 'Send' from navbar, the only network that should be available to select is devnet.

samharperpittam:
@luciorubeens  - I believe that the network selector on the Send form should also be disabled as described in the task when selecting sending from the Nav bar. Otherwise the user can select the network twice which isn't really the desired behaviour.

luciorubeens:
@samharperpittam fixed

samharperpittam (approved):

ItsANameToo (approved):

PR #3570, opened 2021-03-29
Register proper driver for main and renderer

marianogoldman:
## Checklist

-   [ ] Ready to be merged


src/electron/index.ts
```diff
@@ -152,3 +156,17 @@ app.setAsDefaultProtocolClient("ark", process.execPath, ["--"]);
...
+
+const env = new Environment({
+	coins: {
+		ARK,
+	},
+	httpClient: new HttpClient(500),
+	storage: "indexeddb",
+	driver: "memory",
+});
+env.verify()
```
marianogoldman: Can't use await here (top-level / outside an async function).
What would be the best way to have a single instance of the environment created in the main process so that the handlers can be registered here?

luciorubeens: Perhaps something like this is enough, isn't it? :hmhm:

Main:

```typescript
import { ipcMain } from "electron";

const env = new Environment({
	coins: {
		ARK,
	},
	httpClient: new HttpClient(500),
	storage: "indexeddb",
	driver: "memory",
});

const initializeEnvironment = async () => {
	await env.verify();
	await env.boot();

    ipcMain.send("env:ready");
}

export const registerEnvironmentHandlers = async () => {
	await initializeEnvironment();
	// setupIpc();
}

```

Renderer:

```jsx
import { ipcRenderer } from "electron";

const App = () => {
	React.useEffect(() => {
		ipcRenderer.on("env:ready", () => {
			hideSplashScreen();
		});
    }, []);

	return <h1>App</h1>;
}
```

marianogoldman: Let me try. Thanks @luciorubeens

faustbrian:
Closing this for now until we revisit it.

PR #3667, opened 2021-04-22
fix: handle profile balance in navigation bar

goga-m:
## Summary

-   [x] Ensure profile converted balance is always set in navigation balance component
-   [x] Handle exceptions from `profile.convertedBalance` or undefined/NaN values

## Checklist

-   [x] Ready to be merged


faustbrian:
Task linked: [CU-hk8jd8 [General] Fix profile state error caused navigation balance component after dw reload](https://app.clickup.com/t/hk8jd8)

PR #3676, opened 2021-04-26
refactor: update Platform SDK Dependencies

goga-m:
## Checklist

-   [x] Ready to be merged


src/app/components/Notifications/NotificationTransactionItem.tsx
```diff
@@ -12,13 +11,13 @@ export const NotificationTransactionItem = ({
...
 }: NotificationTransactionItemProps) => {
-	const [transaction, setTransaction] = useState<DTO.ExtendedTransactionData>();
+	const [transaction, setTransaction] = useState<any>();
 	const [walletName, setWalletName] = useState<string>();
 
 	useEffect(() => {
 		const fetchTransaction = async () => {
 			const wallet = profile.wallets().findByAddress(notification?.meta?.walletAddress);
-			const notificationTransaction = await wallet?.findTransactionById(notification?.meta?.transactionId);
+			const notificationTransaction = wallet?.transaction().transaction(notification?.meta?.transactionId);
```
faustbrian:
```suggestion
			const notificationTransaction = await wallet?.transactionIndex().findById(notification?.meta?.transactionId);
```

src/app/contexts/Ledger/hooks/connection.ts
```diff
@@ -35,10 +35,14 @@ export const useLedgerConnection = (transport: typeof Transport) => {
 	const importLedgerWallets = useCallback(
 		async (wallets: LedgerData[], coin: Coins.Coin, profile: Contracts.IProfile) => {
 			for (const { address, path } of wallets) {
-				const wallet = await profile
-					.wallets()
-					.importByAddress(address, coin.network().coin(), coin.network().id());
+				const wallet = await profile.walletFactory().fromAddress({
```
faustbrian: Use `fromAddressWithLedgerPath` here.

src/domains/profile/components/ResetProfile/ResetProfile.tsx
```diff
@@ -22,7 +22,7 @@ export const ResetProfile = ({ isOpen, profile, onClose, onCancel, onReset }: Re
 	const { persist } = useEnvironmentContext();
 
 	const handleReset = async () => {
-		profile.initializeSettings();
+		profile.flush();
```
faustbrian: This method will also delete all data. Is that intended here?

goga-m: @faustbrian we need to reset to defaults. Noticed that `ProfileInitializer` is re-initializing settings and erasing modified profile data, same as with `initializeSettings` we used previously. Can you confirm?

faustbrian: `initializeSettings` previously only reset settings but not the data. I'll make a release tomorrow that restores the old behaviour of having 2 separate methods.

src/app/contexts/Ledger/hooks/connection.ts
```diff
@@ -35,10 +35,11 @@ export const useLedgerConnection = (transport: typeof Transport) => {
...
 			for (const { address, path } of wallets) {
-				const wallet = await profile.walletFactory().fromAddress({
+				const wallet = await profile.walletFactory().fromAddressWithLedgerPath({
 					address,
 					coin: coin.network().coin(),
 					network: coin.network().id(),
+					path,
 				});
 
 				wallet.data().set(Contracts.WalletData.LedgerPath, path);
```
faustbrian: This can be removed when using `fromAddressWithLedgerPath`.

goga-m: Seems that `path` is a required prop in `IAddressWithLedgerPathOptions` :hmmm

luciorubeens: @goga-m this line `wallet.data().set(Contracts.WalletData.LedgerPath, path);`, as it is redundant

src/domains/profile/components/SignIn/SignIn.tsx
```diff
@@ -67,9 +69,10 @@ export const SignIn = ({ isOpen, profile, onCancel, onClose, onSuccess }: SignIn
 		}
 	}, [errors, remainingTime, setError, t]);
 
-	const handleSubmit = ({ password }: any) => {
+	const handleSubmit = async ({ password }: any) => {
 		if (profile.auth().verifyPassword(password)) {
```
goga-m: @faustbrian currently we are unable to verify the password when the user submits the sign-in form in order to proceed with the restoration. The `state` error is thrown. 

All the required data though is available in the profile and could be validated without the need to restore the profile first. Made https://github.com/ArkEcosystem/platform-sdk/pull/1389 as an approach for this issue, but happy to go with another one you think fits best.

faustbrian: Will think about another solution as that breaks the single source of data for the authenticator instance because now any profile can use any authenticator instance of any profile which is nonsensical.

faustbrian: https://github.com/ArkEcosystem/platform-sdk/commit/2056a696f3bdbf795c392b4a7c11435e1c724785

```ts
const result = env.profiles().tap("some-id", (profile: IProfile) => {
    // do something with the profile and return the result
});
```

goga-m: @faustbrian looks great! Thanks

src/domains/profile/pages/Welcome/Welcome.tsx
```diff
@@ -70,13 +70,16 @@ export const Welcome = () => {
 	const handleRequestedAction = (profile: Contracts.IProfile, action: any, password?: string) => {
 		closeSignInModal();
 
+		if (password) {
+			container.bind("State<Profile>", profile);
```
faustbrian: Never set container values directly, tap the profile if you want do something to it before it has been restored.

goga-m: @faustbrian it's the memory password that requires the profile and throws exceptions. In this state, the user's password has been successfully verified, and we need to store it in memory to use it later for restoration and have it available (in profile domains). 

Using sdk's `MemoryPassword` doesn't give this flexibility because it requires the profile to be already set in order to set the password (which is not the case in dw because we want restoration to happen in the next tick, not before). 

`Tap` on the other hand "forgets" the profile and unsets password after callback run, and it seems that we cannot safely rely on `MemoryPassword` and make sure password is not removed from some previously called tap.

So practically we need to set the password and make sure it's there before signing out or calling `env.profiles().forget`. And it needs to happen before restoration.

If `MemoryPassword` is not intended to be used by the consumer, but only internally, we could store active profile's password somewhere else outside sdk, but I wanted to check with you first because it's an important decision.

faustbrian: https://github.com/ArkEcosystem/platform-sdk/pull/1395 does this resolve all your issues?

goga-m: Should be ok if it's independent of `State` and can be called prior to profile restoration. Will make a few iterations and send an update. Thanks  :+1:

faustbrian: https://github.com/ArkEcosystem/platform-sdk/releases/tag/6.0.9 should be published in a few

src/domains/dashboard/e2e/dashboard-routing.e2e.ts
```diff
@@ -8,30 +8,33 @@ const translations = buildTranslations();
...
-		.expect(Selector('[data-testid="modal__inner"]').withText(translations.PROFILE.MODAL_SIGN_IN.TITLE).exists)
-		.ok();
-
-	await t.typeText(Selector('[data-testid="SignIn__input--password"]'), "password");
-	await t.click(Selector('[data-testid="SignIn__submit-button"]'));
-
-	await t.expect(Selector("div").withText(translations.COMMON.WALLETS).exists).ok({ timeout: 30000 });
-	await t.expect(getLocation()).contains("/dashboard");
-});
+// TODO: In ci, password protected profile is failing to restore and enter dashboard, possibly due to container.bind('State<Profile', profile)
```
faustbrian, line 17: Is this still an issue?

goga-m: I think it was still having timeouts, but will make a few more tests and re-enable it in a new PR

PR #3723, opened 2021-05-06

faustbrian:


src/app/hooks/use-network-options.test.tsx
```diff
@@ -12,20 +12,9 @@ describe("useNetworkOptions hook", () => {
...
-				value: "ark.devnet",
-			},
-			{
-				label: "Compendia Mainnet",
-				value: "compendia.mainnet",
-			},
-		]);
+		expect(networks).toContainEqual({ value: "ark.mainnet", label: "ARK Mainnet" });
+		expect(networks).toContainEqual({ value: "ark.devnet", label: "ARK Devnet" });
+		expect(networks).toContainEqual({ value: "compendia.mainnet", label: "ARK Mainnet" });
```
faustbrian: This should be Compendia Mainnet, must be either a mistake in the DW or SDK.

goga-m: Fixed to use `network.coinName()` but `network.coin()` for compendia seems to return `ARK` from sdk :hide:

faustbrian: It returns ARK for the coin because that is the parent coin.

PR #3801, opened 2021-05-25
refactor: pass non-satoshi fee value to Platform SDK

alesmit:
## Summary

https://app.clickup.com/t/kbamn1

**Relates to** https://github.com/ArkEcosystem/platform-sdk/pull/1621

## Checklist

-   [x] Tests
-   [x] Ready to be merged


faustbrian:
Task linked: [CU-kbamn1 Pass non-satoshi fee and amount to PSDK](https://app.clickup.com/t/kbamn1)

src/app/components/Input/InputCurrency.tsx
```diff
@@ -1,68 +1,44 @@
...
 
 		useEffect(() => {
-			const evaluateValue = (value: any) => {
-				if (value?.display) {
-					return value;
-				}
-
-				return convertValue(value?.toString());
-			};
+			// when value is changed outside, update amount as well
```
alesmit, line 19: @goga-m would be nice to wrap this logic into a hook

src/app/components/Input/InputCurrency.tsx
```diff
@@ -1,68 +1,44 @@
...
 	onChange?: (value: any) => void;
-	magnitude?: number;
 	as?: React.ElementType<any>;
 } & Omit<React.InputHTMLAttributes<any>, "onChange" | "defaultValue">;
 
-interface ISelectionRange {
-	start: number | null;
-	end: number | null;
-}
+const sanitizeValue = (x?: string) => Currency.fromString(x || "").display;
```
alesmit: moved this outside the component as it's a pure function, not sure its the best approach tho

faustbrian: Avoid abbreviations, always give it a descriptive name even if it is just `value`.

src/app/components/Input/InputCurrency.tsx
alesmit: I ended up not adding a custom hook because this component is the only one that would have used it, but also because logic it handles is very simple and short.

src/domains/transaction/components/MultiSignatureRegistrationForm/__snapshots__/MultiSignatureRegistrationForm.test.tsx.snap
```diff
@@ -392,12 +392,12 @@ exports[`MultiSignature Registration Form should render form step 1`] = `
...
                         name="fee"
                         type="text"
-                        value=""
+                        value="0"
                       />
                     </div>
                   </div>
                   <div
-                    class="absolute bottom-0 px-1 w-full invisible"
+                    class="absolute bottom-0 px-1 w-full"
```
alesmit: @goga-m this snapshot change seems concerning, it was `invisible` before and now it's not. I think it's related to the value above changed from empty to "0". I would review it later when we'll have the sdk updated.

src/app/components/Input/InputRange.tsx
```diff
@@ -1,58 +1,45 @@
...
-		const [inputValue, setInputValue] = useState<CurrencyInput>(convertToCurrency(value));
-		const rangeValues = useMemo(() => formatRange(inputValue, max), [formatRange, inputValue, max]);
-
-		const trackBackgroundMinValue = Number(inputValue.display);
-		const minValue = Math.min(Number(min), trackBackgroundMinValue);
+		// @TODO could be simplified after SDK update
+		const rangeValues = useMemo<number[]>(() => {
+			const sanitized = BigNumber.make(value);
+			/* istanbul ignore next */
+			return isNaN(sanitized.toNumber()) ? [] : [Math.min(sanitized.toNumber(), max)];
```
goga-m, line 29:
```
			if (isNaN(sanitized.toNumber()) || sanitized.isZero()) {
				return [];
			}

			return [Math.min(sanitized.toNumber(), max)];

```
@alesmit this should bring back the invisible state for the range, when zero fee is entered

src/domains/transaction/components/MultiSignatureRegistrationForm/__snapshots__/MultiSignatureRegistrationForm.test.tsx.snap
goga-m: @alesmit you are about right :+1: zero fee was not handled and the range slider was visible when the value was 0 in the input

src/app/components/Input/InputRange.tsx
goga-m:
```suggestion
			const sanitized = BigNumber.make(value);
			/* istanbul ignore next */
			if (isNaN(sanitized.toNumber()) || sanitized.isZero()) {
				return [];
			}

			return [Math.min(sanitized.toNumber(), max)];
```
@alesmit this should bring back the invisible state for the range, when zero fee is entered

src/domains/transaction/components/MultiSignatureRegistrationForm/__snapshots__/MultiSignatureRegistrationForm.test.tsx.snap
```diff
@@ -412,7 +412,7 @@ exports[`MultiSignature Registration Form should render form step 1`] = `
                         data-testid="InputCurrency"
                         name="fee"
                         type="text"
-                        value=""
+                        value="0"
```
alesmit: I'm checking why this changed

goga-m: @alesmit `minParticipants` seems undefined here https://github.com/ArkEcosystem/desktop-wallet/blob/feat/fee-pass-non-satoshi-CU-kbamn1/src/domains/transaction/components/MultiSignatureRegistrationForm/FormStep.tsx#L78  but it should be ok if the validation is failing in the input and preventing the user to continue for 0 participants.

alesmit: It seems that the old implementation was replacing 0 with empty string, and removing `|| 0` here brought this snapshot as it was before https://github.com/ArkEcosystem/desktop-wallet/blob/5afaeb2c8a9325479fc55edd73e15e36740a0ac6/src/domains/transaction/components/MultiSignatureRegistrationForm/FormStep.tsx#L102

goga-m:
@alesmit have been testing the input behavior in dev mode, overall it seems to be working without any major issues. Nice work :1st_place_medal: 
A minor issue that might have been always there but revealed now, is that it allows the user to enter 0 in the input fee without erroring in validation. This should be resolved if we added

```js
				if (feeSatoshi.isZero()) {
					return t("COMMON.VALIDATION.FIELD_REQUIRED", {
						field: t("COMMON.FEE"),
					});
				}

```
before this line https://github.com/ArkEcosystem/desktop-wallet/blob/feat/fee-pass-non-satoshi-CU-kbamn1/src/domains/transaction/validations/Common.ts#L34

src/domains/transaction/components/MultiSignatureRegistrationForm/__snapshots__/MultiSignatureRegistrationForm.test.tsx.snap
goga-m: Ah right, indeed. Got confused from the snapshots and thought it was changed in the `minParticipants` input. What you say is correct. Great :+1:

faustbrian (approved):

PR #3806, opened 2021-05-27
style: prefer interfaces over types for consistency with platform-sdk

delaaxe:
## Summary

Just a proposal to match platform-sdk's style

## Checklist

-   [ ] Ready to be merged


w3ea:
in DW we're using `conditional types` and `conditional props`, and for this we can't use Interface
idk what's the point of using Type and Interface together

resource:
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

delaaxe:
> in DW we're using `conditional types` and `conditional props`, and for this we can't use Interface
> idk what's the point of using Type and Interface together
> 
> resource:
> https://www.typescriptlang.org/docs/handbook/2/conditional-types.html
> https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

Discriminated unions work just fine with interfaces. Also, all the examples in those links use the interface style!

I guess the point could be:
- `interface` for pojos
- `type` for algebraic construction

Didn't know about conditional types thanks

w3ea:
> Discriminated unions work just fine with interfaces. Also, all the examples in those links use the interface style!
> 
> I guess the point could be:
> 
> * `interface` for pojos
> * `type` for algebraic construction
> 
> Didn't know about conditional types thanks

I meant, we can't just use Interface for `conditional types` or `discriminated unions`
we have to combine Interface and Type, like this:
```
interface Circle {
  kind: "circle";
  radius: number;
}

interface Square {
  kind: "square";
  sideLength: number;
}

type Shape = Circle | Square;
```

my point is, no need to use two different structure when Type working for any construction :oleglala:

faustbrian:
Can't find the documentation link right now but Microsoft recommends to use `interfaces` instead of `type` wherever possible because they were always intended to serve distinct purposes, you weren't supposed to always use `interface` or always `type` but people just went with that because it was possible.

TLDR: Use either where appropriate and don't just default to one because it is (or feels) consistent even though it goes against recommendations by even the language authors.

delaaxe:
I agree with @faustbrian, better to use most standard code. @w3ea IMO the example you gave is exactly as it should be (it's from the official docs after all).

Did someone say you have to choose between using if-statements only or switch-statements only? No, you're supposed to use both

w3ea:
[from Typescript document:](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)
For the most part, you can choose **based on personal preference**, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until **you need to use features** from type.

from personal experience:
if you hover on interface you can't see anything (just interface name), but if you hover on type you see a whole structure
type
interface

w3ea:
@faustbrian this is just suggestion, do everything that you think is better for product :oleglala:

faustbrian (approved):

PR #3807, opened 2021-05-27
chore: comments and nitpicks

delaaxe:
## Checklist

-   [ ] Ready to be merged


src/app/App.tsx
```diff
@@ -136,31 +134,28 @@ export const App = () => {
...
-					LUNA,
-					TRX,
-					XLM,
-					XRP,
-					ZIL,
-				},
-				httpClient,
-				storage,
-			}),
-	);
```
w3ea: here we're using [lazy initial state](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state)
to be executed only on the initial render, and it's better to keep it

delaaxe: Didn't know about that, thanks

w3ea (approved):

faustbrian (approved):

PR #3814, opened 2021-05-31
refactor: wallet import options depending on the network

w3ea:
## Summary
https://app.clickup.com/t/jzbxze

## Checklist

-   [ ] My changes look good in both light AND dark mode
-   [ ] The change is not hardcoded to a single network, but has multi-asset in mind
-   [ ] I checked my changes for obvious issues, debug statements and commented code
-   [ ] Documentation _(if necessary)_
-   [x] Tests _(if necessary)_
-   [x] Ready to be merged

PR #3838, opened 2021-06-04
feat: use different transaction signing methods based on wallet import method

alesmit:
## Summary

https://app.clickup.com/t/jz5wn8
https://app.clickup.com/t/kbanb2

## Checklist

-   [x] Tests
-   [x] Ready to be merged


faustbrian:
Task linked: [CU-kbanb2 Use import method of wallet to hide/show elements](https://app.clickup.com/t/kbanb2)

faustbrian:
Task linked: [CU-jz5wn8 Use different transaction signing methods based on import method](https://app.clickup.com/t/jz5wn8)

samharperpittam (changes requested):
So one thing i've noticed on this PR is that the task specifies that the user should be able to import via public key. But within the DW there's no option to do so. 

Is that something we can add @alesmit ?

w3ea:
> So one thing i've noticed on this PR is that the task specifies that the user should be able to import via public key. But within the DW there's no option to do so.
> 
> Is that something we can add @alesmit ?

@samharperpittam this feature doesn't exist right now and i'll add it after this PR merged https://github.com/ArkEcosystem/desktop-wallet/pull/3814

PR #3843, opened 2021-06-08
refactor: add switch component

alesmit:
## Summary

If applied, this PR will add a Switch component, wrapping a Toggle component with two clickable labels on the sides. Since it shows up in both VerifyMessage and the new InputFee redesign, the idea is to wrap its logic in a reusable component.

Note: additional tweaking will be implemented along with InputFee to allow for slightly different designs (e.g. smaller options text).

### Dark
image

### Light
image

## Checklist

-   [x] Tests
-   [x] Ready to be merged

### DIFF
--- a/src/app/components/Switch/Switch.tsx
+++ b/src/app/components/Switch/Switch.tsx
@@ -4,23 +4,26 @@ import React from "react";
 import { Toggle } from "../Toggle";
 import { SwitchText } from "./SwitchText";
 
-interface SwitchOption<TValue> {
+export interface SwitchOption<TValue = string> {
 	label: string;
 	value: TValue;
 }
 
-export type SwitchOptions<TValue = string> = [SwitchOption<TValue>, SwitchOption<TValue>];
-
 interface Props<TOptionValue = string> {
 	value: TOptionValue;
 	onChange: (value: TOptionValue) => void;
-	options: SwitchOptions<TOptionValue>;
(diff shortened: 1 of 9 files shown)


src/app/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { Toggle } from "../Toggle";
+import { SwitchText } from "./SwitchText";
+
+interface SwitchOption<TValue> {
+	label: string;
+	value: TValue;
+}
+
+export type SwitchOptions<TValue = string> = [SwitchOption<TValue>, SwitchOption<TValue>];
```
w3ea: here is good to use object with left/right option key
it's easier when use it and for maintenance in future like here:
https://github.com/ArkEcosystem/desktop-wallet/pull/3843/files#diff-8b63a76540644aa1d1b5d4de1bf27ff0c62d66c96c97d231f9a1113f1b042ef9R128-R137
```suggestion
export type SwitchOptions<TValue = string> = {leftOption: SwitchOption<TValue>, rightOption: SwitchOption<TValue>};
```

alesmit: @w3ea thanks, that is good point. I ended up adding `leftOption` and `rightOption` as props, also to reduce nesting. https://github.com/ArkEcosystem/desktop-wallet/pull/3843/commits/d1b89491d21e41e37475d5da15ecc8ba384c3111

PR #3846, opened 2021-06-09

faustbrian:
https://github.com/ArkEcosystem/platform-sdk/releases/tag/9.1.0


src/domains/transaction/hooks/use-transaction-builder.ts
```diff
@@ -16,10 +18,27 @@ const prepareMultiSignature = async (
...
+	await wallet.ledger().connect(transport);
+
+	const signatory = await wallet
+		.signatory()
+		.ledger(wallet.data().get<string>(ProfileContracts.WalletData.DerivationPath)!);
+
+	return {
+		...input,
+		signatory,
+		nonce: wallet.nonce().plus(1).toFixed(),
```
faustbrian: This can be skipped. The SDK will get the nonce for the address that belongs to the public key of the derivation path and increase it.

src/domains/transaction/hooks/use-transaction-builder.ts
```diff
@@ -16,10 +18,27 @@ const prepareMultiSignature = async (
...
+const prepareLedger = async (
+	profile: ProfileContracts.IProfile,
+	input: Services.TransactionInputs,
+	wallet: ProfileContracts.IReadWriteWallet,
+	connectFn: ConnectFn,
+	transport: typeof Transport,
+) => {
+	await connectFn(profile, wallet.coinId(), wallet.networkId());
+
+	await wallet.ledger().connect(transport);
```
faustbrian: What's the reason for connecting here? The SDK will now automatically connect to the ledger and also disconnect from it after it has signed the transaction.

goga-m: Right, is aborting the connection handled as well?

faustbrian: Not sure what you mean by aborting? We connect, sign the transaction and then disconnect.

goga-m: Ah I mean if this https://github.com/ArkEcosystem/desktop-wallet/pull/3846/files#diff-f7cc672ff8254a45a8770d12189d9ac7d7b7a38c53b2ccfc2f03a06d1fbf14e1R76 will play well because I haven't tested in dev mode

faustbrian: This should probably now instead wrap the `signFn` call which will connect, sign and disconnect all in one go instead of having to call a lot of separate functions.

goga-m: @faustbrian it appears that the connection needs to run through `useLedgerConnection` https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/app/contexts/Ledger/hooks/connection.ts#L52 because it updates internal states that are required to render the ledger review step.  For the context of this PR, I only moved the connect methods higher than `transactionBuilder` for better readability, and we can focus to clean and simplify this particular code on a follow-up PR.

src/domains/transaction/pages/SendIpfs/SendIpfs.tsx
```diff
@@ -90,8 +90,12 @@ export const SendIpfs = () => {
 		};
 
 		try {
-			const abortSignal = abortRef.current?.signal;
+			if (activeWallet.isLedger()) {
```
faustbrian, line 93: Why do we need to manually connect here? The SDK will connect when it actually needs a connection and not otherwise, see https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk-ark/source/transaction.service.ts#L219-L220.

src/domains/transaction/pages/SendTransfer/SendTransfer.tsx
```diff
@@ -236,6 +236,11 @@ export const SendTransfer = () => {
 				setLastEstimatedExpiration(transactionInput.data.expiration);
 			}
 
+			if (activeWallet.isLedger()) {
```
faustbrian, line 239: Why do we need to manually connect here? The SDK will connect when it actually needs a connection and not otherwise, see https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk-ark/source/transaction.service.ts#L219-L220.

src/domains/transaction/pages/SendIpfs/SendIpfs.tsx
goga-m: Yes actually we need to call connect from `useLedgerConnection` https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/app/contexts/Ledger/hooks/connection.ts#L52 because it updates the internal state that is necessary for the ledger review step

src/domains/transaction/pages/SendTransfer/SendTransfer.tsx
goga-m: Same as above, from the initial implementation, dw manages the connection status in its own hook. If you comment out these lines, the tests will fail to render the ledger review step which indicates this issue. We could revisit this implementation, which I think will require some extra effort and testing to make sure the existing functionality will not break.

PR #3860, opened 2021-06-14
chore: update Platform SDK dependencies

faustbrian:

PR #3863, opened 2021-06-14
refactor: ensure human-readable values are used consistently

alesmit:
## Summary

https://app.clickup.com/t/kza6rn

## Checklist

-   [ ] Ready to be merged


faustbrian:
Task linked: [CU-kza6rn Ensure that human-readable values are used consistently](https://app.clickup.com/t/kza6rn)

w3ea:
i think it's already done here https://github.com/ArkEcosystem/desktop-wallet/pull/3860/commits/e88b4218322d36f8f01fcf7c862b6815688a7d68 and https://github.com/ArkEcosystem/desktop-wallet/pull/3860/commits/644454a4500b48ae7f738b7d0981de5d2c32381d

faustbrian:
Those commits were just temporarily solutions.

src/domains/transaction/pages/SendTransfer/SendTransfer.tsx
```diff
@@ -179,7 +181,7 @@ export const SendTransfer = () => {
 
 		const remaining = remainingBalance.minus(fee);
 
-		setValue("displayAmount", remaining.toHuman());
+		setValue("displayAmount", remaining.toHuman(decimals));
```
faustbrian, line 184: The decimals should never be used manually by DW. If there is an issue with normalising this when calling `toHuman()` then this would indicate an issue in the SDK.

faustbrian:
@alesmit As mentioned yesterday in slack, some of the quirks have already been addressed in https://github.com/ArkEcosystem/desktop-wallet/commit/8eda98e2aebe2ebd60f5dcfadb82b9d0d6a8401c but there's more to this issue.

The main issue lies with how DW consumes `BigNumber` values. The SDK returns everything in satoshis because it is agnostic to how the client will consume the data. The issue that arises from this with DW is that it uses a mix of the satoshi and human values. This is made quite obvious by a lot of numerical components that require a `normalize` property because a mix of satoshi and human is used. This is error prone and will eventually lead to wrong calculations in the SDK because satoshis were passed in while a human value was expected and now a huge value is calculated which could result in a user sending out all their funds.

I created https://app.clickup.com/t/mfa0tr yesterday to give an example of what the issue with such components is. The general issue is that all of the normalising is handled on a case-by-case basis instead of passing in the `BigNumber` instances and then letting the components decide how to display or normalise it. Right now we mutate a lot of data before it is passed to components which results in a lot of duplication and inconsistencies, which become more obvious during SDK upgrades, which is why the normalising has to be streamlined.

We either need to always pass `BigNumber` instances to components and let them decide what to do with it without any additional flags like `normalize` or DW needs to make consistent use of `toHuman()`. The consistent use of `toHuman()` would be a potential solution but long-term this will result in a lot of duplication, which again, will lead to inconsistencies and bugs.

alesmit:
@faustbrian makes sense. I also had noticed that calculations and formatting were spread across different components and I agree that this is confusing and even harder to maintain. About your suggestion:

> We either need to always pass `BigNumber` instances to components and let them decide what to do with it without any additional flags like `normalize` or DW needs to make consistent use of `toHuman()`.

I would go with always passing `BigNumber` instances as well. This seems to be straightforward for the values that the SDK gives us. My questions are more around values we pass as input. For example, for components like `InputFee` and `InputCurrency` it's easier to deal with strings rather than having `value` and `onChange` receive and emit `BigNumber` values. But if they keep strings at some point there will be a component that would need to convert that string to a `BigNumber` in order to display its value. What would you recommend doing in this scenario? Should we update these components to use strings only internally?

faustbrian:
I think changing them to internally work with strings makes the most sense because it means we can keep the API of all (or most) components the same by having amounts/fees be `BigNumber` instances and they figure out themselves what to do with it.

If required in some scenarios we could even create decorator components like `DisplayAmount` which could internally use the `Amount` component but will take care of calling `toHuman()`. That means we don't need to pass around flags because you'll be required to explicitly use the correct component. This is just an idea an ideally we wouldn't need those but if they are needed that would be a relatively clean approach.

alesmit:
I will make a new PR for this change

PR #3875, opened 2021-06-15
[EDU] Split components into blocks, styles and contracts

faustbrian:
Illustrates some of the points made in https://app.clickup.com/t/mf9618.

### DIFF
--- a/src/app/components/NavigationBar/NavigationBar.blocks.tsx
+++ b/src/app/components/NavigationBar/NavigationBar.blocks.tsx
@@ -1,0 +1,159 @@
+import { images } from "app/assets/images";
+import { Button } from "app/components/Button";
+import { Icon } from "app/components/Icon";
+import { NotificationsDropdown } from "app/components/Notifications";
+import { Tooltip } from "app/components/Tooltip";
+import { useConfiguration } from "app/contexts";
+import { useScroll } from "app/hooks";
+import { ReceiveFunds } from "domains/wallet/components/ReceiveFunds";
+import { SearchWallet } from "domains/wallet/components/SearchWallet";
+import { SelectedWallet } from "domains/wallet/components/SearchWallet/SearchWallet.models";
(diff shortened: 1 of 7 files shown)


src/app/components/NavigationBar/NavigationBar.blocks.tsx
```diff
@@ -0,0 +1,159 @@
...
+	variant,
+}: NavigationBarProps) => {
+	if (variant !== "full") {
+		return;
+	}
+
+	return <BackButton className="flex w-12" disabled={isBackDisabled} />;
+};
+
+const renderMenu = ({
```
faustbrian, line 36: This function is still pretty large and should probably be extracted into a separate component to work more like `renderBackButton` where a one-liner is returned for a component instead of building it inside the function.

alesmit:
@faustbrian a couple of questions:
- Would it make sense to use this approach for any component or just for the biggest ones?
- By "contract" I assume you mean interfaces and types. What would be the best place to keep constants and enums?

src/app/components/NavigationBar/NavigationBar.contract.ts
```diff
@@ -0,0 +1,28 @@
+import { Contracts } from "@arkecosystem/platform-sdk-profiles";
```
alesmit, line 1: Think it would be good to rename this to `<ComponentName>.contracts.tsx` (plural) to be consistent with other file names

faustbrian:
> * Would it make sense to use this approach for any component or just for the biggest ones?

This should be a general approach to make it a habit and also make it easier for new people to jump in because they will know that specific things are in specific files, without an exception. If you only do it for a few you won't really form a habit out of it.

> * By "contract" I assume you mean interfaces and types. What would be the best place to keep constants and enums?

Constants/Enums I would also classify as contracts because they contain non-negotiable values that your application/component expects.

PR #3900, opened 2021-06-17

faustbrian:


src/domains/transaction/components/IpfsDetail/IpfsDetail.tsx
```diff
@@ -32,7 +32,7 @@ export const IpfsDetail = ({ isOpen, transaction, onClose }: IpfsDetailPropertie
 				border={false}
 			/>
 
-			<TransactionFee currency={wallet.currency()} value={transaction.fee()} />
+			<TransactionFee currency={wallet.currency()} value={transaction.fee().toHuman()} />
```
faustbrian: I'll make some more changes in the SDK after this PR to return this also as a human-readable value.

w3ea:
and here 
https://github.com/ArkEcosystem/desktop-wallet/blob/develop/src/app/components/Amount/Amount.tsx#L59

`defaultProps` :deanwine:

PR #3923, opened 2021-06-22
fix: use `Currency` instead of `Fiat` for labels

faustbrian:


faustbrian:
Task linked: [CU-mvetn2 Wallet list uses "Fiat" while transaction list uses "Currency"](https://app.clickup.com/t/mvetn2)

PR #3926, opened 2021-06-22
fix: treat `amount` and `fee` as numbers for total amount

faustbrian:


faustbrian:
Task linked: [CU-mx7we3 [TX Review Step] TotalAmountBox should sum fee+amount as numbers](https://app.clickup.com/t/mx7we3)

PR #3931, opened 2021-06-23
refactor: reorganise signing method conditionals based on their preferred usage

faustbrian:


faustbrian:
Task linked: [CU-mx7w45 [Second Sig Reg] TX Fails on Devnet](https://app.clickup.com/t/mx7w45)

PR #3938, opened 2021-06-23
fix: only show `return` icon for transfers and multi-payments

faustbrian:


faustbrian:
Task linked: [CU-mz7hxy [Transactions] show is-return icon only in transfers and multipayments](https://app.clickup.com/t/mz7hxy)

PR #3959, opened 2021-06-25
fix: enforce recipient limit in multipayment

alesmit:
## Summary

https://app.clickup.com/t/mxd0fb

## Checklist

-   [x] Tests
-   [ ] Ready to be merged


faustbrian:
Task linked: [CU-mxd0fb [Multipayment] recipient limit not enforced](https://app.clickup.com/t/mxd0fb)

faustbrian (changes requested):
Lacking some coverage

src/domains/transaction/components/AddRecipient/AddRecipient.tsx
```diff
@@ -104,21 +102,21 @@ export const AddRecipient = ({
 	const { network, senderAddress, fee, recipientAddress, amount } = watch();
 	const { sendTransfer } = useValidation();
 
-	const senderWallet = useMemo(() => profile.wallets().findByAddress(senderAddress), [profile, senderAddress]);
-
 	const ticker = network?.ticker();
 	const exchangeTicker = profile.settings().get<string>(Contracts.ProfileSetting.ExchangeCurrency) as string;
 	const { convert } = useExchangeRate({ ticker, exchangeTicker });
 
+	const maxRecipients = wallet?.network().multiPaymentRecipients() ?? 0;
```
goga-m, line 109: Hey @alesmit the Nullish coalescing operator might not be necessary as the `multiPaymentRecipients` will return 0 from sdk for default value :hmhm: https://github.com/ArkEcosystem/platform-sdk/blob/master/packages/platform-sdk/source/networks/network.ts#L278  

```suggestion
	const maxRecipients = wallet?.network().multiPaymentRecipients();
```

faustbrian: I would assume this triggers when `wallet?` is true which means it'll run as `undefined ?? 0`

faustbrian (approved):

PR #4019, opened 2021-07-05
feat: require wallet alias

alesmit:
## Summary

https://app.clickup.com/t/ndcdmk

## Checklist

-   [x] Tests
-   [x] Ready to be merged


faustbrian:
Task linked: [CU-ndcdmk Require wallet names](https://app.clickup.com/t/ndcdmk)

src/domains/wallet/pages/ImportWallet/ImportWallet.tsx
```diff
@@ -98,6 +100,13 @@ export const ImportWallet = () => {
 		setActiveTab(activeTab - 1);
 	};
 
+	const getDefaultAlias = (wallet: Contracts.IReadWriteWallet): string => {
+		const ticker = wallet.network().ticker();
+		const sameCoinWalletsCount = Object.keys(activeProfile.wallets().allByCoin()[ticker] ?? {}).length;
+
+		return `${ticker} #${sameCoinWalletsCount}`;
```
faustbrian:
```suggestion
		return `${ticker} #${sameCoinWalletsCount + 1}`;
```

Unless I'm overlooking something we currently would start from 0, which isn't how people generally count.

alesmit: When we generate the alias the wallet is already pushed in the list so that would never be 0

w3ea: @alesmit then why do we need a nullish operator `??` at the L105?
https://github.com/ArkEcosystem/desktop-wallet/pull/4019/files#diff-ed92ad65a44e2ba457bc9b734ed625d272bed384af4bb1eec5c05b1d5fc67ed9R105

alesmit: @w3ea good catch. I've updated it. Also changed a bit the implementation to cover for possible duplicates when some user manually renames an existing wallet with an alias that we would auto-generate.

w3ea: @alesmit nice refactor 👍🏻
just FYI, here we're saving name twice (If the user doesn't change the name)
https://github.com/ArkEcosystem/desktop-wallet/pull/4019/files#diff-ed92ad65a44e2ba457bc9b734ed625d272bed384af4bb1eec5c05b1d5fc67ed9R158

